PR #113, opened 2014-07-05
ENH: Never return negative uncertainty; return NaN instead.

danielballan:
This is a work in progress.

The Savin-Doyle procedure for estimating uncertainty fails in dense systems or images with distinct background features. This often leads to 'signal' being negative. Instead of returning a negative value for the uncertainty 'ep', we should return NaN to make it clear that an uncertainty could not be estimated.

The tests for this need to be designed carefully, and this PR is not there yet.

### DIFF
--- a/trackpy/tests/test_feature.py
+++ b/trackpy/tests/test_feature.py
@@ -8,7 +8,7 @@ from trackpy.try_numba import NUMBA_AVAILABLE
 
 import unittest
 import nose
-from numpy.testing import assert_almost_equal, assert_allclose
+from numpy.testing import assert_almost_equal, assert_allclose, assert_equal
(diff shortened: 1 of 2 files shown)


danielballan:
This is an important, API-breaking issue that needs to be addressed. Marking for 0.3.

nkeim:
Good. In the release notes, we can recommend using `fillna(-1)` when backward compatibility is an issue.

nkeim:
Bump. As we approach v0.3, it's time to break compatibility a little. @danielballan , let me know if you want me to tackle the rebase.

danielballan:
That would be great. I'll revisit the logging PR and get that ready if you can attack this.

Issue #108, opened 2014-06-24
Roadmap: Enhancements to finish before or during SciPy conference


danielballan:
A tour of our [walkthrough notebook](http://nbviewer.ipython.org/github/soft-matter/trackpy-examples/blob/master/notebooks/walkthrough.ipynb) alone could fill 15 minutes. There's no shortage of features to discuss. But I'd like to use the impending deadline and the conference's "Sprints session" as an excuse to finish work on some enhancements.

This is my list of ideas. If you have anything to offer or to add, please speak up.
- [ ] Try to connect mpld3 to IPython widgets to build a fast, interactive, zoomable feature-filtering tool. No one has connected mpld3 to widgets yet, though it has been suggested. This will be of wide interest if it works. I have done preliminary work on this.
- [x] Display features or trajectories drawn onto HTML5 videos embedded in notebooks. I did the hard part -- the video -- for another project. I just have to connected to trackpy.
- [ ] Consider other ways that widgets and/or mpld3 could make life easier. (_Suggestions?_)
- [ ] Consider making more use of the tools in scikit-image. Broadly, I want to have a clear understanding of how each of our imaging steps (e.g., bandpass) is related to and distinct from tools in scikit-image. That project will be a major focus at the meeting.

These almost definitely won't fit into the talk, but I'd like to finish them soon anyway.
- [ ] Implement trajectory splitting/joining and a more sophisticated gap-closing (like "memory") that is globally optimized in time. See [Jaqaman et. al](http://www.nature.com/nmeth/journal/v5/n8/abs/nmeth.1237.html). This is about half-written.
- [ ] Implement clustering of heterogeneous trajectories into homogenous sub-ensembles using an F-statistic, following [Valetine et. al.](http://crocker.seas.upenn.edu/Valentine2001.pdf). This is written, but I want to test it on real data before I submit a PR.

sofiamagk:
Hello,

Is anyone working on annotating and trajectory-plotting in an interactive way? 

For example, I would love to be able to hover over an identified spot and see its eccentricity, or its mass, so as to choose tracking parameters with a direct sense for what they represent on my dataset. 

I am thinking of integrating some plotly features with trackpy's plotting functions; I see here there have been similar motions with mpld3. 

What is the status of this work? 

Thanks a lot!

PR #114, opened 2014-07-05
Experimental: Interactive, zoomable feature filtering widget

danielballan:
This is the feature that I mentioned in #108. [Here is a demonstration.](http://nbviewer.ipython.org/gist/anonymous/58bdde32f7490845f6af) You will need to run it locally to see the widget. It's pretty cool.

I pushed this to a new branch of soft-matter/trackpy so it will be easy for others to find. We can remove that once we decide to merge it into master. I imagine that this `trackpy.interactive` module will never be auto-imported; it will be exclusively for use in the notebook.

### DIFF
--- a/trackpy/interactive.py
+++ b/trackpy/interactive.py
@@ -1,0 +1,190 @@
+from __future__ import print_function
+
+import warnings
+
+# We will use IPython widgets and traitlets and some of the mpld3 internals.
+import matplotlib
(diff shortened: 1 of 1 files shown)


danielballan:
This idea might someday live on as a demo notebook once the fast-changing landscape of interactive plotting stabilizes, but I don't think it needs to be a function in trackpy.

PR #171, opened 2014-10-15
BLD: Add min versions of required deps to setup.py.

### LINKED ISSUE #159: Specify minimum versions of dependencies.
Specify minimum versions in `setup.py`. This came up because `np.meshgrid` used to be 2D only, so there is some minimum version of numpy before which our ND functionality fails (and fails cryptically).

I'm not sure what the minimum version is, but most projects don't support numpy before 1.7.

danielballan:
Closes #159. These same minimum versions were also specified in the recipes used to build the conda packages for 0.2.4.

Notes:
* scipy 0.12 added full-featured cKDTree
* pandas 0.12 added GroupBy.filter

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -32,8 +32,9 @@ setup_parameters = dict(
     author = "Trackpy Contributors",
     author_email = "daniel.b.allan@gmail.com",
     url = "https://github.com/soft-matter/trackpy",
-    install_requires = ['numpy', 'scipy', 'six', 'pandas',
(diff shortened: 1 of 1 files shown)


nkeim:
This is closed, but it was never merged and the original issue #159 is still open. Mistake?

danielballan:
Yes, mistake. Thanks.

danielballan:
Bump. This is now fixed.

PR #174, opened 2014-10-15
FIX/TST: When an empty frame is put to FramewiseData, warn but continue.

danielballan:
Currently, if one frame happens to have zero features, `PandasHDFStore` or `PandasHDFStoreBig` will raise, stopping the feature-finding loop.

In this PR, the `put` method for each of these is modified to simply do nothing if the DataFrame passed to it is empty. It warns the user that an empty frame was passed, but it doesn't interrupt the process.

### DIFF
--- a/trackpy/framewise_data.py
+++ b/trackpy/framewise_data.py
@@ -3,6 +3,7 @@ from __future__ import (absolute_import, division, print_function,
 import six
 import os
 from abc import ABCMeta, abstractmethod, abstractproperty
+import warnings
 
 import pandas as pd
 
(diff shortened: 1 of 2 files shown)


nkeim:
Sorry @danielballan , I think I missed my @ mention. As a workaround, I used the warnings-test logic from `test_misc.py`, which ignores the unexpected `DeprecationWarning` in some Travis environments. See https://github.com/nkeim/trackpy/tree/handle-empty-put . Let me know if you want me to make that a new PR.

danielballan:
Great. I didn't know about that. Please do make that a pull request.

nkeim:
Superseded by #250

Issue #177, opened 2014-10-24
Colorby='frame' not working


kmsouthard:
Hello everyone!

I am fairly new to python so perhaps this is a user error, but I cannot get colorby='frame' to work for trackpy.plot_traj function.

I either get axis centered around zero and not even getting to 1 pixel or, if I try and add labels the plot disappears, and my output is just something like this:

<matplotlib.figure.Figure at 0x3b73d9550>

If I am doing something wrong or I need to check something, any help would be greatly appreciated!

MTone92:
I am a newbie for python and trackpy. I also encountered a similar problem. Instead of not plotting the lines with `tp.plot_traj(my_track, colorby='frame')`, the plotted lines are in the same color. I manually checked the code in trackpy/[plots.py](https://github.com/soft-matter/trackpy/blob/master/trackpy/plots.py) where the function `plot_traj` is defined.  Under the `if colorby == 'frame':`, if some particles don't appear in all frames, the plotted lines are in the same color, since the number of plotted segmentations is less that the number of colors supplied to the function `lc.set_array(color_numbers)`.
I tested some cases with matching the number of segments and number of color in a stand alone .py file with:
`if colorby == 'frame':
        # Read http://www.scipy.org/Cookbook/Matplotlib/MulticoloredLine
        x = traj.set_index([t_column, 'particle'])['x'].unstack()
        y = traj.set_index([t_column, 'particle'])['y'].unstack()
        color_numbers = traj[t_column].values/float(traj[t_column].max())
        norm = plt.Normalize(color_numbers.min(), color_numbers.max())
        color_max = float(traj[t_column].max())
        logger.info("Drawing multicolor lines takes awhile. "
                    "Come back in a minute.")
        for particle in x:
            x_series = x[particle].dropna()
            y_series = y[particle].dropna()
            frames = np.array(x_series.index)
            frames_map = frames/color_max
            points = np.array(
            [x_series.values, y_series.values]).T.reshape(-1, 1, 2)
            segments = np.concatenate([points[:-1], points[1:]], axis=1)
            lc = LineCollection(segments, cmap=cmap, norm=norm)
            lc.set_array(frames_map)
            ax.add_collection(lc)
            ax.set_xlim(x.apply(np.min).min(), x.apply(np.max).max())
            ax.set_ylim(y.apply(np.min).min(), y.apply(np.max).max())`
I am not sure if it is a correct way to do it, but it solved my problem.
[...]

PR #178, opened 2014-10-27
clean up plots.py a bit

tacaswell:
Issues found looking at #177 none of which seem to actually fix that problem.


danielballan:
Let's address this after mpl 1.5 is released. The improved `plot()` API for labeled data and then `ensure_ax` decorator are both relevant to us. Marking for trackpy v0.4, but this could be v0.3.x if we tag one.

caspervdw:
@tacaswell  Do you want to keep this still open? If so, please rebase.

PR #184, opened 2014-12-11
ENH: Configure proper logging.

### LINKED ISSUE #183: Use logging for printing progress
It would be nice if `print_update` went through normal logging -- possibly with a custom handler added if trackpy detects that it is run in IPython, but some people may just want to use it outside of that...

danielballan:
Closes #183.

This PR creates a logger at the module level and uses it where print calls (to be precise, `print_update` calls) were previously used. A function, `handle_logging()`, called at import time, does logging configuration to create the same user experience we have now. But users can turn off or configure logging by accessing `trackpy.logger`. They can restore "factory default" logging -- undoing the actions of `handle_logging()` -- by calling `ignore_logging()`.

Discuss!


nkeim:
I'd like to bump this PR for merge, as soon as @tacaswell's line-level comments above are addressed. The reason is that, taking @tacaswell's concerns on #192 to heart, it would be great to add some debug-level logging for subnets and adaptive search. This would let users actually see how adaptive search is being (ab)used, and would allow me to give much more useful advice in the tutorial.

trackpy/framewise_data.py
```diff
@@ -8,8 +8,8 @@
 
 from .utils import print_update
 
-__all__ = ['FramewiseData', 'PandasHDFStore', 'PandasHDFStoreBig',
-           'PandasHDFStoreSingleNode']
+__all__ = ['PandasHDFStore', 'PandasHDFStoreBig', 'PandasHDFStoreSingleNode',
```
danielballan: I made the same change in two different PRs by mistake. So, this change is not consequential.

danielballan:
I address @tacaswell's comments and rebased. The rebase was messy -- encompassing the 3D PR -- so please take one more look through to see if I missed anything.

Grep confirms that all references to `print_update` are gone. I added a section on logging to the API docs explaining our approach.

@nkeim I think there should be a way to turn on progress reports (currently INFO-level) separately from details like the number of subnets (proposed DEBUG-level). Do you think we should add another convenience function or simply document how to set a logging level? I lean toward the latter but I'm open to your preference.

trackpy/feature.py
```diff
@@ -223,7 +225,7 @@ def _refine(raw_image, image, radius, coords, max_iterations,
         for iteration in range(max_iterations):
             off_center = cm_n - radius
             if walkthrough:
-                print_update(off_center)
+                logger.info('%f', off_center)
```
tacaswell: Could you put some more words here here describing this?  something like `logger.info('off center: %f', off_center)`

Is there a plan do ditch `walkthrough` and make these debugging-level logging calls eventually?

trackpy/feature.py
```diff
@@ -457,30 +459,39 @@ def locate(raw_image, diameter, minmass=100., maxsize=None, separation=None,
 
     Parameters
     ----------
-    image : image array (any dimensions)
-    diameter : feature size in px
+    raw_image : array
```
tacaswell: re api consistency at day job, should this stay as `image`.  The description should be something like 'the image to find features in'

trackpy/feature.py
```diff
@@ -457,30 +459,39 @@ def locate(raw_image, diameter, minmass=100., maxsize=None, separation=None,
...
+         any N-dimensional image
+    diameter : number or tuple
         This may be a single number or a tuple giving the feature's
         extent in each dimension, useful when the dimensions do not have
         equal resolution (e.g. confocal microscopy). The tuple order is the
         same as the image shape, conventionally (z, y, x) or (y, x). The
         number(s) must be odd integers. When in doubt, round up.
-    minmass : minimum integrated brightness
+    minmass : number
```
tacaswell: number -> scalar

trackpy/feature.py
```diff
@@ -736,10 +758,12 @@ def batch(frames, diameter, minmass=100, maxsize=None, separation=None,
     engine : {'auto', 'python', 'numba'}
     output : {None, trackpy.PandasHDFStore, SomeCustomClass}
         If None, return all results as one big DataFrame. Otherwise, pass
-        results from each frame, one at a time, to the write() method
+        results from each frame, one at a time, to the put() method
         of whatever class is specified here.
-    meta : By default, a YAML (plain text) log file is saved in the current
-        directory. You can specify a different filepath set False.
+    meta : filepath_or_obj
```
tacaswell: `filepath or object`

tacaswell: + ', optional'

trackpy/feature.py
```diff
@@ -771,11 +795,11 @@ def batch(frames, diameter, minmass=100, maxsize=None, separation=None,
                      filter_before=filter_before, filter_after=filter_after)
 
     if meta:
-        if isinstance(meta, str):
-            filename = meta
+        if isinstance(meta, six.string_types):
+            with open(filepath_or_obj, 'w') as file_obj:
```
tacaswell: shouldn’t this be `open(meta, 'w') ... ` ?  

I am also not a huge fan of blind writes (which will nuke what ever is there).  However, apparently the 'barf if file exists at that path' mode is a hdf5-ism, not part of the standard `open` so should probably ignore this complaint for now.

trackpy/feature.py
```diff
@@ -771,11 +795,11 @@ def batch(frames, diameter, minmass=100, maxsize=None, separation=None,
...
+                record_meta(meta_info, file_obj)
         else:
-            filename = 'feature_log_%s.yml' % timestamp
-        record_meta(meta_info, filename)
+            record_meta(meta_info, file_obj)
```
tacaswell: Should this be `record_meta(meta_info, meta)`?

trackpy/utils.py
```diff
@@ -187,11 +189,61 @@ def validate_tuple(value, ndim):
...
+        trackpy.logger.propagate = False
+    trackpy.logger.addHandler(default_handler)
+    trackpy.logger.setLevel(logging.INFO)
+
+
+def ignore_logging():
+    "Reset to factory default logging configuration; remove trackpy's handler."
+    trackpy.logger.removeHandler(default_handler)
+    trackpy.logger.setLevel(logging.NOTSET)
+    trackpy.logger.propagate = 1  # default implemented by the logging module
```
tacaswell: It is really `1` not `True`?

trackpy/feature.py
caspervdw: Or integer? I thought that the purpose here is stating the dtype.

nkeim:
@danielballan: Good point about turning on debug messages. I'm actually leaning toward an API call that covers common cases. I think we expect many `trackpy` functions to produce useful debug messages, and we'll want to mention this in each function's docstring or tutorial. In those places, it would be great to refer users to a single function, instead of a separate tutorial on how to configure logging. 

`quiet()` is nice because it lines up a logging facility feature with a very clear user intent. But here, we have a few problems:

1. The IPython hook to clear the most recent message is going to be confusing and counter-productive when debug output is turned on. (This fact alone means that simply setting the logging level will not suffice.)
- Especially for a short movie or operation, one would probably prefer to see the debug output scroll by in the interactive session.
- For a long movie that runs unattended, it would be best to send the debug messages to a file, while still viewing progress updates in the console.

In other words, here user intents are less obvious. Even though the Python logging facility can address all of these cases, creating a user-friendly API solution could just be overly complicated. If we were to create such an API, I would propose something like:

- `trackpy.debug(True)` (default) — turn off the clear-output hook (if IPython). Set stdout level to DEBUG
- `trackpy.debug(file-like object)` — create a separate logging output to this file
- `trackpy.debug(False)` — turn on the clear-output hook, set stdout level to INFO, and close the output file

Note that only `debug(False)` returns the logging facility to a known state. Note also that `debug(False)` would *cancel* an earlier `quiet(True)` call. This could be confusing. A unified `trackpy.progress()` or `trackpy.log()` function might be better.

Unless everyone agrees that this kind of solution is compelling, it might be wise to save it for a separate PR (that I volunteer to try writing), so that we can try it out with actual code that makes debug messages, and consider the alternative of just instructing the user on how to re-configure the logging facility.

tacaswell:
How far down this rabbit hole do we want to go?  I can also see a case for fairly-fine grain control of debugging output (for example, I trust the linking step of batch, but what to see diagnostics on refine).

I was also thinking that if you are using adaptive linking it may be worth adding a column to the returned dataframe saying what search range was used for each link.  I can see a use-case where you filter tracks (or segments there of) based on the search ranges used.

Would it be worth adding an option to just give up on large sub-nets, break all the tracks in them, and move on to the next frame.  It will artificially break up your tracks, and in some cases like looking for the soft modes obliterate the science, but in others, like micro rheology in sparse systems, it will filter out randomly over-dense regions.  In the latter case you could set the max size to be really small to make sure that you don't miss-link anything (or just turn sub-nets off).

nkeim:
If debug messages are prefixed with the name of the generating function (or algorithm), the user could select the ones of interest. Seems like a good idea.

apiszcz:
The python logger can impose overhead, we should try to keep it out of any high event rate sections of code. Worse case would be to have yet another 'if' test to enable/disable debug modes. I like the suggestions from nkiem, including def trackpy.quiet(suppress=True). Concur with more information on modes like adaptive search. Other suggestions to understand pauses like the one here at frame 577 we have a ~6 minute pause, memory seems stable,  probably working through these with pdb is still needed, so expecting the logger to do it all may not be worth the energy in all places. 

2015-03-06T09:03:22.496000> Frame 576: 43 trajectories present
2015-03-06T09:03:22.613000> Frame 577: 43 trajectories present
2015-03-06T09:09:57.067000> Frame 578: 45 trajectories present
2015-03-06T09:09:57.107000> Frame 579: 45 trajectories present

tacaswell:
I thought part of the selling point of the logging framework as that they were relatively efficient if they are not going to be used (ex if their level is below the cut off) and they defer formatting the strings as long as possible.

apiszcz:
Concur on theory, implementations have trade offs and high event rate
activities will be slowed with logging anything. So sometimes it easier
(esp. with python since source is usually around to dig in). We had some
situations with the logging framework had issues with writing to
intermittently connected shared drives. A few emissions a second or slower
is reasonable for logging, I'm less certain about high I/O event rates into
loggers.

apiszcz:
I am running three adaptive link_df tasks and number of other processing
steps. The linking activity is typically 5-30% of my overall processing
time. (BTW: ADAPTIVE is an impressive performance improvement) I commented
out the lines around 588 shown below

I performed 3 runs of each and the average is

42s RUN TIME WITH print_update MESSAGE
24s RUN TIME WITH print_update COMMENTED OUT

Note certainly the number of detections under consideration impact this
overhead, if there are a lot of detections the print overhead is lower
since linking will take longer, however with few detections < 10, the
overhead is significant. In my testing 42/34=23.53% slower with the print
statement. For my purposes I'll be omitting (commenting out the printing
statement and only using it when I need to understand possible linking
issues).

Another idea might be store this output in a data frame and return as a
part of the predict return dictionary
(tr,statusdataframe)=link_df(...)


FILE: linking.py

        #msg = "%s> Frame %d: %d trajectories present" %
(datetime.datetime.now().isoformat(), frame_no, len(labels))
        #print_update(msg)

danielballan:
The plan is to completely remove print_update from trackpy and use logger
instead. I would be curious to know what the perf cost of logger is in your
use case. It would be good to know both the cost when log messages are
issued (e.g., printed to stdout) and when they are muted.

tacaswell:
@danielballan This needs a rebase.

danielballan:
Rebased again. This touches almost every file, so let's get it in soon. :- D

We currently have `trackpy.quiet()`. I like @nkeim's suggestion of `trackpy.debug(True)`, `trackpy.debug(False)`, and `trackpy.debug(filename or some open file)`. (Review his comment from Jan 11 for details.) Are we agreed that this is the right way to go?

I think @tacaswell is correct that users will want "granular control," debugging location or but not linking, or vice versa. But we don't need to do anything fancy: users can achieve this by switching logging on and off as desired during the process. Maybe there's a clever way to do this, but I don't think we have to tackle it from the start.

I am tempted to nuke the YAML-spewing feature altogether. I never use it now that we have notebooks, and it's easy enough to users to replicate something like that on their own if they want it. Not sure it belongs in trackpy. Let me know if you disagree.

trackpy/feature.py
```diff
@@ -483,7 +496,7 @@ def locate(raw_image, diameter, minmass=100., maxsize=None, separation=None,
         based on their estimated mass and size before refining position.
         True by default for performance.
     filter_after : boolean
-        Use final characterizations of mass and size to eliminate spurious
+        Use final characterizations of mass and size to elminate spurious
```
tacaswell: spelling

trackpy/feature.py
```diff
@@ -663,7 +676,7 @@ def batch(frames, diameter, minmass=100, maxsize=None, separation=None,
           percentile=64, topn=None, preprocess=True, max_iterations=10,
           filter_before=True, filter_after=True,
           characterize=True, engine='auto',
-          output=None, meta=True):
+          output=None, meta=None):
```
tacaswell: Was this intentional?

trackpy/feature.py
```diff
@@ -673,30 +686,39 @@ def batch(frames, diameter, minmass=100, maxsize=None, separation=None,
 
     Parameters
     ----------
-    frames : list (or iterable) of images
-    diameter : feature size in px
+    raw_image : array
```
tacaswell: This seems like `frames` and `raw_image` are very different.

trackpy/feature.py
tacaswell: bump

trackpy/feature.py
```diff
@@ -762,11 +787,11 @@ def batch(frames, diameter, minmass=100, maxsize=None, separation=None,
...
     if meta:
-        if isinstance(meta, str):
-            filename = meta
+        if isinstance(meta, six.string_types):
+            with open(filepath_or_obj, 'w') as file_obj:
+                record_meta(meta_info, file_obj)
         else:
-            filename = 'feature_log_%s.yml' % timestamp
-        record_meta(meta_info, filename)
+            record_meta(meta_info, file_obj)
```
tacaswell: Should'n this be `file_obj -> `meta` as `file_obj` only exists if the above context manager gets hit.

danielballan:
I shouldn't have rebased on so little sleep. Hold off on further review until I can take a closer look at this.

trackpy/feature.py
danielballan: rebasing mistake, fixed

trackpy/feature.py
danielballan: Yes. Current default behavior: spew yaml files with automatically-generated file names. New default behavior: be quiet. Require the user to specify a file handle or filename.

trackpy/utils.py
danielballan: No, I read that somewhere but it was incorrect.

trackpy/feature.py
danielballan: OK, 'int or float' is more clear.

doc/api.rst
```diff
@@ -116,6 +116,26 @@ optimizations. :class:`~trackpy.framewise_data.PandasHDFStoreBig` is a good gene
 That last class cannot be used directly; it is meant to be subclassed
 to support other formats. See *Writing Your Own Interface* in the streaming tutoral for more.
```
sciunto: streaming tutor*i*al :)

danielballan:
This is ready for another round of review.

trackpy/feature.py
```diff
@@ -275,8 +278,7 @@ def _refine(raw_image, image, radius, coords, max_iterations,
         allow_moves = True
         for iteration in range(max_iterations):
             off_center = cm_n - radius
-            if walkthrough:
-                print_update(off_center)
+            logger.debug('off_center: %f', off_center)
             if np.all(np.abs(off_center) < GOOD_ENOUGH_THRESH):
```
sciunto: Not related to this PR but, GOOD_ENOUGH_THRESH looks like a magic number. Maybe an optional parameter to the function would be better. Same for SHIFT_THRESH.

tacaswell:
The `raw_image` -> `image` api change broke all of the tests.

trackpy/feature.py
```diff
@@ -342,7 +344,7 @@ def _refine(raw_image, image, radius, coords, max_iterations,
         return np.column_stack([final_coords, mass, Rg, ecc, signal, raw_mass])
 
 
-def locate(raw_image, diameter, minmass=100., maxsize=None, separation=None,
+def locate(image, diameter, minmass=100., maxsize=None, separation=None,
```
nkeim: This was just a rebase error. Changing `image` back to `raw_image` causes all tests to pass.

trackpy/plots.py
```diff
@@ -298,8 +299,8 @@ def plot_traj(traj, colorby='particle', mpp=None, label=False,
         x = traj.set_index([t_column, 'particle'])['x'].unstack()
         y = traj.set_index([t_column, 'particle'])['y'].unstack()
         color_numbers = traj[t_column].values/float(traj[t_column].max())
-        print_update("Drawing multicolor lines takes awhile. "
-                     "Come back in a minute.")
+        logger.info("Drawing multicolor lines takes awhile. "
```
nkeim: On my setup (IPython 3.2.0, Python 2), this message doesn't appear until the plot appears, making it rather pointless.

nkeim:
I'm happy with this — it addresses a pain point while leaving the door open for future enhancements. And the API feels right.

trackpy/plots.py
tacaswell: Probably has to do with the event loop.  There are some issues on OSX in general.  We have asyncio tests that work perfectly on linux, but go in all the wrong order on mac.

tacaswell: although, @danielballan wrote this presumably on a mac...

apiszcz:
Is a release near? or should I just grab the master to review. I've been
processing a lot of data using trackpy, (no images), and python 3.4.3. My
(code in place modification) for logging in the linking module was to
output to the screen every 50 frames, for my data it gives an update every
few seconds.
​

nkeim:
@apiszcz : If you're not doing any feature-finding, this branch as it currently exists should work. You can just `git pull https://github.com/danielballan/trackpy reinstate-logging`.

I don't think we've ruled out your idea to emit a progress message every *n* frames, but I don't see it going into v0.3.

AFAIK, this PR and #244 are the major remaining work to be done on v0.3 (besides all the usual activities of actually putting out the release).

trackpy/plots.py
danielballan: It doesn't *hurt* so I'll leave it in, and later we'll figure out if there is a way to flush this in advance of the long computation.

danielballan:
I addressed line comments and rebased. I was careful, but the rebase was tricky so please give it one last look before merging.

apiszcz:
Using the 'latest source' as of 8/17/2015 with a workload that represents about about 5% of a typical workload I'm seeing the following: ~12% difference for log vs. mod, understand I'm not writing anything to a file.

    log approach: 112.7893 seconds
    mod approach: 100.8021 seconds

        if frame_no%50==0:
            msg = "%s> Frame %06d: %06d trajectories present\r" % (datetime.datetime.now().isoformat(), frame_no, len(labels))
            sys.stdout.write(msg)

nkeim:
@apiszcz This is a useful data point; thanks. Just to be clear: did you use `trackpy.quiet()`?

I imagine this would be even worse when the notebook server is not local.

apiszcz:
I did not use trackpy.quiet(), I am still interested to know what the
general progress is (by frame). I can look into .quiet() if it provides
that. This is running local, I am not using ipython. When status consumes
evaluation or tracking time in my case it matters. Thanks.

danielballan:
@apiszcz As of Sunday, you have two options:

1. Log every frame
2. Log nothing.

I think the best way forward is to log based on time, as we discussed in #235, skipping logging if less than a second has passed since the last message. IMO we don't need to make this configurable. The new options can be:

1. Log each second or each frame, whichever occurs less frequently.
2. Log nothing.

Logging provides a heartbeat; more sophisticated diagnostics can be obtained from the results set itself.

@apiszcz Would you be interested in contributing a PR to add this to trackpy?

nkeim:
I have one concern about the time-based throttling: if linking raises an exception, you won't know which frame caused the error.

This can be partly addressed by demoting too-frequent messages to `DEBUG` level, instead of suppressing them completely.

danielballan:
Good point, and good solution.

apiszcz:
I have an approach borrowed in part from others that uses threaded timer,
not sure if there will 'event' confusion, when I have time I can test it
and propose for comment.

The API is
   rt=RepeatedTimer(floatseconds, functiontocallatinterval)

Disable
   rt.stop()

tacaswell:
major :-1: on bring threads to the party

nkeim:
Agree with @tacaswell . @apiszcz , would it be practical to store the time of the last emitted log message and compare it with now (is that `os.time()`?)?

apiszcz:
that sounds reasonable and fast. (subtract)

apiszcz:
os.times() on linux provides an elapsed real time since a fixed point in the past call as the 5th output tuple, however it is not supported on Windows.

Run times (in seconds) of three runs tracking on ~1.3 million simulated records, 8,641 frames, creating 10,061 tracks using each of the status methods.~155 trajectories active per frame. In this run, mod50 is updating at about 2 hertz.

    logger:         110.21, 108.80, 109.06
    mod50:          106.07, 105.33, 105.02
    interval(1s):   102.39, 103.93, 109.13
    interval(10s):  103.87, 103.94, 103.85

After reviewing these times this issue may 'not' meet the threshold for implementing a change.

danielballan:
Good to know. Thanks for investigating the numbers.

nkeim:
Sounds good. As alluded to above, I suspect that when using the notebook, especially over a remote connection, the performance difference could become large. In fact, this has been my experience anecdotally. I am happy using `trackpy.quiet()` to solve this problem, but I suspect we haven't seen the last of this issue.

apiszcz:
I like the interval idea, let the user specify the update interval, I'm
using that for now, it really depends on the density of points and how
'easy' they are to link. in easy to link situations the logging can be a
drag, in the difficult case it will be less than 1%.

Intervalupdate
0 = every frame
float = interval to update screen and/or logging

PR #188, opened 2014-12-19
BUG: Use unsigned ints for peak finding

nkeim:
When `locate()` is given a float image, it casts it to integers for more efficient peak-finding. (`refine()` is still done with the float image.) As currently written, float images are cast to signed 8-bit integers (max. value 127). This seems like a needless loss of precision that could actually have a performance cost, as it may increase the need for tie-breaking in the merging step. This PR uses unsigned 8-bit ints instead.

### DIFF
--- a/trackpy/feature.py
+++ b/trackpy/feature.py
@@ -572,7 +572,7 @@ def locate(raw_image, diameter, minmass=100., maxsize=None, separation=None,
     if np.issubdtype(raw_image.dtype, np.integer):
         dtype = raw_image.dtype
     else:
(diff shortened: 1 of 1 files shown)


danielballan:
Yep, I have to think this was a typo. According to "git blame" I touched that line one year ago, so we can't blame 2010-era novice-programmer Dan for this one. :- D

Thanks for catching it. I will open an issue to revisit the unit tests covering `refine` and subpixel precision. Maybe this change will measurably improve the precision and let us tighten the tolerances.

PR #192, opened 2015-01-05
Adaptive search

nkeim:
This PR implements a new feature I call "adaptive search," that is explained and demonstrated in two new tutorials. See https://github.com/soft-matter/trackpy-examples/pull/17

Adaptive search is my attempt to address an age-old problem when tracking dense packings with Crocker-Grier: how to select a `search_range` that does not exclude valid links, without creating many large subnets that make linking impractical or impossible. Conventionally, this is done by guessing, followed by trial, error, frustration, and resignation. However, when I was tracking 30k particles over 12k frames, it made no sense to carefully choose a single value for all particles in all frames — a `search_range` that worked perfectly well for the first 5500 frames or so would cause a `SubnetOversizeException` in frame 5501, because a corner of the image was momentarily displaced.

With adaptive search enabled, one instead specifies a *maximum* `search_range` for the entire movie. If an oversize subnet is encountered, linking becomes re-entrant: the particles of the offending subnet are re-linked with progressively smaller values of `search_range`. In this way, solvable (or trivial) subnets are broken off and solved one by one, until there are no more particles left to link. When it works, this is a huge simplification for the user: guess a reasonable maximum `search_range` based on the largest particle displacements, and leave the rest to `trackpy`.

Implementing this feature required a few big internal changes to the linker. The ones worth mentioning are

- Refactoring the linker into a class, making it relatively easy to recurse without passing around lots of parameters and state.
- Unifying `MAX_SUB_NET_SIZE` and its associated logic, and providing an alternate value more suited to the adaptive search algorithm.
- Properly testing handling of oversize subnets, and adaptive search.

This is a big one — it's taken roughly the past year to dream this up, tell @danielballan about it, implement it, use it in my research, rebase, rebase, rebase, and write the documentation. (Although most of the heavy lifting was done by @danielballan and @tacaswell when they wrote the linking tests.) Some serious scrutiny is needed, though if there are no major problems it would be good to target this for the v0.3 milestone. Thanks!


trackpy/linking.py
```diff
@@ -474,6 +476,12 @@ def link_df(features, search_range, memory=0,
         the next frame.
 
         For examples of how this works, see the "predict" module.
+    adaptive_step : float (optional)
```
tacaswell: This is picky, but I have been parsing numpydoc formatted strings recently and the convention is 

```
name : type, optional
```

trackpy/linking.py
```diff
@@ -744,139 +768,231 @@ def link_iter(levels, search_range, memory=0,
...
+                    # add the unmatched source particles to the new
+                    # memory set
+                    new_mem_set.add(sp)
+
+                # Clean up
+                if dp is not None:
+                    del dp.back_cands
+                if sp is not None:
+                    try:
+                        del sp.forward_cands
```
tacaswell: How do we end up with a source particle with out the forward candidate attribute?

trackpy/linking.py
```diff
@@ -744,139 +768,231 @@ def link_iter(levels, search_range, memory=0,
...
+                    # add points to the hash
+                    prev_hash.add_point(m)
+                    # re-create the forward_cands list
+                    m.forward_cands = []
+
+            prev_set = tmp_set
+
+            yield cur_level
+
+    def assign_links(self, dest_set, source_set, search_range,
```
tacaswell: Should either make a copy of the input sets or explicitly state in the docs that this will mutate the input.

tacaswell: As this is coming back to me, it needs to mutate the set as that is how memory works.  Should this be a private method ?

trackpy/linking.py
```diff
@@ -987,24 +1079,23 @@ def assign_candidates(cur_level, prev_hash, search_range, neighbor_strategy):
...
+    or increase :py:attr:`Linker.MAX_SUB_NET_SIZE`'''
     pass
 
 
-def recursive_linker_obj(s_sn, dest_size, search_range):
-    snl = sub_net_linker(s_sn, dest_size, search_range)
-    return zip(*snl.best_pairs)
+def recursive_linker_obj(s_sn, dest_size, search_range, max_size=30):
+    snl = sub_net_linker(s_sn, dest_size, search_range, max_size=max_size)
+    return [list(particles) for particles in zip(*snl.best_pairs)]
```
tacaswell: Why the conversion to lists here?

tacaswell:
Left some trivial-ish comments from reading the code.  My knee-jerk reaction is to be skeptical, but I am having trouble coming up with a good reason why this is worse than hand tuning.

trackpy/linking.py
nkeim: These collections have to be mutable, and in Python 3 they are not unless I do something like this.

trackpy/linking.py
nkeim: Not sure what you mean. Right now the `link()` method is called by the various API functions; it's the main point of entry for the algorithm. Maybe we should make this whole *class* private for now?

nkeim: Oops, you meant `assign_links()`. Yes, it probably should be private. Also added comments and a remark in the docstring about changes to the particles' candidate lists.

trackpy/linking.py
nkeim: I remember adding this for a reason, but it may have been temporary until I could fix a bug elsewhere. I can't see a reason for it now, and when I comment out the "try" the tests still pass. So I'm going to take it out and we'll hope for the best — the alternative is having it there forever :). Good catch!

nkeim:
Thanks, @tacaswell. I've just force-pushed a rebased branch that addresses your line comments. I'm glad you think that the example notebook gives users a sense of when adaptive search is and is not a good idea. But I'm not quite satisfied with that yet because there's presently no diagnostic info related to subnets or adaptive search — we're warning users that they can shoot themselves in the foot, but it's as though they can't even see where the darn thing is pointed. So as I mention in my comment to #184, some debug-level logging would go a long way here, and should be added before we merge.

nkeim:
Inserting @tacaswell 's https://github.com/soft-matter/trackpy/pull/184#issuecomment-69503525 into this discussion:

- Add an option (or at least a documented trick) to drop subnets entirely. I like this. It could just be a separate, null subnet solver that the user chooses with `link_strategy`.
- Let the linker return data on how adaptive search was used for each particle. This would be useful, but it also seems like it would be overly stretching the present API. I'll give it some thought.

danielballan:
+1 for `link_strategy='drop'`

nkeim:
Lots of new stuff in this rebase/push. Highlights:

- New `drop` link_strategy as requested. I realized that this can also be used to dry-run through a difficult linking job to scan for oversize subnets, before attempting the actual computation.
- With the `diag` keyword argument enabled, linking code can attach arbitrary bits of diagnostic information to each particle, returned as extra columns in the result DataFrame. A tutorial for this feature is in the works.
 - The only sane way to add the diagnostic info in `link_df` was to make a copy of the input DataFrame. This could break some poorly-written user code. So I added the `copy_features` option to turn on copying without diagnostics.
- Some minor performance enhancements, validated by asv benchmarking. When adaptive search and link diagnostics are turned *off*, the performance penalty due to this PR is zero, within error.
- Particle ID numbering no longer depends on ordering of a `set`.

Obviously we have greatly expanded the scope of the original PR. But the diagnostics were motivated by adaptive search, and they mess with some of the same pieces of code, so I decided to take the lazy approach and then get feedback.

trackpy/linking.py
```diff
@@ -515,13 +555,20 @@ def link_df(features, search_range, memory=0,
...
+                           diagnostics=diagnostics) for frame_no, frame
               in features.groupby(t_column))
     labeled_levels = link_iter(
         levels, search_range, memory=memory, predictor=predictor,
+        adaptive_step=adaptive_step, adaptive_limit=adaptive_limit,
         neighbor_strategy=neighbor_strategy, link_strategy=link_strategy,
         hash_size=hash_size, box_size=box_size)
 
+    if diagnostics:
+        features = strip_diagnostics(features)
```
danielballan: Could you add a comment noting that `strip_diagnostics` does not modify the original? I was worried for a moment that it did.

trackpy/linking.py
```diff
@@ -672,12 +736,48 @@ def link_df_iter(features, search_range, memory=0,
         yield features
 
 
-def _build_level(frame, pos_columns, t_column):
-    "Return IndexPointND objects for a DataFrame of points."
```
danielballan: IndexPointND objects -> PointND objects

trackpy/linking.py
```diff
@@ -735,146 +843,262 @@ def link_iter(levels, search_range, memory=0,
...
             if bc_c == 1:
                 # one backwards candidate
                 b_c_p = p.back_cands[0]
-                # and only one forward candidatege the repo
+                # and only one forward candidate
                 b_c_p_0 = b_c_p[0]
                 if len(b_c_p_0.forward_cands) == 1:
-                    # add to the track of the candidate
-                    b_c_p_0.track.add_point(p)
-                    _maybe_remove(mem_set, b_c_p_0)
```
danielballan: Where did this memory set removal logic go? (Sorry, I'm struggling to parse the diff here.)

trackpy/tests/test_link.py
```diff
@@ -383,142 +501,205 @@ def test_memory_removal(self):
...
         assert len(tracks) == p_count, len(tracks)
 
-    def link(self, *args, **kwargs):
-        kwargs.update(self.linker_opts)
-        return tp.link(*args, **kwargs)
 
-    def link_df(self, *args, **kwargs):
-        kwargs.update(self.linker_opts)
-        return tp.link_df(*args, **kwargs)
+class DiagnosticsTests(CommonTrackingTests):
```
danielballan: Neat use of a mixin.

trackpy/linking.py
```diff
@@ -735,146 +843,262 @@ def link_iter(levels, search_range, memory=0,
...
-            p.back_cands.sort(key=lambda x: x[1])
-        for p in prev_set:
-            p.forward_cands.sort(key=lambda x: x[1])
+            new_mem_set = set()
+            for sp, dp in zip(spl, dpl):
+                # Do linking
+                if sp is not None and dp is not None:
+                    sp.track.add_point(dp)
+                    if sp in self.mem_set:  # Very rare
+                        self.mem_set.remove(sp)
```
nkeim: @danielballan: The memory-removal logic from old line 871 ended up here during the adaptive search refactor.

I suppose I could move all the new stuff into a PR against the adaptive search branch of my repo. If you think it would make these diffs significantly less confusing, I'll do it.

nkeim:
Thanks @danielballan . Good catches. I'll fix those issues, and let me know whether you want the new stuff in its own PR.

Also on my to-do list before a (possible) merge:
- Replace `adaptive_limit` (number of recursion steps) with `adaptive_stop` (smallest acceptable `search_radius`). Makes more sense to think in terms of length scales.
- Draft a tutorial on diagnostics, so you don't have to squint at the source code so much.

trackpy/tests/test_link.py
```diff
@@ -58,6 +81,13 @@ def test_one_trivial_stepper(self):
         assert_frame_equal(actual, expected)
         actual_iter = self.link_df_iter(f, 5, hash_size=(10, 2))
         assert_frame_equal(actual_iter, expected)
+        if self.do_diagnostics:
+            assert 'diag_search_range' in self.diag.columns
+            print(self.diag.diag_search_range)
```
nkeim: Oops! Will remove this `print` and check for others.

trackpy/linking.py
danielballan: No, but thanks, this is good enough. Thanks for walking me through that.

danielballan:
Sounds good to me. Good idea with `adaptive_stop`. Best to stay in the physical world (as opposed to Algorithm World) wherever possible.

nkeim:
The promised changes are now up. Also, I wrote a new tutorial on linking diagnostics, and added diagnostics to the tutorial on adaptive search. See https://github.com/soft-matter/trackpy-examples/pull/17 . The 2 tutorials in question are:

- [linking diagnostics](http://nbviewer.ipython.org/github/nkeim/trackpy-examples/blob/adaptive-search/notebooks/linking-diagnostics.ipynb)
- [adaptive search](http://nbviewer.ipython.org/github/nkeim/trackpy-examples/blob/adaptive-search/notebooks/adaptive-search.ipynb)

danielballan:
The linking diagnostics tutorial is just so very cool. Wow.

I think it's time to merge this and the examples PR. Any final revisions?

tacaswell:
I gave these a read over a while ago and had nothing to say but 'cool!'

I can give a more thorough code review if you want, but time scale for that
is iffy

danielballan:
OK, I'm going for it. I know first-hand how busy you are. :- D

PR #195, opened 2015-01-14
FIX: Annotate mirrors image

caspervdw:
Something I encountered in PR #193, annotate displays a mirror image because y axis is not inverted, which is common practice in images. The rest of the functions do invert the y axis. Or is there a special reason for this?

### DIFF
--- a/trackpy/plots.py
+++ b/trackpy/plots.py
@@ -232,6 +232,7 @@ def annotate(centroids, image, circle_size=None, color=None,
         ax.imshow(image, **_imshow_style)
     ax.set_xlim(-0.5, image.shape[1] - 0.5)
     ax.set_ylim(-0.5, image.shape[0] - 0.5)
+    ax.invert_yaxis()
 
     if split_category is None:
         if np.size(color) > 1:
@@ -260,7 +261,6 @@ def annotate(centroids, image, circle_size=None, color=None,
         _plot_style.update(markeredgecolor=color[-1])
         ax.plot(centroids['x'][high], centroids['y'][high],
                 **_plot_style)
-    ax.invert_yaxis()
     return ax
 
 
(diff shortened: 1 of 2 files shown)


caspervdw:
Just to add to this: you can see the mirroring in the walkthrough:

trackpy/plots.py
```diff
@@ -260,6 +260,7 @@ def pairwise(iterable):
         _plot_style.update(markeredgecolor=color[-1])
         ax.plot(centroids['x'][high], centroids['y'][high],
                 **_plot_style)
+    ax.invert_yaxis()
```
tacaswell: I would put this up by line 233 which is what un-inverts the y-axis.

caspervdw: I just wanted to be sure that it is inverted, don't know what is happening in between the `if ... then` clause below line 233. But if you're sure that it's safe I will change it.

tacaswell: 95% sure, sometime the auto-scaling surprises me, but once the limits are set manually, the auto-scaling should be turned off.

danielballan: Yes, I think moving that is fine.

caspervdw: There it goes then

danielballan:
This is an artifact of early days, and I have been reluctant to change it, lest I should add to the confusion. I agree that we want the behavior to be consistent across the plotting functions, and now (0.3) is the time to make the changes.

tacaswell:
Do we have someplace to track api changes?  If not, this is probably a good time to start one...

nkeim:
Could API changes just go in the v0.3 release notes? Perhaps under a separate heading. We're going to have several worth mentioning.

danielballan:
That was my intention. I filled that our retrospectively, looking back through old commits. If people are willing to update the release notes in their PRs, that would be great. Some projects (see sckit-image) have really complicated ways to organizing that, and I didn't feel that was justified for us.

A separate "API changes" heading seems good. The most important thing is to communicate clearly how the results will change in 0.3.

caspervdw:
Like this?

tacaswell:
@caspervdw Yes.  This also matches the very old convention that for images (0, 0) is in the upper left (because that is where CRTs started scanning from and you store the image in the same order you needed to display it in)

PR #196, opened 2015-01-14
ENH: Generalize plotting to 2D/3D

caspervdw:
This implements 3D plotting of points (`scatter` / `scatter3d`) and trajectories (`plot_traj`, `plot_traj3d`). The approach I used slightly altered `plot_traj` to make it usable for both 2D and 3D. Then `plot_traj3d` directly calls `plot_traj`, with the diffence that the 3D version sets up 3D axes, and the 2D version sets up 2D axes. This could be extended to all other plotting functions, but I first would like some feedback on the approach.

The downside is that it makes `plot_traj` a bit more unclear, but I think altogether this is a better solution than making a new `plot_traj3d` that is more or less a copy of `plot_traj`.

To make this working, I added `_plot` that wraps `Axes.plot` to give it the same call signature for 2D
and 3D. For convenience, also `_set_labels` to set up all axis labels according to a string pattern.

This is also compatible with the mpl nbagg backend, so that you can rotate 3D graphs inline. But annotate3d isn't, so you would end up switching between backends.

### DIFF
--- a/trackpy/plots.py
+++ b/trackpy/plots.py
@@ -9,6 +9,7 @@ from functools import wraps
 import warnings
 
 import numpy as np
+from pandas import DataFrame
 
 from .utils import print_update
 
@@ -25,12 +26,15 @@ def make_axes(func):
               (Instant gratificaiton!)
     BUT:      If the uses passes axes to plotting function, write on those axes
               and return them. The user has the option to draw a more complex
-              plot in multiple steps.
+              plot in multiple steps. The axes have to be 2d or else the
+              current figure will be cleared.
     """
     @wraps(func)
     def wrapper(*args, **kwargs):
         import matplotlib.pyplot as plt
         if kwargs.get('ax') is None:
+            if hasattr(plt.gca(), 'zaxis'):
+                plt.clf()  # clear plot when plot is 3d
             kwargs['ax'] = plt.gca()
             # Delete legend keyword so remaining ones can be passed to plot().
             try:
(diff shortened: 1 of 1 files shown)


danielballan:
I think this should go in. There some changes I'd like to make to the way we handle plotting, but they apply to existing code as much as the additions here.

Any objections?

caspervdw:
I have made some progress on this so I would like to go through it once
more before merging. Can this wait until next week? I am at APS march
meeting right now.
Op 2 mrt. 2015 18:14 schreef "Dan Allan" <notifications@github.com>:

danielballan:
Of course. Enjoy that Texas weather. Wish I could have made it this year to
meet up.

caspervdw:
And I'm back! Unexpectedly the Dutch weather was better than the Texas weather ..
I added some functions:
- `plot_to_frame` renders a plot from a Figure or Axes object and outputs a `pims.Frame` containing an RGBA image, with given `dpi`
- `plots_to_frame` does the same for a list of figures, additionally calculates the dpi for the desired `width` and makes sure that each frame has the same size. This could be used by users to make rgb's from any list of Figures or Axes, for nice IPython display as scrollable stack or export to movie (probably using `pims.export` but pyav is not working on my computer)
- `annotate3d` is now based on these functions
- `plot_density_profile` makes a density profile, standard in z direction (= sedimentation profile)
- `plot_displacements` takes now arbitrary columns (todo: arrows in 3D)

Apparently, PIL is not a dependency. Do you have ideas for a workaround? PIL is a dependency of pims, which is a dependency of trackpy..

danielballan:
I see no problem making PIL/Pillow a dependency, but maybe it would be better to put `plot_to_frame` and `plots_to_frame`, whose utility is not limited to particle tracking, in pims. We can coordinate a pims release with the next trackpy release. Does this sound good?

caspervdw:
That sounds good. Also `pims.normalize` should be made public, I wanted to make use of that function from `annotate3d`, eventually I just copied the code. I will put this in one PR.

caspervdw:
This PR is all set now pims#140 is merged!

PR #199, opened 2015-01-19
List versions

danielballan:
This will be useful for bug reports:

    >>> import trackpy.diag
    >>> trackpy.diag.dependencies()
    {'scipy': '0.15.0', 'pytables': None, 'scikit-image': None, 'matplotlib': '1.4.2', 'pyfftw': 0.9.2,
    'numpy': '1.9.1', 'pyyaml': None, 'six': '1.9.0', 'numba': '0.16.0', 'pandas': '0.15.2',
    'python': '3.4.2'}

Any missing deps are mapped to `None` instead of a version string.

### DIFF
--- a/trackpy/diag.py
+++ b/trackpy/diag.py
@@ -2,6 +2,7 @@ from __future__ import (absolute_import, division, print_function,
                         unicode_literals)
 import six
 import sys
+from collections import OrderedDict
 
 from . import try_numba
 from . import preprocessing
@@ -34,7 +35,7 @@ def dependencies():
     """
     packages = ['six', 'numpy', 'scipy', 'matplotlib', 'pandas',
                 'scikit-image', 'pyyaml', 'pytables', 'numba', 'pyfftw']
-    result = dict()
+    result = OrderedDict()
     for package_name in packages:
         try:
             package = __import__(package_name)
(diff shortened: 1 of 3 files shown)


nkeim:
Nice! If we're requiring Python 2.7+ (are we?), you could use `collections.OrderedDict` to make the output more human-readable. Other than that, I'm for merging.

nkeim:
The irony of requiring a minimum version to find out what version you have just hit me.

danielballan:
Ha!

I have assumed we are 2.7+ and I have never heard from a 2.6 user of trackpy. I made it an `OrderedDict`. Merge at will.

trackpy/diag.py
```diff
@@ -20,3 +21,35 @@ def performance_report():
...
+    result : dict
+        mapping the name of each package to its version string or, if an
+        optional dependency is not installed, None
+    """
+    packages = ['six', 'numpy', 'scipy', 'matplotlib', 'pandas',
+                'scikit-image', 'pyyaml', 'pytables', 'numba', 'pyfftw']
+    result = OrderedDict()
+    for package_name in packages:
+        try:
+            package = __import__(package_name)
```
tacaswell: maybe switch out for `importlib` ?

danielballan: Good point. (To explain, we just learned this behaves better, though not sure why.)

Done. Tests pass and behavior looks the same.

PR #201, opened 2015-01-25
FIX Change in way conda is installed

caspervdw:
Travis started giving an error on `source activate testenv`. This should fix it.

quote from travis log:

    $ source activate testenv
    We're making conda environments even better!  Part of this process is
    changing the way environment activation works.  You need to install the
    new conda-env package before you can use these feature:
    conda install -c conda conda-env
    
    For more information, please visit: https://github.com/conda/conda-env


nkeim:
Hooray! I was just getting around to worrying about this. Thanks.

PR #203, opened 2015-01-31
PERF: Faster level-building

nkeim:
I was upset that the features I've been working on made linking several percent slower, so I went looking for ways to claw that back. It turns out that `DataFrame.iterrows()`, which was used by `link_df*`, creates a Series instance for each row (i.e. particle). This was responsible for more than half of the `real_data.DenseLinkingSuite.time_numba` benchmark. I don't know why we didn't notice this before; perhaps there was a change in Pandas.

Bypassing `iterrows()` reduces *all* of the linking benchmark times (and unit test times) by ~50%.

### DIFF
--- a/trackpy/linking.py
+++ b/trackpy/linking.py
@@ -674,10 +674,8 @@ def link_df_iter(features, search_range, memory=0,
 
 def _build_level(frame, pos_columns, t_column):
     "Return IndexPointND objects for a DataFrame of points."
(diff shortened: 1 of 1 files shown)


tacaswell:
:+1: from me!

danielballan:
Wow! I knew `iterrows` was bad news, but wow. Nice catch.

PR #204, opened 2015-02-02
PERF: Faster center of mass

caspervdw:
With the profiling I did for the anisotropic particle finding, I realized that the (pure python) `refine` relies heavily on `scipy.ndimage.center_of_mass`. This function seems to have some features which we are not using. Also it reinstanciates a `np.ogrid` every time it is used.

If I strip away the functions we don't use and reuse the same `np.ogrid` for every call, I get a ~60% performance increase for the center of mass call, which translates in some my tests in a 10%-20% computing time decrease of the total `locate`, using python engine.

Guess that #203 put me on this track!

### DIFF
--- a/trackpy/feature.py
+++ b/trackpy/feature.py
@@ -90,13 +90,12 @@ def estimate_size(image, radius, coord, estimated_mass):
     return Rg
 
 
-def _safe_center_of_mass(x, radius):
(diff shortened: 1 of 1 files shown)


tacaswell:
:+1: from me, leave it to dan/nathan to merge.

nkeim:
Fantastic!

Issue #198, opened 2015-01-19
Issue with PyFFTW


caspervdw:
Hi,

Yesterday I installed PyFFTW and it appears to work smoothly. But when trackpy uses scipy.ndimage.fourier_gaussian on the result of `fftn`, it fails with `datatype not understood`. While the datatype is the same as the numpy function `fftn`

This code reproduces my error:

    import numpy as np
    import trackpy as tp
    
    a = np.random.randint(0, 256, (50,50))
    
    afftw = tp.preprocessing.fftn(a)
    anp = np.fft.fftn(a)
    
    print type(afftw), afftw.dtype, afftw.shape
    print type(anp), anp.dtype, anp.shape  # equal to afftw
    
    tp.preprocessing.fourier_gaussian(anp, (1., 1.))  # no exception
    tp.preprocessing.fourier_gaussian(afftw, (1., 1.))  # exception

The stack:

    C:\Anaconda\lib\site-packages\scipy\ndimage\fourier.pyc in fourier_gaussian(input, sigma, n, axis, output)
        121         sigmas = sigmas.copy()
        122 
    --> 123     _nd_image.fourier_filter(input, sigmas, n, axis, output, 0)
        124     return return_value
        125 
    
    RuntimeError: data type not supported

danielballan:
I haven't seen that one before. I ran your code but I cannot reproduce the error. Which versions of pyfftw, numpy, and scipy? See #199, which I've been meaning to write for awhile, so save yourself a minute or two of trouble.

caspervdw:
My versions: (u'six', '1.8.0'), (u'numpy', '1.9.0'), (u'scipy', '0.14.0'), (u'matplotlib', u'1.4.0'), (u'pandas', '0.14.1'), (u'scikit-image', None), (u'pyyaml', None), (u'pytables', None), (u'numba', '0.14.0'), (u'pyfftw', '0.10.0'), (u'python', u'2.7.8')

I'll do some updating, maybe this solves the problem.

caspervdw:
Updating scipy to 0.15.1 had some effect: now it just hangs on `_nd_image.fourier_filter` instead of giving an error. The numpy-generated FFT just works.
A downgrade to pyfftw 0.9.2 did not make a difference.

danielballan:
Hmm. I won't have time to follow up for a few days. I would start by
inspecting attributes of the numpy array at various stages in the
process...before and after _maybe_align() and before and after the fft.

[...]

PR #206, opened 2015-02-06
FIX: For PyFFTW under windows 64bit

caspervdw:
This fix solves #198 by copying the FFT result to a new array.

### DIFF
--- a/trackpy/preprocessing.py
+++ b/trackpy/preprocessing.py
@@ -29,7 +29,9 @@ else:
                          "several seconds, but subsequent calls will run " +
                          "*much* faster.")
             planned = True
-        result = pyfftw.n_byte_align(a, a.dtype.alignment)
+        fft = pyfftw.n_byte_align(a, a.dtype.alignment)
+        result = np.empty_like(a, dtype=np.complex128)
+        result[:] = fft
         return result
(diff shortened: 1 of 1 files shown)


caspervdw:
OK I just realized that it is also implemented for `ifftn`. Apparently it doesn't raise any error, but the dtype is now also `np.complex128`, which shouldn't be the case.

trackpy/preprocessing.py
```diff
@@ -70,12 +74,12 @@ def bandpass(image, lshort, llong, threshold=None):
...
     axes = range(image.ndim)
     sizes = [x*2+1 for x in llong]
-    boxcar = np.asarray(image) 
-    for (axis,size) in zip(axes,sizes):
-        boxcar = uniform_filter1d(boxcar,size,axis,**settings)
-    gaussian = ifftn(fourier_gaussian(fftn(image), lshort))
+    boxcar = np.asarray(image)
+    for (axis, size) in zip(axes, sizes):
+        boxcar = uniform_filter1d(boxcar, size, axis, **settings)
+    gaussian = ifftn(fourier_gaussian(fftn(image), lshort)).real
```
danielballan: Good catch.

danielballan:
It's bummer that we have to take a perf hit to fix this. Let's leave #198 open and see if this can be fixed upstream in the future.

PR #209, opened 2015-02-17
FIX: Change datatype instead of copying array

caspervdw:
This also solves issue https://github.com/soft-matter/trackpy/issues/198, but faster.


danielballan:
Looks fine. I'm kicking Travis to start again.

danielballan:
Just to be safe, I'll wait for you to rebase this on master and check that Travis passes.

caspervdw:
Rebased

PR #210, opened 2015-02-17
Doc change

caspervdw:
This has the attribution of myself (as discussed before) and an example picture of 3D tracking.


nkeim:
Cool! But the image will be shown at about 1/2 linear size, so it is almost inscrutable in the context of the page. Is it possible to re-plot the data so that the 3D aspect can be understood at that size?

caspervdw:
Sorry for the late response; I had a busy week... I pushed a new plot without axes, I think it works better this way. The size is now 190x190px, as needed for the thumbnails.

nkeim:
Looks great! Merging with enthusiasm.

Issue #202, opened 2015-01-28
Idea for more tests


danielballan:
This might be the key to better test coverage for linking and subpixel feature-finding.

http://blog.jessitron.com/2014/09/tdd-is-dead-long-live-tdd.html

I can think of trends that generative tests could follow: larger particles should be located more precisely; dense images should require more subnets; etc. These flexibility of this kind of testing feels right for scientific applications where we understand many things about our data but struggle to write "actual == expected" style tests that don't break when the code is improved.

PR #212, opened 2015-02-22
BUG: Consistently add subnet distances in quadrature

nkeim:
According to the original [Crocker & Grier paper](http://dx.doi.org/10.1006/jcis.1996.0217), and mathy reasons, distances in n-D space are added in quadrature to make an (inverse) figure of merit for evaluating subnet solutions. This has been the practice of the numba linker basically for as long as it's been the default, so I doubt there's been any issue for the vast majority of our users. However, it was not done by the pure-Python linkers (recursive and non-recursive). (@tacaswell: Was this related to your original application for this code?)

This PR adds a simple test that discerns whether a subnet linker uses quadrature or not. The numba linker passes but the others fail. The Python linkers are then changed so they work in quadrature and pass the test.

There are also a few minor fixes thrown in for good measure.

**NOTE**: The Travis builds have been failing on my repo due to a conda error; the tests pass locally for me in py2.7 and py3. The problem does not appear to be in the conda issue tracker yet, but I really don't know enough about our use of conda to articulate what is going wrong. I'd appreciate some debugging, or perhaps some guidance on what to report over there.

### DIFF
--- a/trackpy/linking.py
+++ b/trackpy/linking.py
@@ -1424,7 +1424,8 @@ def _numba_subnet_norecur(ncands, candsarray, dists2array, cur_assignments,
 
     cur_assignments, tmp_assignments are just temporary registers of length nj.
     best_assignments is modified in place.
-    Returns the best sum.
+    Returns the number of assignments tested (at all levels). This is basically
+    proportional to time spent.
     """
     nj = candsarray.shape[0]
     tmp_sum = 0.
--- a/trackpy/linking.py
+++ b/trackpy/linking.py
@@ -1232,7 +1232,7 @@ class SubnetLinker(object):
     def do_recur(self, j):
         cur_s = self.s_lst[j]
         for cur_d, dist in cur_s.forward_cands:
-            tmp_sum = self.cur_sum + dist
+            tmp_sum = self.cur_sum + dist**2
             if tmp_sum > self.best_sum:
                 # if we are already greater than the best sum, bail we
                 # can bail all the way out of this branch because all
@@ -1255,7 +1255,8 @@ class SubnetLinker(object):
             # if we have hit the end of s_lst and made it this far, it
             # must be a better linking so save it.
             if j + 1 == self.MAX:
-                tmp_sum = self.cur_sum + self.search_range * (self.max_links - len(self.d_taken))
+                tmp_sum = self.cur_sum + self.search_range**2 * (
+                    self.max_links - len(self.d_taken))
                 if tmp_sum < self.best_sum:
                     self.best_sum = tmp_sum
                     self.best_pairs = list(self.cur_pairs)
@@ -1263,7 +1264,7 @@ class SubnetLinker(object):
                 # re curse!
                 self.do_recur(j + 1)
             # remove this step from the working
-            self.cur_sum -= dist
+            self.cur_sum -= dist**2
             if cur_d is not None:
                 self.d_taken.remove(cur_d)
             self.cur_pairs.pop()
@@ -1298,7 +1299,8 @@ def nonrecursive_link(source_list, dest_size, search_range, max_size=30, diag=Fa
             # base case, no more source candidates,
             # save the current configuration if it's better than the current max
             # add penalty for not linking to particles in the destination set
-            tmp_sum = cur_sum + search_range * (max_links - len([d for d in cur_back if d is not None]))
+            tmp_sum = cur_sum + search_range**2 * (
+                max_links - len([d for d in cur_back if d is not None]))
             if tmp_sum < best_sum:
                 best_sum = cur_sum
                 best_back = list(cur_back)
@@ -1328,7 +1330,7 @@ def nonrecursive_link(source_list, dest_size, search_range, max_size=30, diag=Fa
         # get the forward candidate
         cur_d, cur_dist = cand_list_list[j][k]
 
-        tmp_sum = cur_sum + cur_dist
+        tmp_sum = cur_sum + cur_dist**2
         if tmp_sum > best_sum:
             # nothing in this branch can do better than the current best
             j -= 1
--- a/trackpy/linking.py
+++ b/trackpy/linking.py
@@ -1173,11 +1173,8 @@ def assign_candidates(cur_level, prev_hash, search_range, neighbor_strategy):
     if neighbor_strategy == 'BTree':
         # (Tom's code)
         for p in cur_level:
-            # get
             work_box = prev_hash.get_region(p, search_range)
             for wp in work_box:
-                # this should get changed to deal with squared values
-                # to save an eventually square root
                 d = p.distance(wp)
                 if d < search_range:
                     p.back_cands.append((wp, d))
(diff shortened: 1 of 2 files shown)


tacaswell:
In the original use case I wrote for I was doing linking in 1D so it didn't matter.

Looks like this is indeed a bug in my original code, that is devilishly subtle.

nkeim:
Ah, that makes sense! I have used trackpy in 1D and was delighted that it worked as expected.

I was pretty dumbstruck myself that this got past our existing unit tests. Perhaps this would have been caught by the approach in #202 , since we have multiple implementations that should all produce the same result.

tacaswell:
This prompted me to go back and look and I got it right in my c++ :smile: 

I have some memory of wanting to push the computation of the distance squared into the `Point` classes, but looking at the code does not support this.

How big of a surface area does this bug have?  Does it only matter in pathological cases?

tacaswell:
How did you find this?

nkeim:
I was hoping you'd ask. :smile: I was trying to illustrate the algorithm for educational purposes, so I tried to come up with the simplest non-obvious subnet, and then show that you could verify trackpy's solution by hand (which now strikes me as a quintessentially Sid thing to do). I hadn't looked at Crocker-Grier or any of the subnet code in awhile, so I wasn't even sure how the math was done. Once I realized that my subnet was a special case, the rest fell into place. 

I suspect that in real-world use the surface area of this bug is small, because most subnets can be solved by visual inspection, without resort to numbers. (And because most people have used the numba linker for awhile now.) But I imagine it would introduce a small error into any sufficiently large microrheology dataset.

tacaswell:
I need to check my code a bit more before sharing, but trying to answer this question via brute force saying that it is only special points where the ordering disagrees.

danielballan:
How much of a performance gain/hit does this correction involve?

tacaswell:
Good thing thing I checked, had bug with python order of operations evaluation, but

```python
import numpy as np
import matplotlib.pyplot as plt

pt0_0 = np.array([10, 11])
pt0_1 = np.array([10, 8])
pt1_0 = np.array([12, 10])

d00 = np.sqrt(np.sum((pt0_0 - pt1_0) ** 2))
d10 = np.sqrt(np.sum((pt0_1 - pt1_0) ** 2))


nx, ny = 3000, 2001
x = np.linspace(7.9, 12.1, nx)
y = np.linspace(7.9, 11.1, ny)
X, Y = np.meshgrid(x, y)

D01 = np.sqrt((X - pt0_0[0])**2 + (Y - pt0_0[1])**2)
D11 = np.sqrt((X - pt0_1[0])**2 + (Y - pt0_1[1])**2)

fail_map = ((d10 + D01) < (d00 + D11)) != ((d10**2 + D01**2) < (d00**2 + D11**2))

fig, ax = plt.subplots(figsize=(6, 4))
im = ax.imshow(fail_map, extent=[x[0], x[-1], y[0], y[-1]], cmap='PRGn', origin='bottom')
ax.set_aspect('equal')
ax.plot([pt0_0[0]], [pt0_0[1]], 'ws', ms=5)
ax.plot([pt0_1[0]], [pt0_1[1]], 'ws', ms=5)

ax.plot([pt1_0[0]], [pt1_0[1]], 'w^', ms=5)
ax.plot([9], [10], 'wx', ms=5)
ax.set_xlim(x[0], x[-1])
ax.set_ylim(y[0], y[-1])
fig.colorbar(im)
```

so

Green is where the linking will differ.  This is ~4% of the area shown.

Not immediately clear to me how this will scale with the absolute size of the separations.  I also don't understand the sharp top edge of the failing region.

tacaswell:
If it isn't clear the squares markers are the points in the first frame, the triangle is the fixed point is the second frame.  The X is the failing point in the tests and the point that I am moving around.  The color at each coordinate indicates if the linking will be the same (purple) or different (green).

nkeim:
Thanks; it's good to bring up performance early and often. There will be zero change for the numba code. The Python linkers should get slightly slower because of all the squaring they have to do. If we squared the distances once and saved the result, this correction would incur a smaller cost. But there is a larger question of our policy on performance. My own take is that there are very few users who both care acutely about performance (and so will feel the pain of a few percent slowdown), and cannot use numba. But I could easily be wrong. To me, the Python linkers serve as fallbacks with reasonable performance, but also as reference implementations that are much easier to read and debug than the ugly numba version.

If we all think that way and we wanted to have a project-wide guideline for performance, it might be:

"If at all possible, PRs should not slow down 

* 2D or 3D feature-finding (i.e. both Python and numba)
* Linking with cKDTree + numba, for 'production' settings (e.g. no diagnostics). 

Everywhere else, it's acceptable to lose performance in proportion with other gains."

With "slow down" as determined by the applicable asv benchmarks (we need 3D!). But you'll notice that I am making up policy on the spot to avoid doing work. So we should either discuss this idea or ignore it altogether!

nkeim:
Cool. So linking is surprisingly robust to doing it the "wrong" way.

I think that since the algorithm is scale-free (so long as all particles are being linked), this result would be too. The sharp edge is indeed weird, but I can't think of any symmetries that would forbid it.

tacaswell:
The line is at 1 unit in the y-direction which is where  `x**2 == x` which seems like it should matter...

nkeim:
The straight (curved) line is where the quadratic (linear) method switches results. The difference in squared (or linear) lengths is what decides this: Above each of those boundaries, it costs less to link the left point with the top one (and the right point with the bottom one). Below the boundary, it costs less to connect the left point with the bottom one. 

Along the straight line, you'll find that the *difference* in the squared distances to the top and bottom points is constant; the equivalent is true of the curve, which makes it a hyperbola. 

The boundaries have the orientation that they do because of symmetry. You can convince yourself that they should have that *position* by moving the free point over close to where the fixed point is. So the distance of 1 unit is a coincidence.

nkeim:
Thanks to @danielballan for the fix. But I think there is still a bug lurking somewhere in the subnet code, so please wait to merge.

nkeim:
False alarm; the algorithms just break ties in different ways (which I really don't think we should care about). The code appears to be correct, though the Travis build is still failing (darn) and there is still @danielballan 's concern about performance.

danielballan:
I spent a couple hours yesterday fighting with Travis. Some progress, but
we are not home free yet. Soon!

nkeim:
Sounds like quite a slog — thanks. We can't be the only ones with this problem. Maybe it's worth waiting a few days?

danielballan:
Haha, yes, I think I've done my piece. If a solution doesn't present itself soon, I will invoke my BNL-funded "priority conda support" powers and see if ContinuumIO will fix this. Trackpy is a dependency of our X-ray analysis stack.

nkeim:
Follow-up: Comparison of the output of the recursive linker with the output of the Blair-Dufresne MATLAB code, for a dense packing, shows that the new behavior agrees, and the previous behavior did not.

danielballan:
That's good news.

tacaswell:
:+1: on merging this.  Sorry for the original bug.

danielballan:
Tests look good. @nkeim can you rebase to make sure Travis is happy?

nkeim:
Done. Looks ready to merge.

PR #214, opened 2015-03-02
CRITICAL: Prevent pims namespace conflicts

nkeim:
It appears that `pims` now has a `utils` module, and the last line in `api.py` is `from pims import *`. Lots of tests were failing because depending on how `trackpy` was being imported, it was possible to end up with the pims utilities instead of the trackpy ones. 

I was in the habit of not running the full test suite locally, because I knew Travis would take care of that. This issue did not affect the tests I *was* running. Since Travis has been failing for other reasons, that's a very bad habit!

Unless I'm insane, please merge ASAP.

### DIFF
--- a/trackpy/filtering.py
+++ b/trackpy/filtering.py
@@ -7,14 +7,14 @@ by wrapping pandas group-by and filter capabilities."""
 
 
 def filter_stubs(tracks, threshold=100):
-    """Filter out trajectories with few points. They are often specious.
+    """Filter out trajectories with few points. They are often spurious.
 
     Parameters
     ----------
     tracks : DataFrame
         must include columns named 'frame' and 'particle'
     threshold : integer, default 100
-        minimum number of points to survive
+        minimum number of points (video frames) to survive
 
     Returns
     -------
--- a/trackpy/linking.py
+++ b/trackpy/linking.py
@@ -14,11 +14,6 @@ import pandas as pd
 from .utils import print_update
 from .try_numba import try_numba_autojit, NUMBA_AVAILABLE
 
-__all__ = ['HashTable', 'TreeFinder', 'Point', 'PointND',
-           'Track', 'TrackUnstored', 'UnknownLinkingError',
-           'SubnetOversizeException', 'link', 'link_df', 'link_iter',
-           'link_df_iter', 'strip_diagnostics']
-
 
 class TreeFinder(object):
 
--- a/trackpy/api.py
+++ b/trackpy/api.py
@@ -1,6 +1,11 @@
 from __future__ import (absolute_import, division, print_function,
                         unicode_literals)
 import six
+
+# Import all of pims top-level for convenience. But do it first so
+# that trackpy names take precedence.
+from pims import *
+
 from .motion import *
 from .plots import *
 from .linking import *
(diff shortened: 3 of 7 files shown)


tacaswell:
I think a better solution is to do in the `from foo import *` args.

Travis is still failing because it is trying to build scipy?!?

danielballan:
@tacaswell Do you mean `from pims import ImageSequence, Tiffstack, ... # etc.`? I'm not sure what you mean by "args."

We could also consider dropping the global pims import in the future. It is there for backward compatibility with mr, which included the classes that were later split off into pims. It saves a negligible amount of typing, and it might confuse users into thinking that trackpy *requires* images to be loaded using these particular routines. Keeping pims readers out of trackpy makes it clearer that trackpy does not care how you do IO.

The best argument against this change is that it would break *a lot* of user code, albiet in a way that's quite simple to fix. I propose explicitly importing most of pims in 0.3 with a warning that we may drop it in a future release.

tacaswell:
Args should be calls.

nkeim:
I like Dan's compromise plan, to change pims to explicit imports. The only advantage of deprecating it that I can see is that if someone has a problem with `ImageSequence`, they'd know to look at the pims docs and not trackpy. OTOH, one of the slight annoyances of scientific Python is the sheer number of boxes that a new user has to open and poke around in.

@tacaswell , I'm also confused about what you're saying. By "calls" do you mean the names of callable objects, as in Dan's example?

While we're at it, it would be simple to move things from the `__all__` within each module into `api.py`, though I'm not sure it's a clear win.

tacaswell:
I am saying to remove the `*` and do explicit imports. 

One of the great pains in mpl is pylab which greatly confuses the user as to where various things are coming from (and we get a numpy bug report every so often).

nkeim:
OK, now when the user tries to use one of the 3 "historic" pims classes, it works, but they get something like this:

```
/Users/nkeim/projects/trackpy/trackpy/trackpy/api.py:37: UserWarning: trackpy.ImageSequence is being called, but "ImageSequence" is really part of the pims module. It will not be in future versions of trackpy. Consider importing pims and calling pims.ImageSequence instead.
  ).format(call.__name__), UserWarning)
```

Overkill?

nkeim:
I meant "pims package," not "pims module." Will amend and force-push.

danielballan:
This is exactly what I had in mind.

nkeim:
Great! I've moved the contents of `__all__`, etc. to `api.py` as discussed; there are no more `import *`. FWIW, the walkthrough notebook still runs as before.

trackpy/filtering.py
```diff
@@ -7,14 +7,14 @@
 
 
 def filter_stubs(tracks, threshold=100):
-    """Filter out trajectories with few points. They are often specious.
```
danielballan: Ha. Embarrassing.

trackpy/linking.py
```diff
@@ -1423,7 +1418,8 @@ def numba_link(s_sn, dest_size, search_radius, max_size=30, diag=False):
     dest_results = [dcands[i] if i >= 0 else None for i in best_assignments]
     return source_results, dest_results
 
-@try_numba_autojit
+import numba
+@numba.autojit(nopython=True)
```
danielballan: Was this accidentally committed? This would definitely break without numba installed.

nkeim: Yikes!!! Yes. Happy to make my share of stupid mistakes. Will fix.

nkeim:
Python 3 unit test is failing because Python 3 generates its own extra warning. Figures...

PR #187, opened 2014-12-17
BLD: Add a conda recipe that builds binary package from current git rev.

danielballan:
This is the "conda recipe" I have been using to create the conda packages in trackpy's development channel. By including them in the repo itself, it will be possible to make Travis build conda packages and upload them to binstar. ([Others](https://github.com/rmcgibbo/python-appveyor-conda-example#binstar) have already figured out the details of how to do this.)

FYI, the dev conda packages are up to date. Anyone who wants to take the improved 3D capabilities for a spin can easily [install the latest development version](http://soft-matter.github.io/trackpy/installation.html#latest-version-under-development). attn: @mmdriscoll

PR #217, opened 2015-03-03
PERF: Support for numba 0.16+

nkeim:
Due to a refactoring of the numba backend starting with 0.16 (they are now on 0.17), the numba feature-finding code could no longer be compiled to LLVM; instead it was running slower than pure Python. To fix it, I had to remove all array allocation from the function. The result should now run even faster than before, with all compatible versions of numba.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,17 +1,17 @@
 language: python
 
 matrix:
-  include: 
+  include:
+    # "Legacy" environments: oldest supported versions, without and with numba
     - python: "2.7"
-      env: DEPS="numpy=1.7.1 scipy=0.13.0 matplotlib=1.3 pandas=0.13.0 scikit-image=0.9 pyyaml pytables"
+      env: DEPS="numpy=1.7.1 scipy=0.12.0 matplotlib=1.3 pandas=0.12.0 scikit-image=0.9 pyyaml pytables"
     - python: "2.7"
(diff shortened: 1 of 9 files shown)


danielballan:
Thanks for doing this.

I think we should put `nopython=True` into the codebase and drop support for older versions of numba. I just can't imagine how this could realistically hurt our users, and it's a useful assurance that our numba code will stay fast.

nkeim:
That's a valid position. It just seems so weird; we'd lose compatibility with a version of numba released just one year ago. But I guess the time to do it is at v0.3; it offers a lot of improvements in exchange for some potential inconvenience.

To restore `nopython=True`, I think we'd want to 

- require numba 0.13 or later
- recommend 0.17 in the docs
- Tell Travis to test on 0.13 and the latest version

I actually have no idea how to do the first one — I can't even find `meta.yaml` (is it in the repo?).

Would anyone else care to chime in on an appropriate life-cycle for dependencies?

danielballan:
The conda recipe with `meta.yaml` is in #187.

nkeim:
Is the idea to ultimately have the conda stuff in master, or keep it in that branch?

nkeim:
I updated numba versions and added a conda recipe as discussed. I also refreshed the other dependency versions used by Travis to be a little more current.

danielballan:
Yes, the idea was to put it in master. Tagged releases should be built using a modified recipe that hard-codes the version and includes the checksum, but this recipe is intended for dev builds.

danielballan:
This looks good to me if there are no objections to dropping numba < v0.13.4.

Issue #220, opened 2015-03-05
predict reset_index stack trace


apiszcz:
1 Stack trace
2 Sample code

I am trying to follow the example from: http://soft-matter.github.io/trackpy/tutorial/prediction.html

Any thoughts appreciated, not sure what reset_index thinks f is  a string.

== 1 ==========================
Traceback (most recent call last):
  File "lib\testpre\trackpylearn.py", line 21, in <module>
    tr=pd.concat(pred.link_df_iter(f, 1.5))
  File "c:\p\testpre\lib\pandas\tools\merge.py", line 724, in concat
    copy=copy)
  File "c:\p\testpre\lib\pandas\tools\merge.py", line 755, in __init__
    objs = [obj for obj in objs if obj is not None ]
  File "c:\p\testpre\lib\trackpy\predict.py", line 52, in link_df_iter
    for frame in linking.link_df_iter(*args, **kw):
  File "c:\p\testpre\lib\trackpy\linking.py", line 696, in link_df_iter
    labeled_levels, features_forpost, index_iter):
  File "c:\p\testpre\lib\trackpy\linking.py", line 919, in link
    prev_level = next(level_iter)
  File "c:\p\testpre\lib\trackpy\linking.py", line 683, in <genexpr>
    levels = (_build_level(frame, pos_columns, t_column, diagnostics=diagnostics)
  File "c:\p\testpre\lib\trackpy\linking.py", line 681, in <genexpr>
    (frame.reset_index(drop=True) for frame in features_for_reset))
AttributeError: 'str' object has no attribute 'reset_index'

== 2 ==========================
import os,pdb,pprint,sys
sys.path.insert(0,os.path.abspath('lib'))
import numpy as np
import pandas as pd
from pdb import set_trace as dd
import trackpy as tp
import trackpy.predict

f=pd.DataFrame(columns=['x','y','frame'])
f=f.append(({'x':1,'y':1,'frame':0}),ignore_index=True)
f=f.append(({'x':2,'y':2,'frame':1}),ignore_index=True)
f=f.append(({'x':3,'y':3,'frame':2}),ignore_index=True)
f=f.append(({'x':4,'y':4,'frame':3}),ignore_index=True)
f=f.append(({'x':10,'y':10,'frame':0}),ignore_index=True)
f=f.append(({'x':20,'y':20,'frame':1}),ignore_index=True)
f=f.append(({'x':30,'y':30,'frame':2}),ignore_index=True)
[...]

PR #221, opened 2015-03-05
BUG/API: Reduce link_df[_iter]() confusion

nkeim:
This addresses https://github.com/soft-matter/trackpy/issues/220 :

1. Useful error message when `link_df_iter()` is passed a single DataFrame
2. Predictors now have `link_df()` methods. It is actually a wrapper for `link_df_iter()` and so does *not* have the default in-place behavior that `link_df()` does.
3. It turned out that `link_df_iter()`'s implementation of the `t_column` option was not being tested and did not actually work. Fixed.

My solution on the second point is not ideal. The docstring clearly states the limitation of my solution, and I think that overall it's more helpful to have a drop-in replacement that works for most users, than to require those users to learn some Pandas tricks if they want to try out prediction on their data. I concluded that implementing a *true* `link_df()` wrapper would require us to either (i) add special prediction code to `link_df()`, (ii) seriously refactor it, or (iii) copy its code into `predict.py`.

### DIFF
--- a/trackpy/linking.py
+++ b/trackpy/linking.py
@@ -674,8 +674,9 @@ def link_df_iter(features, search_range, memory=0,
 
     # Non-destructively check the type of the first item of features
     feature_iter, feature_checktype_iter = itertools.tee(iter(features))
-    if not isinstance(next(itertools.islice(feature_checktype_iter, 1)),
-                      pd.DataFrame):
+    try:  # If it quacks like a DataFrame...
+        next(itertools.islice(feature_checktype_iter, 1)).reset_index()
+    except AttributeError:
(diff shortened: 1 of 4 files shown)


danielballan:
I'm sold. This is a clear improvement and the right compromise. If we hear from a lot of folks making use of prediction and getting confused by this limitation, we can consider revisiting. I leave it to @tacaswell or @caspervdw to sign off and merge.

caspervdw:
I am not enough into linking (yet) to merge this, so I leave it to @tacaswell

danielballan:
Paging @tacaswell :- )

tacaswell:
This is on my to-do list for tomorrow.

Issue #205, opened 2015-02-06
Selling trackpy on the homepage


nkeim:
trackpy has many important strengths, and we pride ourselves on the overall quality and design of the code. However, there is a subset of those strengths that one might call "killer" features — ones that are sufficiently impressive that they could grab attention and/or convince someone to switch from an already-working solution. I can think of a few possible examples:

- Running on Python (but that's hard to miss)
- Anisotropic 3D support, which will be a major improvement in v0.3
- Handling of arbitrarily long movies in many formats
- Attractive built-in plotting
- More?

When one Googles "trackpy," the top result is the GitHub page. If we start the stopwatch when someone clicks on that link, how long until they discover one or more of these features? What could we do to shorten this time?

- Should these marquee features be listed on the documentation front page? The GitHub README?
- Should one of the banner images on the documentation front page represent 3D data?
- Should there be a tutorial on anisotropic 3D features? ( @caspervdw ?)

This was prompted by a person who, after looking briefly at the documentation, had the impression that trackpy does not support 3D.

caspervdw:
Good that you mention it, I will finish the tutorial this weekend.

nkeim:
As far as I'm concerned, this issue was adequately addressed by @caspervdw . Onward to v0.3!

PR #224, opened 2015-03-19
DOC: Update README to advertise a little more

nkeim:
This addresses #205 by putting "3D" on the first page that a Google searcher will see. It should be merged *after* https://github.com/soft-matter/trackpy-examples/pull/18 is merged and the gh-pages are updated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@ trackpy
 What is it?
 -----------
 
-**trackpy** is a Python package providing tools for 2D and 3D particle tracking.
+**trackpy** is a Python package for particle tracking in 2D, 3D, and higher dimensions.
 **[Read the walkthrough](http://soft-matter.github.io/trackpy/tutorial/walkthrough.html)** to skim or study an example project from start to finish.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -6,17 +6,20 @@ trackpy
 What is it?
 -----------
 
-**trackpy** is a Python package providing tools for particle tracking.
+**trackpy** is a Python package providing tools for 2D and 3D particle tracking.
```
tacaswell: It really can do ND...

nkeim: Yeah, I was thinking about that. "3D" needs to appear for… well, for SEO. But I should probably add "nD" to the list.

tacaswell: How about "2D, 3D, and higher dimensions".

tacaswell: Er, "...provides particle tracking in 2D, 3D, and higher dimensions."

danielballan:
I updated the docs. (The updated includes the minor PRs I just submitted to trackpy and trackpy-examples.)

PR #229, opened 2015-03-23
ENH: Artificial feature images

caspervdw:
I combined some work that I did on the creation of artificial feature images. This is more flexible than the functions in `test_feature.py`, enabling the user to write any radial function and generate artificial images based on that.

A new file `artificial.py` is created so that users can use `tp.artificial.draw_spots` to make their own tests. I tried to make the documentation in the functions complete.

I changed `test_feature.py` to make use of these functions. The changes are:
 - `draw_gaussian_spot` needed the radius of gyration as a parameter. The new `draw_feature` defaults to a gaussian, but you need to supply the box size on which is drawn. By default, this is 6x the radius of gyration. 
 - `test_feature.compare` is now protected against features that overlap. Incidentally, this could go wrong before.
 - `test_rg` was going right coincidentally all along. See commit message https://github.com/soft-matter/trackpy/commit/75877a4dd39b27253dc78afa6e1f85597ff48fec for a further clarification

### DIFF
--- a/trackpy/artificial.py
+++ b/trackpy/artificial.py
@@ -77,7 +77,7 @@ def draw_feature(image, position, diameter, max_value=None,
             raise ValueError("Diameter is already anisotropic; eccentricity is"
                              " not defined.")
         diameter = (diameter[0] / (1 - ecc), diameter[1] * (1 - ecc))
-    radius = tuple([(d - 1) / 2 for d in diameter])
+    radius = tuple([d / 2 for d in diameter])
     if max_value is None:
         max_value = np.iinfo(image.dtype).max - 3
     rect = []
--- a/trackpy/artificial.py
+++ b/trackpy/artificial.py
@@ -1,0 +1,180 @@
+from __future__ import (absolute_import, division, print_function,
+                        unicode_literals)
+import six
+import numpy as np
+from scipy.spatial import cKDTree
+from trackpy.utils import validate_tuple
+
+
+def draw_point(image, pos, value):
+    image[tuple(pos)] = value
+
+
+def feat_gauss(r, rg=0.333):
+    """ Gaussian at r = 0 with max value of 1. Its radius of gyration is
+    given by rg. """
+    return np.exp((r/rg)**2 * r.ndim/-2)
+
+
+def feat_gauss_edge(r, value_at_edge=0.1):
+    """ Gaussian at r = 0 with max value of 1. Its value at r = 1 is given by
+    value_at_edge. """
+    return np.exp(np.log(value_at_edge)*r**2)
+
+
+def feat_ring(r, r_at_max, value_at_edge=0.1):
+    """ Ring feature with a gaussian profile, centered at r_at_max. Its value
+    at r = 1 is given by value_at_edge."""
+    return np.exp(np.log(value_at_edge)*((r - r_at_max) / (1 - r_at_max))**2)
+
+
+def feat_hat(r, disc_size, value_at_edge=0.1):
+    """ Solid disc of size disc_size, with Gaussian smoothed borders. """
+    mask = r > disc_size
+    spot = (~mask).astype(r.dtype)
+    spot[mask] = feat_ring(r[mask], disc_size, value_at_edge)
+    spot[~mask] = 1
+    return spot
+
+
+def feat_step(r):
+    """ Solid disc. """
+    return r <= 1
+
+
+def draw_feature(image, position, diameter, max_value=None,
+                 feat_func=feat_gauss, ecc=None, **kwargs):
+    """ Draws a radial symmetric feature and adds it to the image at given
+    position.
+
+    Parameters
+    ----------
+    image : ndarray
+        image to draw features on
+    position : iterable
+        coordinates of feature position
+    diameter : number
+        defines the box that will be drawn on
+    max_value : number
+        maximum feature value. should be much less than the max value of the
+        image dtype, to avoid pixel wrapping at overlapping features
+    feat_func : function. Default: feat_gauss
+        function f(r) that takes an ndarray of radius values
+        and returns intensity values <= 1
+    ecc : positive number, optional
+        eccentricity of feature, defined only in 2D. Identical to setting
+        diameter to (diameter / (1 - ecc), diameter * (1 - ecc))
+    kwargs : keyword arguments are passed to feat_func
+    """
+    if len(position) != image.ndim:
+        raise ValueError("Number of position coordinates should match image"
+                         " dimensionality.")
+    diameter = validate_tuple(diameter, image.ndim)
+    if ecc is not None:
+        if len(diameter) != 2:
+            raise ValueError("Eccentricity is only defined in 2 dimensions")
+        if diameter[0] != diameter[1]:
+            raise ValueError("Diameter is already anisotropic; eccentricity is"
+                             " not defined.")
+        diameter = (diameter[0] / (1 - ecc), diameter[1] * (1 - ecc))
+    radius = tuple([(d - 1) / 2 for d in diameter])
+    if max_value is None:
+        max_value = np.iinfo(image.dtype).max - 3
+    rect = []
+    vectors = []
+    for (c, r, lim) in zip(position, radius, image.shape):
+        if (c >= lim) or (c < 0):
+            raise ValueError("Position outside of image.")
+        lower_bound = max(int(np.floor(c - r)), 0)
+        upper_bound = min(int(np.ceil(c + r + 1)), lim)
+        rect.append(slice(lower_bound, upper_bound))
+        vectors.append(np.arange(lower_bound - c, upper_bound - c) / r)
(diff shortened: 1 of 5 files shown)


trackpy/artificial.py
```diff
@@ -0,0 +1,180 @@
...
+        to_drop = []
+        for pair in duplicates:
+            to_drop.append(pair[1])
+        f = np.delete(f, to_drop, 0)
+    return f * separation
+
+
+def gen_nonoverlapping_locations(shape, count, separation, margin=0):
+    """ Generates `count` number of positions within `shape`, that have minimum
+    distance `separatin` from each other. The number of positions returned may
```
nkeim: `separation`

trackpy/tests/test_feature.py
```diff
@@ -464,43 +413,35 @@ def test_rg(self):
...
-        actual = tp.locate(image, 7, 1, preprocess=False,
-                           engine=self.engine)['size']
-        expected = SIZE
-        assert_allclose(actual, expected, rtol=0.1)
 
         SIZE = 3
         image = np.ones(dims, dtype='uint8')
-        draw_gaussian_spot(image, pos, SIZE)
-        actual = tp.locate(image, 11, 1, preprocess=False,
+        draw_feature(image, pos, (SIZE*2 + 1) * 3, rg=0.333)
```
nkeim: It would really help to explain here that `rg=0.333` is dimensionless, and that its value in pixels is meant to be `SIZE`.

nkeim:
This is a good idea and I like the code! But the *arithmetic* difference of 0.2 is significant and confusing. Maybe it's because the scale factor is `(SIZE * 2 + 1) * 3` and not `(SIZE * 2) * 3`?

caspervdw:
I got quite confused, but I think there are 2 things at play here. I will put them inside line notes at the point they are relevant.

trackpy/artificial.py
```diff
@@ -0,0 +1,180 @@
...
+
+
+def draw_point(image, pos, value):
+    image[tuple(pos)] = value
+
+
+def feat_gauss(r, rg=0.333):
+    """ Gaussian at r = 0 with max value of 1. Its radius of gyration is
+    given by rg. """
+    return np.exp((r/rg)**2 * r.ndim/-2)
```
caspervdw: n [v0.2.4](https://github.com/soft-matter/trackpy/blob/41f800bcdaa403c5df4497e8690b974166857735/trackpy/tests/test_feature.py), the Gaussian is $e^{-(x^2+y^2)/(ndim*rg^2)}$
in which x, y are the meshgrids of coordinates and rg the absolute Rg.

If I write down the integrals for Rg of the general Gaussian $e^{-a r^2}$, I end up with $a = ndim / (2 *rg^2)$. So the factor in the v0.2 Gaussian is wrong. For 2D, there is a factor 2 difference, that I also mean in https://github.com/caspervdw/trackpy/commit/75877a4dd39b27253dc78afa6e1f85597ff48fec.

trackpy/artificial.py
```diff
@@ -0,0 +1,180 @@
...
+                         " dimensionality.")
+    diameter = validate_tuple(diameter, image.ndim)
+    if ecc is not None:
+        if len(diameter) != 2:
+            raise ValueError("Eccentricity is only defined in 2 dimensions")
+        if diameter[0] != diameter[1]:
+            raise ValueError("Diameter is already anisotropic; eccentricity is"
+                             " not defined.")
+        diameter = (diameter[0] / (1 - ecc), diameter[1] * (1 - ecc))
+    radius = tuple([(d - 1) / 2 for d in diameter])
```
caspervdw: We use a continuous function and discretize it. A feature with diameter `d` has a radius `d / 2`, that is subsequently discretized according to the pixel grid. In the continuous world the `- 1` makes no sense.

I cannot explain my thought on this any better, does anybody see my point?

nkeim: Now I'm really confused :wink:. Here's my idealized picture of how a symmetric 2D blob should work:

1. The Gaussian blob is a continuous function defined on (real numbers)^2, centered on an arbitrary real-valued location $x, y$. It has an Rg of `SIZE` units.
- That function is discretized. In a camera, each pixel is a (possibly weighted) average of the function over a certain region. For the artificial features, right now we sample the function at a single point, presumably the center of the pixel.
- The `locate` function tries to recover x, y, and Rg.

How is this different from what we are doing now? Is the problem at step 2? Are the various `+ 1` and `- 1` attempts to compensate for that?

What really troubles me is that the correction of 0.2 is approximately the same for different sizes. As the Gaussian gets larger (and therefore smoother), the details of the discretization should matter less and less. So why is it constant?

caspervdw:
@nkeim I think we are on the same line of thoughts. So trackpy puts the origin of the pixel at the *center* of the topleft pixel? I would call that position (0.5, 0.5). I never realized this. It should be somewhere in the docs.

Currently, the sampling is done at the topleft of every pixel. This should be adapted using 0.5 at some places then. Ideally there should be a (weighted) average indeed. For the general feature profile, you could do that numerically by generating a much finer grid of pixels and then average. But that is a future extension.

About the 0.2 mistake, I think this is because of the `- 1` in the (continuous) artificial feature **and** because in `test_rg` the background is chosen using `np.ones` instead of `np.zeros`. This background will increase the Rg. If I adapt these two things, the `0.2` is gone and values agree within 0-3% depending on subpx position of the feature.

caspervdw:
Sorry my previous message was very confusing. What I meant is the following:

When pixel (y, x) is the gaussian maximum brightest, trackpy will give you the coordinates (y, x). 

If you define the image origin at the topleft of pixel (0, 0), the feature having the gaussian maximum at pixel (y, x) will actually be at (y+0.5, x+0.5). This origin definition is clearly wrong because trackpy returns (y, x); the definition should be (0,0) at the center of the topleft px.

This agrees with my artificial code: pixels get the value of the function evaluated at the topleft of the pixel, which already gets you the (0.5, 0.5) shift.

nkeim:
Great! I take it that the Rg side of the tests is completely "clean" now.

I agree that trackpy's "coordinate system" should be known. I think that you were initially correct — we currently use the coordinate system of `imshow()` and `annotate()`, which puts integer coordinates on the pixel edges (with one image corner at 0,0). I think it's really important to use the same coordinate system as matplotlib; it avoids a lot of needless confusion. However, as I inadvertently showed in my earlier comment, this is not how one might naively imagine the coordinate system of a real image sensor, where there is dead space between pixels. So I agree that this should be documented, perhaps in the docstring for `locate()`.

So then the desired behavior is: if the artificial feature is perfectly centered on a pixel (i.e. the sampling grid is aligned with the center of the Gaussian), the coordinates returned by `locate()` should be half-integers. Am I right that we are testing for this desired behavior (or its equivalent)?

If so, a note in the `locate()` docstring should wrap this whole thing up. :smile:

caspervdw:
The current `locate` behaviour matches with `imshow`, which puts integer coordinates at pixel *centers*. You can test that with:

    import numpy as np
    import matplotlib.pyplot as plt
    frame = np.zeros((10, 10))
    frame[4, 4] = 1
    plt.imshow(frame)
    plt.plot(4, 4, marker='o')
    plt.show()

The produced plot has the marker centered in the pixel.
A Gaussian symmetric around pixel (5, 5) should be found at location (5, 5). The origin of trackpy (and imshow) is at the center of the topleft pixel. This is how it is tested now, I think it is ok to mention it in the docstring of locate.

nkeim:
So it is! I should pay more attention. It would be great to have some input from others on these changes, but for my part, I'm OK with merging once the convention is documented.

caspervdw:
It's a confusing subject and I didn't really see it documented in matplotlib (but I only inspected the `imshow` docstrings...). I am glad that it's finally cleared up now ;)

The convention is now documented inside the `draw_feature` and `draw_spots`, and I also put a note inside `locate` about the origin location, as well as about the y axis direction. I remember that I was confused in the beginning about the y-axis direction. This clears that up a bit.

I am also OK with merging as it is now. For the future: it would be nice to have some testing routines on differently shaped features and see how trackpy behaves on them. Also it could be implemented in the benchmarking suite.

tacaswell:
Be ware that if you use the `extent` kwarg on `imshow` in changes to the _edges_ being on the integers.

@caspervdw Could you add a note to the imshow docstring?

danielballan:
This is great! It's exciting to see this done right. We can use this to explore some outstanding questions about subpixel accuracy. I agree that it would be nice to see more tests on differently-shaped features. I also agree merging at this stage makes sense, and then anyone can use these building blocks to experiment.

About `extent`: thanks for the reminder, @tacaswell. If users invoke `extent` invoke it through `annotate(..., imshow_style={'extent': ...})`, we have to assume they know what they are doing. We don't use it ourselves anywhere in `plots.py`.

I see @caspervdw and @nkeim are +1 on merging, and @tacaswell is requesting a matplotlib PR, so I will swoop in and merge.

Issue #231, opened 2015-04-09
PANDAS warning using pred.link_df_iter()


apiszcz:
I am getting the warning below when calling pd.concat(pred.link_df...)

pred=trackpy.predict.NearestVelocityPredict(span=ld['velspan'])
f_iter=(frame for fnum, frame in df.groupby('frame'))
dft=pd.concat(pred.link_df_iter(f_iter, search_range=ld['d'], memory=ld['m'],adaptive_stop=ld['astop'],adaptive_step=ld['astep']
 ))

elsewhere I use df.is_copy = False to disable this warning. Is there any advice to eliminate this warning when calling pd_concat? 

C:\p\lib\pandas\core\frame.py:2513: SettingWithCopyWarning: esent
A value is trying to be set on a copy of a slice from a DataFrame
See the the caveats in the documentation: http://pandas.pydata.org/pandas-docs/stable/indexing.html#indexing-view-versus-copy
  **kwargs)

Thank you for any advice.

nkeim:
I can't seem to reproduce this. It might help to have a stack trace. Try something like

```
import warnings
with warnings.catch_warnings():
    warnings.simplefilter('error')
    (your code...)
```

Your pandas and trackpy versions could be helpful too. Thanks!

apiszcz:
trackpy is pre3.0
pandas 0.16
That is the full trace I see, for now I'm disabling with -Wignore, but it
may be nice to understand the issue.
​

apiszcz:
import warnings
                    with warnings.catch_warnings():
                        warnings.simplefilter('error')
                    oL.logDisplay("%s> %s APTrack:run velocity prediction
search\n"%(oU.time_stamp(),src))

pred=trackpy.predict.NearestVelocityPredict(span=ld['velspan'])
                    f_iter=(frame for fnum, frame in df.groupby('frame'))
                    #df.is_copy=False
                    dft=pd.concat(pred.link_df_iter(f_iter,
                                                    search_range=ld['d'],
                                                    memory=ld['m'],

adaptive_stop=ld['astop'],

adaptive_step=ld['astep']
                                                    ))


C:\p\lib\pandas\core\frame.py:2513: SettingWithCopyWarning: esent
A value is trying to be set on a copy of a slice from a DataFrame
See the the caveats in the documentation:
http://pandas.pydata.org/pandas-docs/stable/indexing.html#indexing-view-versus-copy
  **kwargs)
[...]

PR #236, opened 2015-04-25
BUG/TST: Address Pandas potential chained-assingment issue

nkeim:
See #231. Under certain conditions (still not quite understood), the prediction code could try to set values in an inadvertently-generated *copy* of a DataFrame. This is bad if the values in the original DataFrame were ever needed again, so Pandas generates a warning.

This PR fixes the line of code that caused the warning reported by @apiszcz in #231. It also makes Pandas more strict about this condition in tests, as I explain in one of the commit messages below:

"See http://pandas.pydata.org/pandas-docs/stable/indexing.html#indexing-view-versus-copy . This commit addresses no past or current known bugs, but it could prevent future surprises by causing tests to fail if there are chained assignments inherent in the code. In regular (non-test) use, such a condition would generate a warning only (which would promptly get cleared from the IPython output during multi-frame operations), while *possibly* leading to incorrect results."

I believe this PR *would* have caught exactly such a bug that I created when implementing diagnostics, and that, as it was, I had a very hard time tracking down.

### DIFF
--- a/trackpy/predict.py
+++ b/trackpy/predict.py
@@ -96,7 +96,7 @@ class _RecentVelocityPredict(NullPredict):
         positions = disps[self.pos_columns]
         vels = disps[[cn + '_disp_' for cn in self.pos_columns]] / dt
         # 'vels' will have same column names as 'positions'
-        vels.rename(columns=lambda n: n[:-6], inplace=True)
+        vels = vels.rename(columns=lambda n: n[:-6])
         return dt, positions, vels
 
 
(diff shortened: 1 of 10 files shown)


nkeim:
Pandas 0.12 doesn't detect this condition, which means this PR breaks compatibility. I'm in favor of backing out the changes to the tests, and maybe doing something like this at a later date, when we can deprecate 0.12 in good conscience. But adding a version conditional is also an option.

danielballan:
+1 for a version conditional

nkeim:
All set — tests pass on Pandas 0.12, 0.13, and 0.15.

apiszcz:
FYI: I'm using PANDAS 0.16. 0.16.0 final (March 22, 2015)
<http://pandas.pydata.org/#final-march-22-2015>
​

Issue #237, opened 2015-04-26
Issues with feature characterization


caspervdw:
For making some estimates of static error in anisotropic feature finding, I did some work on the static error estimation. In [this](http://nbviewer.ipython.org/urls/dl.dropboxusercontent.com/u/36853001/Feature%20characterization.ipynb) notebook I tried to make clear that there are some fishy things going on in the feature characterization.

I would like to propose to do all feature characterization on the raw image instead of the preprocessed (rescaled, bandpassed) image. For my 12-bit images the current implementation gives entirely wrong mass estimates.

If we then calculate the static error from the (correctly calculated) mass, and not from the max pixel intensity, the error estimate will be more robust to noise and I will be able to extend it for anisotropic features.

There are still some more things I need to solve, so I will probably update this post in the future.

What are your thoughts on this?

danielballan:
Thanks for investigating and presenting this thoroughly. The "signal" is certainly the most error-prone input to the noise computation, and your new way of calculating epsilon using mass instead of signal is clearly superior.

Do you know whether Savin & Doyle intended "signal" to come from the processed on unprocessed image? I couldn't convince myself either way, but I decided that the max pixel intensity could more accurately be measured with noise removed.

Do you think that rescaling is the whole source of the problem with mass estimates? Could they be due to overflow errors? I am surprised that unprocessed and processed images give such starkly different estimates for mass.

nkeim:
Wow, the current behavior does not look good!

Preprocessing is critical for some experimental images, because various optical errors or artifacts can add a long-wavelength "background" that must be subtracted off. This background would otherwise inflate the signal. Perhaps there is a different way to do the rescaling? Or as @danielballan suggests, maybe there is a bug somewhere.

caspervdw:
The mass is overestimated with a factor 16, so it seems to have something to do with the `scale_to_gamut` preprocessing that rescales my 12-bit image to 16-bit (giving you the factor 16).

I understand that background subtraction is needed, but why the rescaling?

danielballan:
The bandpass returns a float image, which is rounded to an integer dtype for performance reasons. (IIRC, I found that the dilation step in particular was much faster for ints.) Before this rounding takes place, the rescaling fills the gamut to retain as much contrast as possible. For uint8, this makes a difference: more detail is kept. But for higher bit depths, I would guess that is makes little if any difference.

[...]

PR #239, opened 2015-04-28
ENH Anisotropic feature characterization

caspervdw:
A number of changes that reach deep into the feature finding: please review carefully! The fixes are linked in subtle ways, that's why this is in one big commit. Please checkout [this](http://nbviewer.ipython.org/urls/dl.dropboxusercontent.com/u/36853001/Feature%20characterization%20v2.ipynb) notebook, which is a newer version that is referenced in #237, including demonstration of the new noise calculation and some tests.

- Changes the behaviour of the `mass` column: it is now calculated from
the raw image. This solves the issue #237 due to the rescaling of the
processed image and the subsequent calculation of mass on this image.
- Changes the behaviour of the `signal` column: previously it was
calculated from the raw image and subsequently the `black_level` was
subtracted. Because of the high sensitivity to noise, this gives
unpredictable results. It has been changed to: `signal =
preprocessed_feature.max()`. The rescaling has been accounted for via an
extra variable coming from `scale_to_gamut`.
- eccentricity is calculated on anisotropic masks too, assuming that
pixels are square
- size is calculated in all directions separately when an anisotropic
mask is used, giving `size_x`, `size_y` columns
- filtering on `maxsize` is done on sqrt(size_x^2 + size_y^2), which
equals the current behaviour if sizes are equal
- static error is calculated in all directions separately when an
anisotropic mask is used *or* when noise_size is anisotropic, giving
`ep_x`, `ep_y` columns
- `preprocessing.bandpass` now calculates the background level and
standard deviation (only if needed, defaults to current behaviour)
- `scale_to_gamut` returns the scale factor (only if needed, defaults to
current behaviour)
- docstring for eccentricity: 0 is circular, not 1 as it is documented
now
- characterization arrays are only initialized if needed

### DIFF
--- a/trackpy/feature.py
+++ b/trackpy/feature.py
@@ -215,6 +215,7 @@ def _refine(raw_image, image, radius, coords, max_iterations,
     # Declare arrays that we will fill iteratively through loop.
     N = coords.shape[0]
     final_coords = np.empty_like(coords, dtype=np.float64)
+    mass = np.empty(N, dtype=np.float64)
     raw_mass = np.empty(N, dtype=np.float64)
     if characterize:
         if radius[1:] == radius[:-1]:
@@ -277,32 +278,32 @@ def _refine(raw_image, image, radius, coords, max_iterations,
             plt.imshow(neighborhood)
 
         # Characterize the neighborhood of our final centroid.
-        raw_neighborhood = mask*raw_image[rect]
-        raw_mass[feat] = raw_neighborhood.sum()  # based on raw image
+        mass[feat] = neighborhood.sum()
         if not characterize:
             continue  # short-circuit loop
-        mass = neighborhood.sum()
         if radius[1:] == radius[:-1]:
             Rg[feat] = np.sqrt(np.sum(r_squared_mask(radius, ndim) *
-                                      neighborhood) / mass)
+                                      neighborhood) / mass[feat])
         else:
             Rg[feat] = np.sqrt(np.sum(x_squared_masks(radius, ndim) *
                                       neighborhood,
                                       axis=tuple(range(1, ndim + 1))) /
-                               mass)[::-1]  # change order yx -> xy
+                               mass[feat])[::-1]  # change order yx -> xy
         # I only know how to measure eccentricity in 2D.
         if ndim == 2:
             ecc[feat] = np.sqrt(np.sum(neighborhood*cosmask(radius))**2 +
                                 np.sum(neighborhood*sinmask(radius))**2)
-            ecc[feat] /= (mass - neighborhood[radius] + 1e-6)
+            ecc[feat] /= (mass[feat] - neighborhood[radius] + 1e-6)
         else:
             ecc[feat] = np.nan
         signal[feat] = neighborhood.max()  # based on bandpassed image
+        raw_neighborhood = mask*raw_image[rect]
+        raw_mass[feat] = raw_neighborhood.sum()  # based on raw image
 
     if not characterize:
-        return np.column_stack([final_coords, raw_mass])
+        return np.column_stack([final_coords, mass])
     else:
-        return np.column_stack([final_coords, raw_mass, Rg, ecc, signal])
+        return np.column_stack([final_coords, mass, Rg, ecc, signal, raw_mass])
 
 
 @try_numba_autojit(nopython=False)
@@ -581,7 +582,7 @@ def locate(raw_image, diameter, minmass=100., maxsize=None, separation=None,
             char_columns += ['size']
         else:
             char_columns += ['size_' + cc for cc in coord_columns]
-        char_columns += ['ecc', 'signal']
+        char_columns += ['ecc', 'signal', 'raw_mass']
     columns = coord_columns + char_columns
     # The 'ep' column is joined on at the end, so we need this...
     if characterize:
@@ -666,6 +667,7 @@ def locate(raw_image, diameter, minmass=100., maxsize=None, separation=None,
         # signal value has to be corrected due to the rescaling
         # mass was obtained from raw image; size and ecc are scale-independent
         f['signal'] /= scale_factor
+        f['mass'] /= scale_factor
         if not preprocess:
             # black_level and noise have not been determined, do it here
             _, black_level, noise = bandpass(raw_image, noise_size,
--- a/trackpy/feature.py
+++ b/trackpy/feature.py
@@ -579,22 +579,22 @@ def locate(raw_image, diameter, minmass=100., maxsize=None, separation=None,
         coord_columns = ['x', 'y', 'z'][:image.ndim]
     else:
         coord_columns = map(lambda i: 'x' + str(i), range(image.ndim))
-    char_columns = ['mass']
+    MASS_COLUMN_INDEX = len(coord_columns)
+    columns = coord_columns + ['mass']
     if characterize:
         if radius[1:] == radius[:-1]:
-            char_columns += ['size']
+            SIZE_COLUMN_INDEX = len(columns)
+            columns += ['size']
         else:
-            char_columns += ['size_' + cc for cc in coord_columns]
-        char_columns += ['ecc', 'signal', 'raw_mass']
-    columns = coord_columns + char_columns
-    # The 'ep' column is joined on at the end, so we need this...
-    if characterize:
+            SIZE_COLUMN_INDEX = range(len(columns),
+                                      len(columns) + len(coord_columns))
+            columns += ['size_' + cc for cc in coord_columns]
+        SIGNAL_COLUMN_INDEX = len(columns) + 1
+        columns += ['ecc', 'signal', 'raw_mass']
         if radius[1:] == radius[:-1] and noise_size[1:] == noise_size[:-1]:
-            all_columns = columns + ['ep']
+            columns += ['ep']
         else:
-            all_columns = columns + ['ep_' + cc for cc in coord_columns]
-    else:
-        all_columns = columns
+            columns += ['ep_' + cc for cc in coord_columns]
 
     # Find local maxima.
     # Define zone of exclusion at edges of image, avoiding
@@ -608,7 +608,7 @@ def locate(raw_image, diameter, minmass=100., maxsize=None, separation=None,
     count_maxima = coords.shape[0]
 
     if count_maxima == 0:
-        return DataFrame(columns=all_columns)
+        return DataFrame(columns=columns)
 
     # Proactively filter based on estimated mass/size before
     # refining positions.
@@ -628,18 +628,16 @@ def locate(raw_image, diameter, minmass=100., maxsize=None, separation=None,
 
     if count_qualified == 0:
         warnings.warn("No maxima survived mass- and size-based prefiltering.")
-        return DataFrame(columns=all_columns)
+        return DataFrame(columns=columns)
 
     # Refine their locations and characterize mass, size, etc.
     refined_coords = refine(raw_image, image, radius, coords, separation,
                             max_iterations, engine, characterize)
+    refined_coords[:, MASS_COLUMN_INDEX] /= scale_factor
+    if characterize:
+        refined_coords[:, SIGNAL_COLUMN_INDEX] /= scale_factor
 
     # Filter again, using final ("exact") mass -- and size, if set.
-    MASS_COLUMN_INDEX = image.ndim
-    if radius[1:] == radius[:-1]:
-        SIZE_COLUMN_INDEX = image.ndim + 1
-    else:
-        SIZE_COLUMN_INDEX = range(image.ndim + 1, image.ndim + image.ndim + 1)
     exact_mass = refined_coords[:, MASS_COLUMN_INDEX]
     if filter_after:
         condition = exact_mass > minmass
@@ -652,7 +650,7 @@ def locate(raw_image, diameter, minmass=100., maxsize=None, separation=None,
 
     if count_qualified == 0:
         warnings.warn("No maxima survived mass- and size-based filtering.")
-        return DataFrame(columns=all_columns)
+        return DataFrame(columns=columns)
 
     if topn is not None and count_qualified > topn:
         if topn == 1:
@@ -662,22 +660,19 @@ def locate(raw_image, diameter, minmass=100., maxsize=None, separation=None,
         else:
             refined_coords = refined_coords[np.argsort(exact_mass)][-topn:]
 
-    f = DataFrame(refined_coords, columns=columns)
-
     # Estimate the uncertainty in position using signal (measured in refine)
     # and noise (measured here below).
     if characterize:
         # signal value has to be corrected due to the rescaling
         # mass was obtained from raw image; size and ecc are scale-independent
-        f['signal'] /= scale_factor
-        f['mass'] /= scale_factor
         if not preprocess:
             # black_level and noise have not been determined, do it here
             _, black_level, noise = bandpass(raw_image, noise_size,
                                              smoothing_size, threshold, True)
-        ep = static_error(f, black_level, noise, radius, noise_size,
-                          coord_columns)
-        f = f.join(ep)
+        ep = static_error(refined_coords[:, MASS_COLUMN_INDEX], black_level,
+                          noise, radius, noise_size)
+
+    f = DataFrame(np.column_stack([refined_coords, ep]), columns=columns)
 
     # If this is a pims Frame object, it has a frame number.
     # Tag it on; this is helpful for parallelization.
--- a/trackpy/feature.py
+++ b/trackpy/feature.py
@@ -633,6 +633,8 @@ def locate(raw_image, diameter, minmass=100., maxsize=None, separation=None,
     # Refine their locations and characterize mass, size, etc.
     refined_coords = refine(raw_image, image, radius, coords, separation,
                             max_iterations, engine, characterize)
+    # mass and signal values has to be corrected due to the rescaling
+    # raw_mass was obtained from raw image; size and ecc are scale-independent
     refined_coords[:, MASS_COLUMN_INDEX] /= scale_factor
     if characterize:
         refined_coords[:, SIGNAL_COLUMN_INDEX] /= scale_factor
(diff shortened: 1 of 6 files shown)


danielballan:
Great! I am pleased and impressed that you have made anisotropic feature-finding a first-class citizen in all these parts of the feature-finding code.

I'm not yet convinced that we should be using the raw image to perform characterization. I think the important mass is the mass of the image that the algorithm is actually using. I think of the processed image as closer to the "truth" than the unprocessed one. In particular, the long-wavelength variations due to uneven lighting are a systematic error and we should be removing them. Let's discuss this further.

We currently use `raw_image` in just one place: to compute the noise. This is actually problematic, because if users want to perform their own preprocessing, `locate` does not let them pass in the original image. Perhaps `locate` should optionally accept two images, raw and processed. If we do ultimately use `raw_image` in more places, this definitely should be addressed.

caspervdw:
I think I have to agree with you: the mass from the processed picture should be closest to the true value. I pushed a commit that let me compare between the `raw_mass` and `mass`. In my tests, the raw mass increases with added noise, as expected. The preprocessed mass stays neatly constant but underestimates the true mass by 15-20%, independent of noise.

This seems to have to do with the moving average (boxcar) subtraction: at the feature, boxcar is subtracted which is a sort of average of the feature intensity. That's why mass underestimates the true mass. I can't figure out a way to correct for that, if there is any.

See [this](http://nbviewer.ipython.org/urls/dl.dropboxusercontent.com/u/36853001/Feature%20characterization%20v3.ipynb) notebook voor the tests I did.

Maybe it would be useful to standard output `mass` and only when characterize=True, also output `raw_mass` as an extra column?

nkeim:
If we agree that gamut rescaling is not appropriate for these feature characterizations, could we just un-rescale the calculated mass as well as the signal?

To second @danielballan: I would definitely appreciate better support for roll-your-own preprocessing.

caspervdw:
@nathankeim I undid the rescale in signal and mass already, after that the error was around 15-20%, independent of added noise.

Its just the way the preprocessing is designed, I guess. The lowpass filter keeps the intensity, only the rolling average subtracts too much.

nkeim:
Thanks. I looked at the source code but I missed the un-rescaling of mass. 

I see now why keeping and sometimes using the "raw" image (as optionally defined by the user) is advantageous. In the past, when I've had to do a custom convolution step to handle weird particle shapes, I've run `locate()` on the convolved image (usually with `preprocess=False`), then fed those coordinates and the raw image to `refine()` to obtain more accurate masses. The way I'm imagining @danielballan 's proposed solution, I would instead use a `raw_image` argument to `locate()`, which would work like this [?]:

- With `preprocess=True, characterize=True`, the `raw_mass` and epsilon columns are computed from the original image passed to `locate()`.
- With `preprocess=False, characterize=True`, this is also true, except now `raw_mass == mass`, and we note in the docstring that the epsilon columns are likely wrong.
- With `characterize=True, raw_image=raw_image`, `raw_image` is used to compute `raw_mass` and epsilon, *regardless* of the value of `preprocess`. This would be my use case.

Am I summarizing this idea correctly?

A separate question before (eventual) merge: Technical superiority aside, would the changes in this PR ever generate major disagreements with the output of other Crocker-Grier implementations? (I guess we are not counting the characterization bug in the historic IDL code.) I can't think of such a situation.

nkeim:
I forgot about R_g ("size"). It seems like the bandpassed image would be better for that, since otherwise, noise will always inflate R_g. But that means people like me who do weird nonlinear preprocessing would still need to do a second pass with `refine()` to get more accurate R_g.

The most generally correct way to compute R_g would be from the raw image, with noise removed *only*.

tacaswell:
The rolling average is super useful on non-uniformly illuminated samples.  From my NIPAM experiance, the intensities in the raw image were almost meaningless as there was a 10x larger difference in my images due to large scale variations as compared to the intensity of my particles against the local background.

I suspect to understand this we need to understand those spectral filter diagrams as the supression of the mass is likely due to the rolloff/sholders in the filter function which I think means the band-pass is too narrow.

For well seperated partcales local-otsu might be helpful.

caspervdw:
@nkeim @danielballan When the user supplies a `raw_image` to `locate`, we can just as well separate out the characterization in a different function like `tp.characterize(refined_features, image)`. This will make it clearer to use in my opinion. As you would have to do the image loading and slicing twice anyway, the performance wouldn't suffer much.

@tacaswell I completely agree on the usefulness of the rolling average. There is no problem in feature finding and I think we shouldn't change that, also to keep functionality close to the IDL code. The only issue is that the mass & max signal are always underestimated, because you subtract an average that consists of the feature itself and the background.

The issue is related to the normalization that is described in the original Corcker&Grier paper, eqn. 3 and 4. We don't do this normalization (at least for the rolling average we don't do it, for the gaussian convolution, I am not sure).
The error in our apparent mass is approx. M_apparent = M (1 - Nw/Ns), Nw the number of pixels inside the binary mask and Ns the number of pixel in the rolling average box. I investigated this a bit in [this notebook](http://nbviewer.ipython.org/urls/dl.dropboxusercontent.com/u/36853001/Processed%20mass%20deviation.ipynb) but I think that this should be solved inside the preprocessing and not afterwards like I did in that notebook.

Increasing the `smoothing_size` is no option, it is already by default `2*diameter+1` (so `2*(2*radius+1)+1`, I think this is a discrepancy between Crocker&Grier). Larger `smoothing_size` will give you enormous margins and issues with closely spaced features. Decreasing `smoothing_size` will increase the error in mass&signal (and therefore in epsilon).

Summarizing:

- For this PR, I think adding a new column `raw_mass` and fixing the rescale issue with `mass` and `signal` will solve most of the issues
- A new issue should be started in which the normalization is investigated
- A new issue in which the characterization on raw image should be facilitated

tacaswell:
It might be worth investigating wavelet decompositions for the band-pass.  (warning: throw away comment that will eat a week of someone's time)

danielballan:
I agree with @caspervdw's summary of next steps.

I have been thinking for awhile that `locate()` rolls too much functionality into one function. Even Crocker & Grier's API separates the bandpass and location steps into two function calls. I like the idea of documenting a slightly lower-level workflow with `refine()` and a new `characterize()` function.

tacaswell:
I am always :+1: on clean lower level APIs :smiling_imp: .

caspervdw:
I added some tests and rebased. There is a new issue here: `ImageSequence` may read images as integers (`skimage.io.imread`) or floats (`matplotlib.pyplot.imread`). So depending on if skimage is installed, you will get masses differing by a factor of 255 (for 8-bit pictures).

I think that the clearest way for calculating mass is still the sum of pixel intensities as provided by the image reader. The different dtypes returned by pims are actually a pims issue.

A possible solution could be: interpret float values as 8-bit integers. For most cases this should suffice. But I would rather let the user make this choice by picking the correct reader or using some dtype conversion.

What are your thoughts?

trackpy/feature.py
```diff
@@ -471,7 +486,7 @@ def locate(raw_image, diameter, minmass=100., maxsize=None, separation=None,
     DataFrame([x, y, mass, size, ecc, signal])
         where mass means total integrated brightness of the blob,
         size means the radius of gyration of its Gaussian-like profile,
-        and ecc is its eccentricity (1 is circular).
+        and ecc is its eccentricity (0 is circular).
```
tacaswell: Was this a glaring error in our docstring?

trackpy/masks.py
```diff
@@ -30,19 +35,42 @@ def r_squared_mask(radius, ndim):
...
 
 @memo
 def theta_mask(radius):
     "Mask of values giving angular position relative to center"
     # 2D only
-    tan_of_coord = lambda y, x: np.arctan2(radius - y, x - radius)
-    diameter = 2*radius + 1
-    return np.fromfunction(tan_of_coord, (diameter, diameter))
+    radius = validate_tuple(radius, 2)
+    tan_of_coord = lambda y, x: np.arctan2(radius[0] - y, x - radius[1])
```
tacaswell: attn @danielballan We had a vigorous discussion about this at work, see https://github.com/Nikea/scikit-xray/blob/b7c2316ad6a454916a4d56eb0c65207d503f8478/skxray/core.py#L637 

I think the `-` on the y term is incorrect.

trackpy/preprocessing.py
```diff
@@ -77,10 +77,16 @@ def bandpass(image, lshort, llong, threshold=None):
...
 
-def scale_to_gamut(image, original_dtype):
-    max_value = np.iinfo(original_dtype).max
-    scaled = (max_value/image.max()*image.clip(min=0.))
-    return scaled.astype(original_dtype)
+def scale_to_gamut(image, original_dtype, return_scale_factor=False):
+    scale_factor = np.iinfo(original_dtype).max / image.max()
+    scaled = (scale_factor * image.clip(min=0.)).astype(original_dtype)
+    if return_scale_factor:
+        return scaled, scale_factor
```
tacaswell: This is really bad practice (to change the return type/number of returns) depending on the input.

I would suggest splitting the `scale_factor` computation in to it's own function which gets called here and can be called where ever it is needed elsewhere in the code or to break the API and always return the scale factor.

trackpy/feature.py
```diff
@@ -552,30 +563,38 @@ def locate(raw_image, diameter, minmass=100., maxsize=None, separation=None,
                 # To avoid degrading performance, assume gamut is zero to one.
                 # Have you ever encountered an image of unnormalized floats?
                 raw_image = 1 - raw_image
-        image = bandpass(raw_image, noise_size, smoothing_size, threshold)
+        image, black_level, noise = bandpass(raw_image, noise_size,
```
tacaswell: I have not read the bandpass code yet, but I am guessing the last kwarg turn on/off the return of `black_level` and `noise` which (as commented on `scale_to_gamut`) is really bad practice.  In this case I would add a `bandpass_with_details` function and leave the api of `bandpass` as-is.

In terms of  implementation I am not sure of the best way (as I have not looked at the code yet).

trackpy/preprocessing.py
```diff
@@ -37,7 +37,7 @@ def ifftn(a):
         return pyfftw.interfaces.numpy_fft.ifftn(a)
 
 
-def bandpass(image, lshort, llong, threshold=None):
+def bandpass(image, lshort, llong, threshold=None, analyze_background=False):
```
tacaswell: I would split this up into three functions, on `_bandpass_raw` which returns `signal` and `result` and then two public functions `bandpass` which just calls the `where` and `bandpass_bkgnd_stats` which also does the background computations.

tacaswell:
I left a bunch of API level comments and one picky detail comment.

I don't fully understand the math of where the mass is being used, but I don't like the idea of making any assumptions about the values of input data.

trackpy/feature.py
caspervdw: Yes, you always get +-0 for round features. It was correct on the `refine` docstring.

trackpy/masks.py
caspervdw: I am no mathematician, but I would say that in the upper two quadrants (negative y), there should be positive angles and in the lower two quadrants (positive y) negative angles.

The scikit-xray reference works the other way around. It measures angles counterclockwise *from (x, y) to the positive x-axis*. I think in the usual definition it is measured *from the positive x-axis to (x, y)*.

    In [6]: angle_grid([1,1], (3,3))
    Out[6]: 
    array([[-2.35619449, -1.57079633, -0.78539816],
           [ 3.14159265,  0.        ,  0.        ],
           [ 2.35619449,  1.57079633,  0.78539816]])

From a second perspective: the `np.arctan2(y, x)` is made for normal Caretisian coordinates; as our y-axis is flipped, we should do `np.arctan2(-y, x)`.

trackpy/feature.py
```diff
@@ -552,30 +563,38 @@ def locate(raw_image, diameter, minmass=100., maxsize=None, separation=None,
...
+                                             smoothing_size, threshold, True)
     else:
         image = raw_image.copy()
     # Coerce the image into integer type. Rescale to fill dynamic range.
     if np.issubdtype(raw_image.dtype, np.integer):
         dtype = raw_image.dtype
     else:
-        dtype = np.uint8
-    image = scale_to_gamut(image, dtype)
+        dtype = np.uint16
```
caspervdw: One thing I forgot to mention: for floats, default was to downsample to 8-bit. I found this a bit harsh. But on the other side, this is the fastest possibility. Why did you put 8-bit @danielballan ?

trackpy/masks.py
tacaswell: The skxray definition is following the numpy definition (
http://docs.scipy.org/doc/numpy/reference/generated/numpy.arctan2.html)
which derives from (x, y) = (r * cos(th), r * sin(y)).

If you invert the y-axis before computing `atan2`   things will not round
trip correctly and computing what pixel a given angle corresponds to will
fail.

caspervdw: I think I misunderstood the skxray docs. So you want to stick to `np.arctan2(y, x)`, which will cycle clockwise through the picture? I put in the `-` to make it counterclockwise (so to account for the inverted y axis).

tacaswell: You need to pin your clock on that 12 is on the positive y-axis and 3 is the positive x-axis :smiling_imp:

caspervdw: In the case of the inverted y axis,  the 12 would be pointing downwards so that you would have to look against the back side of the clock.

caspervdw:
@tacaswell I restructured the `bandpass`, `scale_to_gamut` and `measure_noise` APIs.

In short, the new `scalefactor_to_gamut` calculates the necessary scalefactor. `scale_to_gamut` optionally takes the scalefactor, if not provided it calls `scalefactor_to_gamut`.

`bandpass` is back to the original state, with only some cosmetic changes

`measure_noise` now optionally takes the bandpassed image as argument. If not provided, it will call bandpass by itself. So if we provide the bandpassed image in `locate`, we circumvent a second call to the (slow) bandpass function and probably increase performance when `characterize=True`


Still pending is a test function to actually test the static error calculation. I will push it soon. It is absolutely necessary to put a `threshold` when noise size is high. Possibly a global otsu could be built into `measure_noise`, and used if threshold is not given.

trackpy/feature.py
```diff
@@ -274,24 +281,29 @@ def _refine(raw_image, image, radius, coords, max_iterations,
         mass[feat] = neighborhood.sum()
         if not characterize:
             continue  # short-circuit loop
-        Rg[feat] = np.sqrt(np.sum(r_squared_mask(radius, ndim) *
-                                  neighborhood) / mass[feat])
+        if radius[1:] == radius[:-1]:
```
tacaswell: Doesn't this need an `np.all` wrapped around it in general?

```python
In [100]: if np.arange(5) == np.arange(5):
   .....: print('yo')
   .....: 
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-100-03023ab06073> in <module>()
----> 1 if np.arange(5) == np.arange(5):
      2     print('yo')
      3 

ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

trackpy/feature.py
tacaswell: From talking with @danielballan I think this was a pure speed / memory usage thing.

trackpy/feature.py
```diff
@@ -609,20 +628,23 @@ def locate(raw_image, diameter, minmass=100., maxsize=None, separation=None,
...
 
     # Filter again, using final ("exact") mass -- and size, if set.
-    MASS_COLUMN_INDEX = image.ndim
-    SIZE_COLUMN_INDEX = image.ndim + 1
     exact_mass = refined_coords[:, MASS_COLUMN_INDEX]
     if filter_after:
         condition = exact_mass > minmass
         if maxsize is not None:
-            exact_size = refined_coords[:, SIZE_COLUMN_INDEX]
+            exact_size = np.sqrt(np.sum(refined_coords[:, SIZE_COLUMN_INDEX]**2))
```
tacaswell: Why does this get beat down to a scaler now?

trackpy/masks.py
tacaswell: It is mirrored, so it is 

```
   6
9      3
   12
```

not viewed from the back.  'counter clockwise' is from 3 -> 12.

trackpy/feature.py
caspervdw: It shouldn't! Forgot to specify axes.. Thanks, good one.

trackpy/masks.py
caspervdw: Against the back and upside down ;), either way clockwise and counter clockwise interchange.

`np.arctan(y, x)` will give you a *counterclockwise* movement for increasing angle in the standard ('right handed') coordinate system. It will give you a *clockwise* movement for the inverted ('left handed') coordinate system.

The movement as you sketch it is clockwise in the image and counter clockwise in the corresponding right handed axes system.

Check the output, increasing angle, clockwise motion:

    In [6]: angle_grid([1,1], (3,3))
    Out[6]: 
    array([[-2.35619449, -1.57079633, -0.78539816],
           [ 3.14159265,  0.        ,  0.        ],
           [ 2.35619449,  1.57079633,  0.78539816]])

caspervdw: Anyway, it boils down to if we want to stick to `np.arctan2(y, x)` which gives you a clockwise motion for increasing angle in the image coordinate system. I tend to agree with that to go with the ISO conventions and remove the `-` sign. Do you agree @danielballan @tacaswell  ?

tacaswell: Yes

trackpy/feature.py
caspervdw: Not necessarily, as `radius` is a tuple:

    if tuple(np.arange(5)) == tuple(np.arange(5)):
        print('yo')
    yo

tacaswell: But it gets used as a ndarray in line 231

caspervdw: Probably the line numbering changed, I can't find the line you are referring to. It should be a tuple, except for the use inside `_numba_refine`, where everything is an ndarray (but only in this function).

trackpy/feature.py
```diff
@@ -207,6 +208,7 @@ def _refine(raw_image, image, radius, coords, max_iterations,
     GOOD_ENOUGH_THRESH = 0.005
 
     ndim = image.ndim
+    anisotropic = radius[1:] != radius[:-1]
```
tacaswell: https://github.com/soft-matter/trackpy/pull/239/files#diff-8caca4dcbf45ed9519ea2c65b2c391e6R239 is the line I was looking at.  

If the length is 2 than this will work correctly with ndarrays (as length 1 arrays are apparently unambiguous).

caspervdw: The anisotropic 3d test does pass, but I will wrap an np.all alround to be
safe ok?
Op 13 mei 2015 22:42 schreef "Thomas A Caswell" <notifications@github.com>:

tacaswell: That does satisfy my paranoia.

Sorry if it feels like I am giving you a hard time about this PR.  I am very appreciative of all the work you have put in.

caspervdw: I am giving myself a hard time in doing this pr... thanks for all your
feedback :+1:

caspervdw:
In the previous week I added tests for static errors and also restructured the `static_error` routine so that it has the same API as before. I think it is ready for merge and with that, anisotropic feature finding support is complete in the python engine :smile:

danielballan:
Heroic effort! Very nice to see some of the older code reviewed and refactored in a better way.

PR #242, opened 2015-05-21
ENH 3D feature finding with numba

caspervdw:
This adds some performance increases for the 2D feature finding, as well as an extension to anistoropic and to 3D. I did this by writing several similar `_numba_refine` routines in a new `numba_feature.py`.

Apparently, the characterization was still slowing down the numba engine, even with `characterize=False`. Splitting the functions got me a 30% performance increase. Also replacing the length-2 buffers with scalar variables got some performance increase.

This PR contains some commits from #239, only the last 7 are for numba. Please refer to https://github.com/soft-matter/trackpy/commit/bf1509a8632608c590cebac6f4bfdb8c2bad9ed9, https://github.com/soft-matter/trackpy/commit/389eaa7ab3e05197f671100b4b922c6506e96c4a and https://github.com/soft-matter/trackpy/commit/5507595d0e29efaaed0a5d9907050feec8d8df97 for the major changes in the 2D version.

### DIFF
--- a/trackpy/feature.py
+++ b/trackpy/feature.py
@@ -167,12 +167,7 @@ def refine(raw_image, image, radius, coords, separation=0, max_iterations=10,
         _numba_refine(np.asarray(raw_image), np.asarray(image), int(radius[0]),
                       coords, N, int(max_iterations), characterize, shape,
                       mask_coordsY, mask_coordsX, N_mask_coords, r2_mask,
-                      cmask, smask, results,
-                      np.empty(2, dtype=np.float64),  # Buffer arrays.
-                      np.empty(2, dtype=np.float64),  # See function def.
-                      np.empty(2, dtype=np.float64),
-                      np.empty(2, dtype=np.float64),
-                      np.empty(2, dtype=np.int64))
+                      cmask, smask, results)
     else:
         raise ValueError("Available engines are 'python' and 'numba'")
 
@@ -312,8 +307,7 @@ def _refine(raw_image, image, radius, coords, max_iterations,
 @try_numba_autojit(nopython=False)
 def _numba_refine(raw_image, image, radius, coords, N, max_iterations,
                   characterize, shape, maskY, maskX, N_mask, r2_mask, cmask,
-                  smask, results,
-                  coord, cm_n, cm_i, off_center, new_coord):
+                  smask, results):
     SHIFT_THRESH = 0.6
     GOOD_ENOUGH_THRESH = 0.01
     # Column indices into the 'results' array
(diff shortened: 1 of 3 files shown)


danielballan:
Numba 0.19, in beta today, allows memory allocation in nopython mode. This radically simplifies things. If you feel like taking it for a spin on 0.19, they have requested feedback.

I am generally of the opinion that, as long as we support "no numba" mode, we can press our users to keep up with the latest version of numba where they run trackpy. It's a fast-moving project.

tacaswell:
I second tracking bleeding edge numba, it is moving fast enough that supporting every existing version is more work than we have the bandwidth for.

caspervdw:
It would be great to simplify things a bit, maybe obtaining some extra performance.
I agree that we shouldn't support old versions of numba as there is a fine pure python engine anyway, as long as the numba dependency is actually released. As 0.19 is unreleased, I would have to build from source. I don't really have time to do that currently, this is cumbersome on windows.

I was hoping this PR could go in before v0.3 as it already gives a significant performance increase for the 3D tracking. The 0.19 support may be something for a later version.

nkeim:
A few things, in order of decreasing urgency:
1. I just looked back at #217 and, to my horror, the final PR had `nopython=False`, even though *the whole point of that PR* was to keep numba 0.16+ from ever falling back to Python mode. So: If your code follows my example and allocates outside of the numba function, it should be safe to use `nopython=True`, and you should turn that on in this PR.
2. This is fantastic!!!! I think that having N implementations of the same algorithm is not always desirable — think what it must be like for an FFT library developer. However, we don't: We have a single Python version that always works, plus a set of accelerated implementations that are expected to give the same answer. So I think this is really quite manageable and well worth the performance increase.
3. It would be great to require numba 0.19+ in the long run, because the code would be so much more readable if we could move the array allocation back into the numba refine functions. But I don't think there should be any hurry to do that if everything's working. Closely tracking numba development is too much of a bumpy ride for my taste. Perhaps we could revisit the issue in several months?

danielballan:
My favorite edit:

```
raise NotImplementedError("The numba engine only supports 2D or 3D "  <-- :-)
                                       "images. You can extend it if you feel "
                                       "like a hero.")
```

OK, I agree we should merge and defer numba-wrangling for a later release. Given the large number of enhancements targeted for v0.3, a v0.3.1 may follow not long after.

@caspervdw If tests pass with nopython flipped to True for 2D and 3D, this is ready to go.

caspervdw:
I fixed the `nopython=True` and rebased. Ready to go!

Issue #219, opened 2015-03-05
A Faster Bandpass


danielballan:
[This notebook](http://nbviewer.ipython.org/urls/dl.dropbox.com/s/5cfyj0b046wq3ny/Making%20Bandpass%20Faster.ipynb?dl=0) benchmarks several different implementations of `bandpass`. I arrived at one that is twice as fast as trackpy's current implementation and up to 4X as fast with reasonable compromises in accuracy.

@nkeim, @caspervdw, @tacaswell and anyone else who is interested, would you run this notebook locally and confirm my timings? If this is correct and reproducible, I will make a PR to replace the current bandpass.

Unfortunately, the winning implementation does not allow us to hook into FFTW. Taking the long view, I don't want to completely blow away our work with FFTW even if it turns out to be slower. Instead, we can maintain several functions that give identical results, making `preprocessing.py` into a (very boring) research experiment on bandpass implementations in Python. The fastest one will be used by `locate(..., preprocess=True)`, but the others will be maintained for tinkering or future improvement.

caspervdw:
This looks great!! I will run it and report the benchmarks on my windows box.

nkeim:
Fantastic! I'd been waiting all day to try this for myself. It looks like you had some of your best ideas toward the end (as always happens), because I had to make some edits to get the notebook to run. So I'm sharing [my modified version](http://nbviewer.ipython.org/gist/nkeim/cb77b01385e3a1e39a7d). I adjusted the sample pool down to 200 images; 7.6 GB was a more comfortable fit on my MacBook.

I'm happy to report that the version with `gaussian_filter()` also consistently won on my Mac, by similar margins. Interestingly enough, the presence/absence of MKL seemed to make no difference — maybe numpy natively makes good use of SIMD?

Anyhow, this is just thrilling! This will put a big dent in the feature-finding time. I agree with your idea to have multiple implementations in `preprocessing.py`, just like we have multiple implementations of the subnet linker.

The biggest pain of adding this to trackpy will be changing the docs regarding FFTW. The performance tutorial can also suggest truncation as an additional speedup. I'd be happy to help with those changes if you'd like.

danielballan:
Thanks. Yes, it occurred to me that documenting this will be tricky. FFTW becomes a *very* optional dependency. @nkeim I'll take you up on your offer to help. I'll start on `preprocessing.py` if you start on the performance tutorial.

[...]

Issue #142, opened 2014-08-12
Optimize bandpass


danielballan:
* Re-examine `ndimage.uniform_filter`, which is not exactly a boxcar and which is accidentally maybe slower than a boxcar filter. Checkout the [boxcar filter](http://docs.scipy.org/doc/scipy-0.7.x/reference/generated/scipy.stsci.convolve.boxcar.html) scipy 0.7 and the [(1D-only?) version](http://docs.scipy.org/doc/scipy-0.13.0/reference/generated/scipy.signal.boxcar.htm) in modern scipy. Where did the 2D version go? Is it somewhere in astropy?
* Can we implement a 2D version as two 1D versions?
* Look into whether letting FFTW do more optimization (write more "wisdom") would be worthwhile.
* Also look into performance of FFTW in 2d, nd, "multiple 1D."

PR #244, opened 2015-05-25
ENH: Faster bandpass

danielballan:
@caspervdw I put the wining function from #219 into this PR. The simple test, which is based on what I did in the notebook, is failing. Feel free to take a look if you are inspired.

### DIFF
--- a/trackpy/preprocessing.py
+++ b/trackpy/preprocessing.py
@@ -2,23 +2,83 @@ from __future__ import (absolute_import, division, print_function,
                         unicode_literals)
 import six
 import numpy as np
-from scipy.ndimage.filters import uniform_filter1d
+from scipy.ndimage.filters import uniform_filter1d, gaussian_filter
 from scipy.ndimage.fourier import fourier_gaussian
 
 from .utils import print_update, validate_tuple
 
 
-# When loading module, try to use pyFFTW ("Fastest Fourier Transform in the
-# West") if it is available.
+def bandpass(image, lshort, llong, threshold=None, truncate=4):
+    """Remove noise and background variation.
+
+    Convolve with a Gaussian to remove short-wavelength noise and subtract out
+    long-wavelength variations, retaining features of intermediate scale.
+
+    This implementation relies on scipy.ndimage.filters.gaussian_filter, and it
+    is the fastest way known to the authors of performing a bandpass in
+    Python.
+
+    Parmeters
+    ---------
+    image : ndarray
+    lshort : small-scale cutoff (noise)
+    llong : large-scale cutoff
+    for both lshort and llong:
+        give a tuple value for different sizes per dimension
+        give int value for same value for all dimensions
+        when 2*lshort >= llong, no noise filtering is applied
+    threshold : float or integer
+        By default, 1 for integer images and 1/256. for float images.
+
+    Returns
+    -------
+    result : array
+        the bandpassed image
+
+    See Also
+    --------
+    legacy_bandpass, legacy_bandpass_fftw
+    """
+    lshort = validate_tuple(lshort, image.ndim)
+    llong = validate_tuple(llong, image.ndim)
+    if np.any([x*2 >= y for (x, y) in zip(lshort, llong)]):
+        raise ValueError("The smoothing length scale must be more" +
+                         "than twice the noise length scale.")
+    if threshold is None:
+        if np.issubdtype(image.dtype, np.integer):
+            threshold = 1
+        else:
+            threshold = 1/256.
+    settings = dict(mode='nearest', cval=0)
+    axes = range(image.ndim)
+    sizes = [x*2+1 for x in llong]
+    boxcar = np.asarray(image)
+    for (axis, size) in zip(axes, sizes):
+        boxcar = uniform_filter1d(boxcar, size, axis, **settings)
+    gaussian = gaussian_filter(image, lshort, truncate=truncate, **settings)
+    result = gaussian - boxcar
+    return np.where(result > threshold, result, 0)
+
+
+# Below are two older implementations of bandpass. Formerly, they were lumped
+# into one function, ``bandpass``, that used pyfftw if it was available and
+# numpy otherwise. Now there are separate functions for the pyfftw and numpy
+# code paths.
+
+# Both of these have been found to be slower than the new ``bandpass`` above
+# when benchmarked on typical inputs. Nonetheless, they are retained in case
+# they offer some advantage unforeseen by the authors.
+
+# All three functions give identical results, up to small numerical errors.
+
+
 try:
     import pyfftw
 except ImportError:
     # Use numpy.
-    USING_FFTW = False
-    fftn = np.fft.fftn
-    ifftn = np.fft.ifftn
+    FFTW_AVAILABLE = False
 else:
-    USING_FFTW = True
+    FFTW_AVAILABLE = True
     pyfftw.interfaces.cache.enable()
     planned = False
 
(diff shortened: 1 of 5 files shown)


caspervdw:
I will have a look. As the `gaussian_filter` is based on `ndimage.correlate1d` together with a Gaussian kernel generator, it will probably be even faster with that function. The Gaussian kernel can be generated using a memoizer.

caspervdw:
I set up a PR to your branch @danielballan . The new tests are still failing because margin values differ. Because these are irrelevant to out tracking I propose slicing them off in the legacy tests.

Speed tests of the current pyfftw, your gaussian_filter implementation (`modified`) and my correlate1d implementation (`combined`):

    In [16]:
    %timeit -n 10 current_bandpass(next(small), 3, 11)
    %timeit -n 10 modified_bandpass(next(small), 3, 11)
    %timeit -n 10 combined_bandpass(next(small), 3, 11)
    10 loops, best of 3: 23.3 ms per loop
    10 loops, best of 3: 16.1 ms per loop
    10 loops, best of 3: 15.3 ms per loop
    In [17]:
    
    %timeit -n 10 current_bandpass(next(prime), 3, 11)
    %timeit -n 10 modified_bandpass(next(prime), 3, 11)
    %timeit -n 10 combined_bandpass(next(prime), 3, 11)
    10 loops, best of 3: 45.4 ms per loop
    10 loops, best of 3: 14.9 ms per loop
    10 loops, best of 3: 13.9 ms per loop
    In [18]:
    
    %timeit -n 10 current_bandpass(next(medium), 3, 11)
    %timeit -n 10 modified_bandpass(next(medium), 3, 11)
    %timeit -n 10 combined_bandpass(next(medium), 3, 11)
    10 loops, best of 3: 595 ms per loop
    10 loops, best of 3: 377 ms per loop
    10 loops, best of 3: 348 ms per loop

nkeim:
Mentioning #142 here, which this will presumably close.

caspervdw:
@danielballan I will be able to assist with this PR until Thursday, from then I will be on holiday until August 7th. (see the relevant PR on your branch https://github.com/danielballan/trackpy/pull/2)

nkeim:
OK, this is fascinating: When you attempt to merge this into master (hopefully after merging fixes by @caspervdw and myself), you'll find that [`test_uncertainty_failure`](https://github.com/soft-matter/trackpy/pull/252) fails, and you won't know why. Turns out that the bandpass done in `uncertainty.roi()` changes just enough that it no longer leads to `ep == NaN` (instead, we get `ep == -0.0`!). I couldn't find a new set of parameters in `test_uncertainty_failure` to fix this. 

What does fix this test is to not have `uncertainty.roi()` do its own bandpass. (The test calls `locate()` with `preprocess=False`, because the test image is already built from gaussians.) 

So, 2 questions for discussion:

1. Why does this fix the problem? I.e. why the new `bandpass` behavior so different in this instance?

2. Why is `uncertainty.roi()` even doing a bandpass when `preprocess=False`? If the user calls `locate()` with `preprocess=False`, shouldn't trackpy presume that the image is already bandpassed? In the [discussion on uncertainty estimation](https://github.com/soft-matter/trackpy/pull/239), there was some mention of ways to accommodate users who do their own bandpass, but I can't see how the current code does. The present behavior instead acts as a performance "gotcha." Shouldn't `locate()` accept *both* `image` and `raw_image` when `preprocess=False`?

danielballan:
@nkeim I merged the fixes, and it looks like Travis does not reproduce the failure of `test_uncertainty_failure`. The mystery deepens!

As I look at `roi`, I can't believe I ever thought that function signature made sense. It requires `image` but only uses it if nothing is passed to the optional argument `image_bandpassed`, in which case it does its own bandpass using hard-coded parameters.

Instead, there should only be one image argument, and it should be bandpassed before it goes in, either explicitly by the user or automatically by `locate()`. Sound better?

nkeim:
A passing test?! Oh no!! :wink: 

Getting `-0.0` instead of `NaN` *would* be consistent with a mystery bug. But `test_uncertainty_failure` is not in this branch, only in master. Do you need to rebase this branch in order for Travis to run that test?

I agree with your proposed change to `roi`. Would `locate` then look like `locate(image, preprocess=True, raw_image=None, characterize=True)`? And maybe issue a warning if the user calls it with `preprocess=True, raw_image=None, characterize=True`, saying that `ep`, `signal`, and `raw_mass` may be incorrect?

There was the other idea tossed around in #239 to break down `locate()` into e.g. `preprocess()`, `find_peaks()`, `characterize()`, etc., so that the user could call `locate()` in most cases, but that it would be straightforward to build a functionally equivalent pipeline when needed. That is a little radical, but it does have its charms.

nkeim:
If I may offer some cheerleading for trackpy v0.3: I dusted off our benchmarks to answer a question (that turned out to be moot). But look how far we've come from v0.2.4 to the tip of this branch!

```
asv compare --machine shortcake 5ec747d4470fe9decceb5d2de8b588983189f828 6f746fc02147f3f9bfc11cde6469ef862a580d18

All benchmarks:

    before     after       ratio
  [5       ] [6       ]
-  561.91ms   129.75ms      0.23  artificial.TimeSuite.time_locate_dense
-  499.00ms    26.67ms      0.05  artificial.TimeSuite.time_locate_sparse
      2.38s      2.72s      1.14  real_data.DenseFeatureSuite.time_numba
-     9.94s      4.75s      0.48  real_data.DenseLinkingSuite.time_numba
-     4.95s      2.20s      0.44  real_data.WaterFeatureSuite.time_numba
     18.91s     16.17s      0.85  real_data.WaterFeatureSuite.time_python
   347.52ms   190.18ms      0.55  real_data.WaterLinkingSuite.time_numba
   344.81ms   180.60ms      0.52  real_data.WaterLinkingSuite.time_python
```

The regression for `real_data.DenseFeatureSuite.time_numba` could use some attention — I'm going to give it a once-over before I drop this — but when you look at the gains…

(Note: I used numba 0.14 to put everyone on the same footing.)

nkeim:
Update: #254 fixes the regression in `real_data.DenseFeatureSuite.time_numba`. I predict that the runtime will drop from 2.72 s to about 1.7 s. So, good news across the board, and the work remaining on this PR will be well worth it.

danielballan:
@nkeim Are all of your compute servers dessert-themed?

This is very motivating! Before we settle this, I want to benchmark Casper's alternative myself. When I tried that approach  it didn't do quite as well as the one I proposed in this PR. Perhaps a slightly different implementation or a difference of architectures. I'll circle back to it once pims is tagged. Soon!

tacaswell:
The dessert/food themed computer names is a trend from the Nagel lab (that I never picked up).

caspervdw:
Good work on this! The benchmarks look promising.

I agree with your ideas about `uncertainty.roi()`, it should simply accept any image and assume that it is already bandpassed. I don't remember why I put the weird calling signature, good that you caught this before tagging 0.3.

A more modular approach of `locate()` would be great. This will hopefully produce more readable code and give a more flexible use. This however is a big refactoring and it should be done right. I'll make a separate issue to discuss this.

trackpy/preprocessing.py
```diff
@@ -37,10 +100,16 @@ def ifftn(a):
...
+    """Remove noise and background variation.
+
+    Convolve with a Gaussian to remove short-wavelength noise and subtract out
+    long-wavelength variations, retaining features of intermediate scale.
+
+    This implementation performs a Fourier transform using numpy.
+    In benchmarks using typical inputs, it was found to be slower than the
+    ``bandpass`` function in this module.
+
     Parmeters
```
caspervdw: typo

danielballan:
Note: superseded by #259 and thus marked as merged by GH

PR #245, opened 2015-05-25
Conda recipe cleanup

danielballan:
* Remove stuff that really belongs in the pims conda recipe, not here.
* Add numexpr


nkeim:
Looks good. @danielballan , perhaps you should do the merge if you want to watch binstar the next day.

PR #250, opened 2015-06-24
FIX/TST: When an empty frame is put to FramewiseData, warn but continue

nkeim:
Revived version of #174


danielballan:
Thanks, buddy.

PR #254, opened 2015-07-26
PERF: Faster peak-merging

nkeim:
The code that merges nearby peaks (in `refine()`) was using numpy calls to compare peak masses. Since there are only 2 elements in each pair of peaks, switching to pure Python results in a huge speedup when there are lots of peaks. Merging ~10,000 pairs (don't ask!) went from ~20 s to < 0.5 s.

This also makes the operation in #252 in-place, and fixes some docstring typos.

### DIFF
--- a/trackpy/feature.py
+++ b/trackpy/feature.py
@@ -223,13 +223,14 @@ def refine(raw_image, image, radius, coords, separation=0, max_iterations=10,
             to_drop = []
             for pair in duplicates:
                 # Drop the dimmer one.
-                if np.equal(*mass.take(pair, 0)):
+                m0, m1 = mass[pair[0]], mass[pair[1]]
+                if m0 == m1:
                     # Rare corner case: a tie!
                     # Break ties by sorting by sum of coordinates, to avoid
                     # any randomness resulting from cKDTree returning a set.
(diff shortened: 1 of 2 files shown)


trackpy/feature.py
```diff
@@ -223,13 +223,14 @@ def refine(raw_image, image, radius, coords, separation=0, max_iterations=10,
             to_drop = []
             for pair in duplicates:
                 # Drop the dimmer one.
-                if np.equal(*mass.take(pair, 0)):
```
tacaswell:
```python
to_drop = [np.argmin(np.sum(positions.take(pair, 0), 1)) if m0 == m1 else 0 if m0 < m1 else 1 for m0, m1 in ((mass[p0], mass[p1]) for p0, p1 in duplicates]
```

I would not put money on that being actually correct.

nkeim: Typos aside, I couldn't get m0 and m1 to be assigned correctly.

nkeim:
…but since we're weighing economy vs. clarity, I just pushed a clearer (and possibly faster) version of that logic.

Sadly Python does not have a branching version of `cmp()`.

PR #259, opened 2015-08-12
Static error API

caspervdw:
As discussed in #244, this simplifies the API of `measure_noise`. It contains the commits of #244, only the latest commit is my addition. The function will never do image bandpassing.

The function `uncertainty.roi` is dropped and its functionality integrated into `measure_noise`. I am not sure wether to leave `uncertainty.roi` as a legacy function, but as the API will be changed anyway I can't think of a use for this. I am open for comments on this ;)

As the API is changing anyway, I replaced `diameter` by `radius`, to stop the tedious interconversion.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,14 +5,14 @@ matrix:
   include:
     # "Legacy" environments: oldest supported versions, without and with numba
     - python: "2.7"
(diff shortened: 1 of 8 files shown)


danielballan:
I agree with your choices here. I had mentioned a desire to dig into the bandpass benchmarking one more time, but I now think that would give diminishing returns. We need to get v0.3 out there, and this is a big boost. Merging.

PR #262, opened 2015-08-16
Build cleanup

danielballan:
Most of the changes here were automatic, the result of running `versioneer install` with the latest version of versioneer.


danielballan:
Merging own, because this only touches build components, no science code.

PR #264, opened 2015-08-17
FIX Image processing when preprocess=False

caspervdw:
This changes the preprocessing / rescaling logic. Rescaling is necessary because the subsequent grey dilation only takes integer images. Now 3 cases are considered:

- `preprocess == True`: do preprocessing and gamut rescaling
- `preprocess = False` and image of integer type: do nothing, rescaling is obsolete and would only degrade picture quality. omit copy because image is not changed after this.
- `preprocess = False` and float image: do gamut rescaling

### DIFF
--- a/trackpy/feature.py
+++ b/trackpy/feature.py
@@ -597,7 +597,7 @@ def locate(raw_image, diameter, minmass=100., maxsize=None, separation=None,
         if preprocess:  # identify background regions from the processed image
             black_level, noise = measure_noise(image, raw_image, radius)
         else:  # identify background regions from the provided image
-            black_level, noise = measure_noise(raw_image, raw_image, radius)
+            black_level, noise = measure_noise(image, raw_image, radius)
         Npx = N_binary_mask(radius, ndim)
         mass = refined_coords[:, SIGNAL_COLUMN_INDEX + 1] - Npx * black_level
         ep = _static_error(mass, noise, radius[::-1], noise_size[::-1])
--- a/trackpy/feature.py
+++ b/trackpy/feature.py
@@ -472,6 +472,8 @@ def locate(raw_image, diameter, minmass=100., maxsize=None, separation=None,
         warnings.warn("I am interpreting the image as {0}-dimensional. "
                       "If it is actually a {1}-dimensional color image, "
                       "convert it to grayscale first.".format(dim, dim-1))
+
+    # Determine `image`: the image to find the local maxima on
     if preprocess:
         if invert:
             # It is tempting to do this in place, but if it is called multiple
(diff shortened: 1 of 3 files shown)


caspervdw:
I am confused. Appearently the gamut rescaling of an integer image is necessary even when `preprocess == False`. Now the test `test_subpx_precision` does not pass.

Consider the picture

        pos1 = np.array([7, 13])
        pos2 = np.array([8, 13])
        pos = [7.25, 13]  # center is between pixels, biased left
        image = np.ones(dims, dtype='uint8')
        draw_point(image, pos1, 100)
        draw_point(image, pos2, 50)

Locate now returns [7.36, 13] instead of [7.25, 13]. Refine should have nothing to do with `image` and should work on `raw_image` right?

danielballan:
No, in C-G, refine works on the processed image, which is in many (perhaps
not all) contexts is closer to the ground truth, having removed artifacts.

caspervdw:
Ok appearantly I missed that aspect of refine.

What do you think about gamut rescaling when there is already an integer
image?
Op 17 aug. 2015 18:12 schreef "Dan Allan" <notifications@github.com>:

> No, in C-G, refine works on the processed image, which is in many (perhaps
> not all) contexts is closer to the ground truth, having removed artifacts.
>
> On Mon, Aug 17, 2015 at 11:43 AM Casper van der Wel <

danielballan:
Right, the code as it was implicitly assumed that `image` would always be float because of the `bandpass`. Of course that's not true when preprocessing is turned off. I agree that we should not rescale integer images in this case, and I'm puzzled that the accuracy of the subpx test is degraded. Can you turn on the walkthrough option to track the progress of refine?

caspervdw:
I think I got the mistake, the tests were actually expecting the wrong results. I also updated the 4-pixel tests, see commit message.

An image with a feature with intensities [100, 50] has its center of mass at 0.33 and not (as currently expected in the `subpx_precision` tests) at 0.25. That was the mistake in the tests. Because of the gamut rescaling we got away with this mistake.

So actually, leaving out gamut rescaling (with no preprocessing) enhances subpx precision for these two-pixel test images!

danielballan:
Well that makes much more sense! I'm glad *one* of the core devs knows how to compute a center of mass. *reluctantly places "dunce" hat on own head*

This probably explains the issues with GOOD_ENOUGH_THRESH as well, where I had use different values for numba and numpy to get the tests to pass.

danielballan:
This looks good. Since it touches critical code let's get a second review. @nkeim and @tacaswell please take a careful look and merge.

nkeim:
Great!! The new code is very easy to follow. As a longtime user of integer images with preprocess=False, this makes me happy.

PR #267, opened 2015-08-20
BLD: Add pytables to conda-recipe.

danielballan:


danielballan:
Any objections to this?

PR #269, opened 2015-08-20
TST: Fix test broken on pandas 0.16.

danielballan:
Minor fix: Pandas improved their test utils, and our faked-up expected result was missing some attributes.


danielballan:
Added a commit updating the package versions for the 'modern deps' Travis builds. All pass!

Issue #257, opened 2015-08-07
More modular feature finding


caspervdw:
As discussed in #239 and recently in #244, I would like to open a discussion on making feature finding more modular and hopefully more flexible and readable. I had a look at the code and we could breakdown locate in the following steps:

```
<parameter checks>
invert(raw_image)  # I would prefer to put this in a process_func and remove it from locate
image = preprocess(raw_image, noise_size, smoothing_size, threshold)
coords = local_maxima(image, radius, percentile, margin)
filter(coords, minmass, maxsize)
refine(raw_image, coords, separation, max_iterations, engine)  #possibly with characterize
filter(coords, minmass, minsize, topn)
characterize(raw_image, image, coords, radius, threshold)
```

To make the API user-friendly, we can work with `function`, `_function` combinations respectively including and not including the parameter checks.

I wonder what the performance loss would be if we would remove characterization from `refine` and write a new (possible numba) `characterize`.

What is the best place to make the transition `ndarray` -> `DataFrame`?

danielballan:
I think this is a good direction. Some quick thoughts:

1. I want `locate` to stick around in some form. It's good for new users with simple tasks.
2. But I think `batch` should be retired, in favor of a for loop. We don't have to remove it from the package, but we shouldn't use it in our examples.
3. I agree that `invert` should be removed from `locate`.

In my opinion, lowest-level API should favor numpy arrays. Converting *back* from a DataFrame to numpy array (`.values`) unfortunately copies data. And the column labels don't add much utility until you get to `characterize` anyway. Depending on how others feel, I'd suggest either (1) all low-level functions return numpy arrays or (2) `characterize` returns a DF, but nothing else does.

tacaswell:
I am ambivalent to 1. heartily agree with 2. and 3. 

I strongly agree with (1) and grudgingly accept (2).

nkeim:
I'm in favor of `locate` staying around. The main reason is that so many parameters are length scales that go as `diameter`. The secondary reason is that we are lucky that for many users, our core API is 2 functions. That makes trackpy super easy to get started with, as @danielballan says.

[...]

PR #270, opened 2015-08-20
Make refinement thresholds consistent; expose them through refine API

### LINKED ISSUE #191: GOOD_ENOUGH_THRESH different for numba and not-numba
In feature.py, the GOOD_ENOUGH_THRESH is different for the numba version and the not numba version:

line 199, thresh is set to 0.005 (was previously 0.01 back in February, changed in commit f963d4)
line 303, thresh is set to 0.01

Is this difference intentional?

### LINKED ISSUE #246: Clean up *_THRESH magic numbers.
Make SHIFT_THRESH and GOOD_ENOUGH_THRESH optional arguments of the refine functions and document them. Thanks @sciunto for catching this.

danielballan:
As the title says, this does two separate things:

1. Set  `GOOD_ENOUGH_THRESH` in the numba codepath to the same value it has in the numpy code path. This closes #191 (thanks, @RebeccaWPerry!) This will slightly change the output of the numba code path, but all the tests pass without modification.
2. Stop hard-coding `GOOD_ENOUGH_THRESH` and `SHIFT_THRESH`, instead exposing them as new optional arguments to `refine`. This does not change the output at all. This closes #246 (thanks again, @sciunto )

These parameters are obscure. I welcome opinions on my choice of names ('break_thresh` and `shift_thresh`) and my attempt at explaining them in the docstring.

I intentionally do not expose these parameters through the already-overstuffed `locate`. Users who want to customize these will have to build a processing pipeline the long way.

### DIFF
--- a/trackpy/feature.py
+++ b/trackpy/feature.py
@@ -140,6 +140,7 @@ def refine(raw_image, image, radius, coords, separation=0, max_iterations=10,
     """
     # ensure that radius is tuple of integers, for direct calls to refine()
     radius = validate_tuple(radius, image.ndim)
+    separation = validate_tuple(separation, image.ndim)
     # Main loop will be performed in separate function.
     if engine == 'auto':
         if NUMBA_AVAILABLE and image.ndim in [2, 3]:
--- a/trackpy/feature.py
+++ b/trackpy/feature.py
@@ -117,6 +117,9 @@ def refine(raw_image, image, radius, coords, separation=0, max_iterations=10,
         processed image, used for locating center of mass
     coord : array
         estimated position
+    separation : float or tuple
+        Minimum separtion between features.
+        Default is 0. May be a tuple, see diameter for details.
     max_iterations : integer
         max number of loops to refine the center of mass, default 10
     engine : {'python', 'numba'}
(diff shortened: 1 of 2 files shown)


danielballan:
P.S. Someday @caspervdw may want to make these new arguments accept tuples. :- )

trackpy/feature.py
```diff
@@ -118,12 +119,21 @@ def refine(raw_image, image, radius, coords, separation=0, max_iterations=10,
         estimated position
     max_iterations : integer
         max number of loops to refine the center of mass, default 10
+    engine : {'python', 'numba'}
```
tacaswell: this should have an `, optional` and include 'auto'

tacaswell: also, `seperation` is not documented

caspervdw:
I went through your changes, :+1: for merging!

trackpy/feature.py
nkeim: The `engine='auto'` code is in `locate()`, not here. Suggest we move it here as part of #257.

tacaswell:
I am 50/50 on holding this up over docstrings.  I won't merge it, but won't complain if someone else does.

danielballan:
I'll fix and self-merge right before I tag the RC.

PR #275, opened 2015-09-06
Minmass change convenience function

caspervdw:
`tp.minmass_version_change` now converts old minmass to new minmass values. It actually repeats the bandpass filter or else I couldn't get it to work. There are also some tests with it.

### DIFF
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -143,6 +143,7 @@ Utility functions
 .. autosummary::
     :toctree: generated/
 
+    minmass_version_change
     utils.fit_powerlaw
     strip_diagnostics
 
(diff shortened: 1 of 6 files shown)


danielballan:
Good. I think this is well explained. Can you add the new function to `doc/api.rst`? And can you copy some of the verbiage from the new docstrings into the release notes? It might be worth making a warning block in sphinx to help this stand out. Like so:

```
.. warning::

   blah blah blah
```

caspervdw:
I updated the docstrings. Also included some extra info on the 3d plotting and characterization.
Sorry that I didn't give that its own PR.

nkeim:
Looks good! This will really help reduce frustration with the transition. Thanks!

I know it's late, but do we want to mention this change in the warning generated when zero features survive the cuts? Might be too much at this point.

danielballan:
I meant to tack on a commit addressing @nkeim's comment. I opened a new PR for that minor change.

Issue #273, opened 2015-08-28
Warning when using `tp.subpx_bias`


caspervdw:
Something with pandas and plotting, we should fix the warning message when you use `tp.subpx_bias`

`C:\Anaconda3\lib\site-packages\pandas\tools\plotting.py:3227: UserWarning: To output multiple subplots, the figure containing the passed axes is being cleared
  "is being cleared", UserWarning)`

tacaswell:
Because you shouldn't be using pandas plotting at all :wink:

PR #277, opened 2015-09-08
FIX Warning message subpx_bias

caspervdw:
Solves #273


tacaswell:
This is an API break and needs to be documented as such.

caspervdw:
It's more like removing something that is not working (the ax argument in `subpx_bias`). But I will write a line in the release notes.

PR #278, opened 2015-09-09
FIX Check for right shape in locate

caspervdw:
This is an API change I would like to propose: instead of producing a warning message, locate will check at the start if the margins are larger than (half of the) image shape. If that is the case, no features can be found and an exception is raised.

I think it is good to have an exception instead of a warning, because in many cases users try to input RGB files (y, x, 3-4) or multichannel files (2-4, y, x). Then the file will be interpreted as 3D and it will take a long time to do the locate, while there are no features returned at all. I cannot think of any cases where the user would decide to ignore this situation. So an exception seems to be more in place.

The proposed changes are:
- shift the margin calculation to the front of locate
- check whether the margins are smaller then half the image shape
- the error message depends on whether the image looks suspiciously like RGB/RGBA/multichannel

### DIFF
--- a/trackpy/feature.py
+++ b/trackpy/feature.py
@@ -482,12 +482,25 @@ def locate(raw_image, diameter, minmass=100., maxsize=None, separation=None,
 
     noise_size = validate_tuple(noise_size, ndim)
 
-    # Check whether the image looks suspiciously like a color image.
-    if 3 in shape or 4 in shape:
-        dim = raw_image.ndim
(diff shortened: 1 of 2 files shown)


danielballan:
I think this is OK but I want to tread carefully. Can we punt this to v0.3.1?

nkeim:
Same. Agree that this would definitely improve trackpy.

caspervdw:
Agreed, we leave it for later

Issue #280, opened 2015-09-10
Linking performance issue


caspervdw:
With a colleague I had the following issue:

    z = f[(f.x > 860) & (f.x < 960) & (f.y > 609) & (f.y < 690) & (f.frame < 100)].copy()
    %timeit tp.link_df(z, search_range=10, memory=1) 
    Frame 99: 3 trajectories present
    1 loops, best of 3: 537 ms per loop

    z = f[(f.x > 860) & (f.x < 960) & (f.y > 609) & (f.y < 690) & (f.frame < 100)]
    %timeit tp.link_df(z, search_range=10, memory=1) 
    Frame 99: 3 trajectories present
    1 loops, best of 3: 30.7 s per loop

So much lower performance when linking a view. Is this a known issue @nkeim? Can we give some kind of warning if this case arises?

By the way, we are using 0.3.0rc1

nkeim:
Fascinating!! Will try to reproduce so I can profile it. Maybe we're just using pandas wrong.

On Sep 10, 2015, at 5:58 AM, Casper van der Wel <notifications@github.com<mailto:notifications@github.com>> wrote:


With a colleague I had the following issue:

z = f[(f.x > 860) & (f.x < 960) & (f.y > 609) & (f.y < 690) & (f.frame < 100)].copy()
%timeit tp.link_df(z, search_range=10, memory=1)
Frame 99: 3 trajectories present
1 loops, best of 3: 537 ms per loop

z = f[(f.x > 860) & (f.x < 960) & (f.y > 609) & (f.y < 690) & (f.frame < 100)]
%timeit tp.link_df(z, search_range=10, memory=1)
Frame 99: 3 trajectories present
1 loops, best of 3: 30.7 s per loop


So much lower performance when linking a view. Is this a known issue @nkeim<https://github.com/nkeim>? Can we give some kind of warning if this case arises?

caspervdw:
We might need to do something with `f.is_copy`, if that returns `None`, the dataframe is a copy and not a view.

nkeim:
I took the walkthrough notebook, restricted feature-finding to the first 100 frames, and then ran

`%prun t = tp.link_df(f[f.frame > 0], 5, memory=3)`

and

`%prun t = tp.link_df(f[f.frame > 0].copy(), 5, memory=3)`

There was a large timing difference, though not as dramatic as what you report. It appears to be entirely due to `gc.collect` going crazy when the input is a view:

```
[...]

PR #281, opened 2015-09-11
FIX Copy the features when trying to link a view

caspervdw:
This solves #280: performance was greatly degraded when trying to link a filtered dataframe.

### DIFF
--- a/trackpy/linking.py
+++ b/trackpy/linking.py
@@ -8,6 +8,7 @@ from copy import copy
 import itertools
 import functools
 from collections import deque
+from distutils.version import StrictVersion
 
 import numpy as np
 from scipy.spatial import cKDTree
@@ -547,7 +548,10 @@ def link_df(features, search_range, memory=0,
     if hash_size is None:
         MARGIN = 1  # avoid OutOfHashException
         hash_size = features[pos_columns].max() + MARGIN
-    if features.is_copy is not None and not copy_features:
+    # Check if DataFrame is writeable.
+    # I don't know how to do this for pandas < 0.16.
+    if (StrictVersion(pd.__version__) >= StrictVersion('0.16.0') and
+        features.is_copy is not None and not copy_features):
         warn('The features DataFrame is a view, so it is not writeable. '
              'The results will be output to a copy. Use copy_features='
              'True to prevent this warning message.')
--- a/trackpy/linking.py
+++ b/trackpy/linking.py
@@ -3,6 +3,7 @@ from __future__ import (absolute_import, division, print_function,
 import six
 from six.moves import zip, range
 import logging
+from warnings import warn
 from copy import copy
 import itertools
 import functools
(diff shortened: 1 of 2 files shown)


caspervdw:
I gave it some tries, but I can't identify views from copies in pandas 0.12 or 0.13. In 0.16, the attribute `is_copy` is also undocumented. Anyone has an ideas to solve it, apart from really testing the behaviour of the dataframe by trying to change it?

Otherwise, we just disable this check for pandas <0.16.

trackpy/linking.py
```diff
@@ -546,6 +547,10 @@ def link_df(features, search_range, memory=0,
     if hash_size is None:
         MARGIN = 1  # avoid OutOfHashException
         hash_size = features[pos_columns].max() + MARGIN
+    if features.is_copy is not None and not copy_features:
+        warn('The features DataFrame is a view, so it is not writeable. '
+             'The results will be outputted to a copy.')
```
nkeim: "Fun" with English: Past participle of "output" is "output".

nkeim: At the risk of being too verbose, it might be nice if this warning offered a more permanent solution: 'Use the "copy_features" argument to remove this warning.'

nkeim:
Thanks for taking a stab at this! It's too bad there is no easy solution, but, `getattr(df, 'is_copy')` might suffice — I'm OK with leaving this unfixed for earlier Pandas versions. Alternately, [this page](http://stackoverflow.com/questions/26879073/checking-whether-data-frame-is-copy-or-view-in-pandas) suggests checking `df2.values.base is df.values.base`.

In any case, given the uncertainty about this area of Pandas, I think this PR should include a test. See [this example](https://github.com/soft-matter/trackpy/blob/master/trackpy/tests/test_misc.py#L24).

caspervdw:
The is_copy attrubute always returns False on 0.12 / 0.13 so there is really no way to tell the difference.

I pushed your textual notes, the version conditional and test will have to wait for Monday... if you want to give it a shot this weekend please feel welcome to copy my branch and continue the work. Or we just postpone to 0.3.1.

danielballan:
@nkeim I am pretty sure that calling `df.values` will cost you, as the BlockManager's various single-type numpy arrays are copied into one object-type array.

nkeim:
Yikes! The aforelinked Stack Overflow page also mentions `df._is_view`. Maybe that works in older Pandas versions. But even if it does, I think a version check is much more desirable.

caspervdw:
I just pushed the version conditional and test. I didn't try the `df._is_view` on 0.12/0.13, but on 0.16 I always got `False` as return value. Strange that such a basic thing is so complicated...

danielballan:
+1 for version check and leaving it at that. Better safe than overhelpful and wrong. I think this can go into v0.3.0. Sounds like @nkeim is happy too -- merge at will.

trackpy/linking.py
```diff
@@ -546,6 +548,14 @@ def link_df(features, search_range, memory=0,
     if hash_size is None:
         MARGIN = 1  # avoid OutOfHashException
         hash_size = features[pos_columns].max() + MARGIN
+    # Check if DataFrame is writeable.
+    # I don't know how to do this for pandas < 0.16.
+    if (StrictVersion(pd.__version__) >= StrictVersion('0.16.0') and
+        features.is_copy is not None and not copy_features):
```
tacaswell: sorry, picky white space complaint, can you indent this a bit more?  I think the sin of messing up the continuation indentation is far less than the sin of aligning with the next indentation block.

nkeim:
Thanks, @caspervdw ! It's always great to snuff out a surprise :).

PR #285, opened 2015-09-16
BLD: Drop numpy 1.7.1

nkeim:
This changes the "legacy" Travis build environments so that they work with the [new tifffile 0.7.0](
https://pypi.python.org/pypi/tifffile), which is a dependency of pims. The oldest tested version of numpy is now 1.8.2 (released in August 2014). For users with an older version of tifffile, trackpy should still be compatible with numpy 1.7.1 and 1.8.0 for now.

Additionally, this addresses @tacaswell 's whitespace issue in #281 (sorry for my premature merge there!).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,9 +5,9 @@ matrix:
   include:
     # "Legacy" environments: oldest supported versions, without and with numba
     - python: "2.7"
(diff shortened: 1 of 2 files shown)


danielballan:
Fine by me.

Issue #282, opened 2015-09-12
FYI: Pandas 0.17rc1 warnings with trackpy 0.30rc1


apiszcz:
C:\p\lib\trackpy\trackpy\linking.py:748: FutureWarning: sort(columns=....) is deprecated, use sort_values(by=.....)
  features.sort('particle', inplace=True)

nkeim:
Thanks for reporting this! `sort_values` is not even in Pandas 0.16.2, so we cannot truly "fix" this without dropping compatibility with the current stable release.

If it turns out that Pandas can be instructed to suppress these warnings (except during testing), would that be wise?

nkeim:
If you want to see the changes required to get zero `FutureWarning`s from [Pandas 0.17.0rc1](http://pandas.pydata.org/#release-candidate-september-11-2015), see https://github.com/nkeim/trackpy/tree/pandas-0.17-future

apiszcz:
Thanks, Pandas 0.17rc1 deprecates a few areas, and I've add similar if
statements for initial testing. I can probably start testing this branch at
some point.

https://github.com/soft-matter/trackpy/compare/master...nkeim:pandas-0.17-future
​

apiszcz:
I copied over your changes for the 5 file, initial test is good with pandas 0.17rc1. Is there a way I can monitor your pandas 0.17 branch for future updates?

nkeim:
There may be a way to do that on Github. I myself don't plan on updating that branch at all until we're comfortable dropping Pandas 0.16, which I would guess is roughly a year from now.

apiszcz:
Thanks, this page has a list of 0.17 api conversion issues:
http://pandas-docs.github.io/pandas-docs-travis/whatsnew.html#backwards-incompatible-api-changes
​

apiszcz:
Note I've switched to 0.17 and will provide feedback as out on the edge. ​

danielballan:
Thanks, @apiszcz.

PR #287, opened 2015-09-19
BUG: Fix version check error for Pandas release candidates.

nkeim:
Discovered while investigating #282. Apprently `StrictVersion` is… strict.

### DIFF
--- a/trackpy/utils.py
+++ b/trackpy/utils.py
@@ -8,6 +8,7 @@ import re
 import sys
 import warnings
 from datetime import datetime, timedelta
+from distutils.version import StrictVersion
 
 import pandas as pd
 import numpy as np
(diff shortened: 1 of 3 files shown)


caspervdw:
:+1: It might make sense to move the bit of code to the top, so make a global `is_recent_pandas`.

nkeim:
Good idea. The test also contained a version check; I've now fixed the code there also. And since the test was in two places, I moved it to `utils.py`.

trackpy/utils.py
```diff
@@ -16,6 +17,14 @@
 
 import trackpy
 
+# Set is_pandas_recent for use elsewhere.
+# Pandas >= 0.16.0 lets us check if a DataFrame is a view.
+try:
+    is_pandas_recent = (StrictVersion(pd.__version__) >=
+                        StrictVersion('0.16.0'))
+except ValueError:  # Probably a release candidate
```
tacaswell: where does the valueerror come from?

nkeim: From the "rc" in the version string. We could use `LooseVersion`, which allows alphabetic characters. But I figured that the test needed a fallback anyhow, so I went for that.

danielballan:
Good enough!

PR #308, opened 2015-11-05
ENH Annotate3d works with mpl notebook backend

caspervdw:
This makes annotate3d work with the `%matplotlib notebook` backend. Before, you got a lot of whitespace above the real (scrollable) plot.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,16 +5,16 @@ matrix:
   include:
     # "Legacy" environments: oldest supported versions, without and with numba
     - python: "2.7"
-      env: DEPS="numpy=1.8.2 scipy=0.12.0 matplotlib=1.3 pandas=0.12.0 scikit-image=0.9 pyyaml pytables pyfftw"
+      env: DEPS="numpy=1.8.2 scipy=0.12.0 matplotlib=1.3 pillow<3 pandas=0.12.0 scikit-image=0.9 pyyaml pytables pyfftw"
     - python: "2.7"
-      env: DEPS="numpy=1.8.2 scipy=0.13.3 matplotlib=1.3 pandas=0.13.0 scikit-image=0.9 pyyaml numba=0.13.4 pytables pyfftw"
+      env: DEPS="numpy=1.8.2 scipy=0.13.3 matplotlib=1.3 pillow<3 pandas=0.13.0 scikit-image=0.9 pyyaml numba=0.13.4 pytables pyfftw"
     # "Recommended" environments: More recent versions, for Py2 and Py3.
     - python: "2.7"
-      env: DEPS="numpy=1.9 scipy=0.16 matplotlib=1.4 pandas=0.16 scikit-image=0.11 pyyaml numba=0.20 pytables pyfftw"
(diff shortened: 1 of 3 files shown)


caspervdw:
@nkeim @danielballan  Would be nice to have this inside 0.3, actually a bugfix for `annotate3d` in the notebook backend.

danielballan:
@tacaswell How do you like pseudo-context manager approach?

tacaswell:
I do not understand why you need this at all, but looks like a reasonable pattern.  

`plt.ioff` in idempotent so you could simplify top half as

```python
was_interactive = plt.isinteractive()
plt.ioff()
```

caspervdw:
With interactive mode on, there is a lot of empty space generated in the notebook (which keeps being there after closing the figure).

nkeim:
Would it make sense to use a `try...finally`? Besides the possibility of an exception, the user might just lose patience and Ctrl-C.

caspervdw:
Good point! I will add that.

caspervdw:
Done, also added tests for `annotate3d`. The function is still quite slow, there are options to improve (recycling the same figure object, for instance). But I'll leave that for later.

caspervdw:
Apparently, scikit-image >= 0.11 installed Pillow 3.0.0 automatically. I now added the Pillow dependency in Travis explicitly, so that tests for annotate3d (which needs pillow) pass.

However `plt.savefig` fails with pillow 3.0.0 (checkout https://travis-ci.org/soft-matter/trackpy/jobs/89812107)

    File "/home/travis/mc/envs/testenv/lib/python2.7/site-packages/matplotlib/figure.py", line 1476, in savefig
      self.canvas.print_figure(*args, **kwargs)
    File "/home/travis/mc/envs/testenv/lib/python2.7/site-packages/matplotlib/backend_bases.py", line 2211, in print_figure
      **kwargs)
    File "/home/travis/mc/envs/testenv/lib/python2.7/site-packages/matplotlib/backends/backend_agg.py", line 590, in print_tif
      dpi=dpi)
    File "/home/travis/mc/envs/testenv/lib/python2.7/site-packages/PIL/Image.py", line 1665, in save
      save_handler(self, fp, filename)
    File "/home/travis/mc/envs/testenv/lib/python2.7/site-packages/PIL/TiffImagePlugin.py", line 1320, in _save
      offset = ifd.save(fp)
    File "/home/travis/mc/envs/testenv/lib/python2.7/site-packages/PIL/TiffImagePlugin.py", line 608, in save
      if len(data) <= 4:
      TypeError: object of type 'int' has no len()

@tacaswell  I couldn't find a matplotlib issue about Pillow 3.0.0 incompatibility, does this ring a bell for you? Else I can make the issue.

For now, I fix the version to <3, just as in pims.

tacaswell:
No, but I can reproduce it and think it is a bug in pillow (things are going sideways in their tiff tag writer).

caspervdw:
Right, then it makes sense to fix Pillow to 2.9 for now.

danielballan:
This needs a rebase.

caspervdw:
Done

PR #310, opened 2015-11-06
BUG: characterize=False

nkeim:
When `locate()` called `refine()`, it assumed a certain order of keyword args. At some point `refine()` was changed. The result is that the value of `characterize` was not getting passed.

This PR adds a failing test, then fixes the bug.

This is arguably a fix that should be included in v0.3.

### DIFF
--- a/trackpy/tests/test_feature.py
+++ b/trackpy/tests/test_feature.py
@@ -428,6 +428,15 @@ class CommonFeatureIdentificationTests(object):
         expected = DataFrame(np.asarray(pos).reshape(1, -1), columns=cols)
         assert_allclose(actual, expected, atol=PRECISION)
 
+    def test_nocharacterize(self):
+        pos = gen_nonoverlapping_locations((200, 300), 9, 5)
+        image = draw_spots((200, 300), pos, 5)
+        f_c = tp.locate(image, 5, engine=self.engine, characterize=True)
+        f_nc = tp.locate(image, 5, engine=self.engine, characterize=False)
+
+        assert len(f_nc.columns) == 3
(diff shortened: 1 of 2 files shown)


danielballan:
Agreed. In a post-py2 world, many of these should become keyword-only args.

nkeim:
Tests are failing because x, y, mass appear to depend on `characterize`. Is that expected?

danielballan:
Hmm. I see the error. I'm not sure how that should depend on `characterize`.

trackpy/tests/test_feature.py
```diff
@@ -428,6 +428,15 @@ def test_subpx_precision(self):
...
         assert_allclose(actual, expected, atol=PRECISION)
 
+    def test_nocharacterize(self):
+        pos = gen_nonoverlapping_locations((200, 300), 9, 5)
+        image = draw_spots((200, 300), pos, 5)
+        f_c = tp.locate(image, 5, engine=self.engine, characterize=True)
+        f_nc = tp.locate(image, 5, engine=self.engine, characterize=False)
+
+        assert len(f_nc.columns) == 3
+        assert_allclose(f_c.values[:3], f_nc.values[:3])
```
caspervdw: You need to slice the feature ndarrays on the second axis, now there is a shape mismatch. I would say, don't slice f_nc and do [:, :3] for f_c.

caspervdw:
Good catch! The test has a bug, see line note. x, y, mass should certainly not depend on characterize.

nkeim:
@caspervdw Thanks! That seems to fix it.

PR #299, opened 2015-10-27
Replace pd.DataFrame.sort by pd.DataFrame.sort_values

hadim:
Replace pd.DataFrame.sort by pd.DataFrame.sort_values. Supress deprecated warning.

I also noticed a lot of 

```
/home/hadim/local/conda/envs/st/lib/python3.4/site-packages/PIL/Image.py:1130: ResourceWarning: unclosed file <_io.BufferedReader name='/home/hadim/Insync/Documents/phd/dev/contrib/python/trackpy/trackpy/tests/video/image_sequence/T76S3F00001.png'>
```

Any idea how to fix that ?

Issue #295, opened 2015-10-14
Test failure with Pandas 0.17


anntzer:
```
======================================================================
FAIL: test_zero_emsd (test_motion.TestMSD)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/tmp/makepkg/python-trackpy/src/trackpy-0.2.4/trackpy/tests/test_motion.py", line 136, in test_zero_emsd
    assert_series_equal(actual, expected)
  File "/usr/lib/python3.5/site-packages/pandas/util/testing.py", line 903, in assert_series_equal
    obj='{0}.index'.format(obj))
  File "/usr/lib/python3.5/site-packages/pandas/util/testing.py", line 673, in assert_index_equal
    assert_attr_equal('names', left, right, obj=obj)
  File "/usr/lib/python3.5/site-packages/pandas/util/testing.py", line 708, in assert_attr_equal
    left_attr, right_attr)
  File "/usr/lib/python3.5/site-packages/pandas/util/testing.py", line 798, in raise_assert_detail
    raise AssertionError(msg)
AssertionError: Series.index are different

Attribute "names" are different
[left]:  ['lagt']
[right]: [None]
```

nkeim:
We are testing trackpy with Python 3.5 and this error does not occur. Can you reproduce the reported failure with a more recent version of trackpy? You can also report the output of `import trackpy.diag; trackpy.diag.dependencies()`

anntzer:
With the current HEAD, this doesn't happen anymore, but now `test_link.test_copy` fails when checking for UserWarning, as pandas 0.17 now raises a FutureWarning at that spot.

nkeim:
@anntzer Thank you! I'm going to reference #299 and #282 as this pertains to the sticky issue of Pandas 0.17 (because it's backwards-incompatible with Pandas 0.16.).

PR #313, opened 2015-11-10
Fix pandas017 test

nkeim:
Addresses #295 by testing on Pandas 0.17 and fixing the failure. It also eliminates the top sources of Pandas 0.17 FutureWarnings.  I see this as a band-aid for #299 and #282 also. PRs against this one (or following this one) are welcome to fix the rest of the FutureWarnings.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,7 @@ matrix:
     - python: "3.4"
       env: DEPS="numpy=1.9 scipy=0.16 matplotlib=1.4 pandas=0.16 scikit-image=0.11 pyyaml numba=0.20 pytables pyfftw"
     - python: "3.5"
-      env: DEPS="numpy=1.9 scipy=0.16 matplotlib=1.4 pandas=0.17 scikit-image=0.11 pyyaml pytables"
+      env: DEPS="numpy scipy matplotlib pandas=0.17 scikit-image pyyaml pytables"
 
 install:
   # See:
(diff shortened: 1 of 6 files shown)


trackpy/utils.py
```diff
@@ -269,3 +275,16 @@ def quiet(suppress=True):
...
     else:
         trackpy.logger.setLevel(logging.INFO)
+
+def _pandas_sort_pre_017(df, *args, **kwargs):
+    """Use sort() to sort a DataFrame"""
+    return df.sort(*args, **kwargs)
+
+def _pandas_sort_post_017(df, *args, **kwargs):
+    """Use sort_values() to sort a DataFrame"""
+    return df.sort_values(*args, **kwargs)
```
caspervdw: Because the first argument changed its name from `columns` (in `df.sort`) to `by` (in `df.sort_values`), I think it would be better to make the call signature like this:

    pandas_sort(df, by, *args, **kwargs)

So that there cannot be any confusion.

caspervdw:
Great that you tackled this problem! You might also want to tackle #313 by replacing `icol` in motion.py, line 52.

nkeim:
Done and done.

.travis.yml
```diff
@@ -14,7 +14,7 @@ matrix:
     - python: "3.4"
       env: DEPS="numpy=1.9 scipy=0.16 matplotlib=1.4 pandas=0.16 scikit-image=0.11 pyyaml numba=0.20 pytables pyfftw"
     - python: "3.5"
-      env: DEPS="numpy=1.9 scipy=0.16 matplotlib=1.4 pandas=0.16 scikit-image=0.11 pyyaml pytables"
```
danielballan: Any particular reason to leave =0. after pandas?

danielballan: Nevermind. This rendered funny on either my phone or my brain. ;- )

PR #314, opened 2015-11-11
BLD: Import, don't bundle, IPython sphinx ext.

danielballan:
Housekeeping.


danielballan:
@tacaswell Check the line-count diff. #winning

PR #322, opened 2015-11-12
DOC Fixes

caspervdw:
Fixes missing links on http://soft-matter.github.io/trackpy/dev/api.html , plotting section

Plus bugfix of `subpx_bias`, `pos_columns` argument was ignored.

### DIFF
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -71,14 +71,14 @@ Trackpy includes functions for plotting the data in ways that are commonly usefu
     :toctree: generated/
 
     annotate
-	scatter
+    scatter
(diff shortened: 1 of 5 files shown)


danielballan:
Looks fine to me. I will leave it to @nkeim to merge --- don't want to risk interrupting his process. Can do this in 0.3.1 if need be.

caspervdw:
And added Pillow to optional deps in the installation documentation

nkeim:
It's a tough call, but I'm going to merge. The stakes are low because we can fix problems with the docs anytime.

Issue #301, opened 2015-10-30
RLS: Time to release v0.3?


nkeim:
It's been two months since v0.3rc1. The only open issues tagged for v0.3 are #200 and #286, both of which can be easily moved to v0.3.1. Are we ready to go?

If so, we have to 

- [x] Polish the release notes
- [x] Write a mailing list announcement and post it here (including a part for #271)
- [x] Set up versioned sphinx docs on Github pages (#182)
- [x] Follow the rest of the [checklist](https://github.com/soft-matter/trackpy/wiki/Release-Checklist)

I think @danielballan knows more about the versioned docs. I *might* have time to do the rest in the next several days.

danielballan:
Sounds good to me. I don't have a clear plan for versioned docs yet, but
maybe we can postpone that for 0.3.1 as well. The current docs are aimed at
0.3.0.

caspervdw:
Agreed, let's put it out. Only thing that I would like to add:

- [ ] Release PIMS 0.3.2

nkeim:
Here's a draft of the announcment. Release notes revisions pending.

*Subject:* ANNOUNCE: trackpy and pims releases

We're pleased to announce a major release of trackpy (v0.3), and a minor release of the movie-reading package pims (v0.3.2). The trackpy release is strongly recommended for all users, as it brings major improvements in speed and 3D feature-finding, and many other functionality and accuracy improvements.

### Survey

This release represents a tremendous amount of work by several contributors over the past year. For the core contributors especially, trackpy is a professional activity. We hope it can stay that way, so we need to gauge the impact of trackpy. If you consider yourself a trackpy user, please take 15–60 seconds to let us know by going to http://bit.ly/1Q2xTfg 

### Upgrading

Please be sure to read the [trackpy release notes](http://soft-matter.github.io/trackpy/whatsnew.html) before upgrading, especially the item under "Enhancements" about calculating masses. You may have to make minor edits to your scripts and notebooks to use them with v0.3.

Finally, if you have not been keeping up with pims, the v0.3.x series brings enhancements to 3D, features for slicing and processing movies, and a new reader for the StreamPix format.

If you are an Anaconda user, upgrading is as simple as "conda update -c soft-matter trackpy". Other methods are shown on the [installation page](http://soft-matter.github.io/trackpy/installation.html). The source for both packages is available on github (https://github.com/soft-matter).

------------

[...]

PR #323, opened 2015-11-12
BLD: Add test env representing "clean Anaconda"

nkeim:
This is a _possible_ way to check that the most basic installation instructions will work — that someone can simply install Anaconda and run `conda -c soft-matter install trackpy`. See https://github.com/soft-matter/pims/pull/200 and https://github.com/soft-matter/trackpy/issues/301#issuecomment-156223916

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,7 +26,8 @@ install:
   # If you're reading this, good luck out there. I'm not sure what to tell you.
   - conda update --yes conda
   - conda config --add channels soft-matter
-  - if [ $BUILD_CONDA_PKG ]; then conda install --yes conda-build; conda build --python $TRAVIS_PYTHON_VERSION conda-recipe; fi
+  # Set version and build package while we're still in the root env.
+  - if [ $BUILD_CONDA_PKG ]; then python setup.py build; conda install --yes conda-build; conda build --python $TRAVIS_PYTHON_VERSION conda-recipe; fi
   - conda create -n testenv --yes $DEPS pip nose setuptools python=$TRAVIS_PYTHON_VERSION
   - source activate testenv
   # for debugging...
--- a/.travis.yml
+++ b/.travis.yml
@@ -44,4 +44,4 @@ before_install:
   - ./miniconda.sh -b -p /home/travis/mc
   - export PATH=/home/travis/mc/bin:$PATH
 
-script: nosetests --nologcapture -a '!slow'
+script: cd trackpy/tests; nosetests --nologcapture -a '!slow'
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,17 +26,27 @@ install:
   # If you're reading this, good luck out there. I'm not sure what to tell you.
   - conda update --yes conda
   - conda config --add channels soft-matter
+  - |
+        if [ ! $BUILD_CONDA_PKG ]; then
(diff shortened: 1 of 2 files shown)


caspervdw:
I think you could drop scikit-image and pillow for this test.

danielballan:
Why not DEPS=trackpy? Probably have to refactor to add the soft-matter
channel below:

nkeim:
`DEPS=trackpy` would decouple test results from specific commits. It could be chaos…or it could be really convenient :smile:.

Could Travis do `conda build`, install the resulting package, and then run the tests?

tacaswell:
In principle doing the package build, installing in and then testing _should_ be possible, but you have to be careful that `nose` picks up the right version of the code.  We tried to do this for bnl code and ran into issues with the tests always picking up the source version.

nkeim:
Good point. What about moving the tests into the package root, then deleting `trackpy`?

nkeim:
OK, I have a working solution. The conda recipe can run nosetests automatically as part of the build, in a clean environment (this can be turned off if you're in a hurry). So one of the test scripts just runs "conda build" instead of nosetests.

The actual setup required for that is a mess. Any improvements to `.travis.yml` would be welcome. The other downside is that this build takes almost 5 minutes, instead of the usual 3. The side benefit is that testing would be baked into our conda release process, which seems like a pretty good idea.

Please forgive all the Travis emails you probably got, and the messy commit history. I'm still not 100% sold on this idea, but it would catch a whole class of particularly embarassing issues.

danielballan:
As it happens, we're having this same conversation at work.

I'm not completely sold -- debugging this kind of thing on Travis is so painful -- but it would have caught things like the pillow 3.0.0 problem right away.

caspervdw:
What is exactly the difference between the clean trackpy build and the older py3.5 build with the explicit dependency list?
I think the problem here is that we are a bit sloppy on the dependencies. If we would be concise on that, the old way of doing it would suffice. 
Until we do that, I'd say :+1: for this test setup. 

 But we really should discuss which dependencies we want (including for example deps-of-deps like skimage depending on pillow).

nkeim:
@danielballan I've just learned how hard it is to debug :stuck_out_tongue_closed_eyes:. I would argue that it's not quite as bad as it seems, since the only thing that happens on Travis exclusively is building the package straight from the repo. conda is meant to make the subsequent testing as reproducible as possible.

BUT, maybe the answer is
- Don't change the structure of `travis.yml`. It was already hard enough to debug. Just add an env with no version pinning. This will still be a pretty good "early warning system" for dependency problems.
- Merge the conda recipe that includes testing. I really like this way of validating the dependency list and the built package. One more bit of QA for stable releases. Does anyone know if Binstar will also run testing in our `dev` builds? (And do we even want that?)

danielballan:
I'm coming around this. Soon, I'd like to automatically upload conda packages of the master branch from Travis. And since we would be doing that blindly, I can see the case for running the tests on the built package. I like that all this happens in a separate build from the plain "Python 3.5 + latest version of everything" build.

Binstar won't run the tests, no. They have a "binstar build server" but it's unclear whether that project is still actively maintained. I would plan on using Travis to do our conda-building for the foreseeable future.

danielballan:
Note: all this was superceded by the entrance of conda-forge onto the scene.

PR #330, opened 2015-11-21
TST Be more picky about dtypes in motion.py

### LINKED ISSUE #329: Tests is failing on pandas 0.17.1
I commited directly to master with the change of the Zenodo DOI ( #326 ) . That triggered Travis and the latest-version testcase failed because of the recent release of pandas.

Another reason why not commit to master directly. Sorry about that :S I will look into it today.

caspervdw:
Pandas 0.17.1 triggered some failing tests. I fixed this by:

* fixing particle column dtype to float, frame to integer, x/y to float
* `subtract_drift` now only adapts the position columns, before it converted everything to float
* removed all references to `.sort` with `pandas_sort`, to fix the flood of deprecation messages

This fixes #329

### DIFF
--- a/trackpy/motion.py
+++ b/trackpy/motion.py
@@ -184,7 +184,10 @@ def subtract_drift(traj, drift=None):
 
     if drift is None:
         drift = compute_drift(traj)
-    return traj.set_index('frame', drop=False).sub(drift, fill_value=0)
+    traj.set_index('frame', inplace=True, drop=False)
+    for col in drift.columns:
+        traj[col] = traj[col].sub(drift[col], fill_value=0)
+    return traj
 
 
 def is_typical(msds, frame, lower=0.1, upper=0.9):
--- a/trackpy/tests/test_feature.py
+++ b/trackpy/tests/test_feature.py
@@ -20,6 +20,7 @@ import trackpy as tp
 from trackpy.try_numba import NUMBA_AVAILABLE
 from trackpy.artificial import (draw_feature, draw_spots, draw_point,
                                 gen_nonoverlapping_locations)
+from trackpy.utils import pandas_sort
                                 
 from scipy.spatial import cKDTree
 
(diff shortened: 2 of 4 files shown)


danielballan:
> fixing particle column dtype to float, frame to integer, x/y to float

Why make 'particle' float and not int?

caspervdw:
I agree: int would be better. But float is the current behaviour of link_df. @nkeim raised an issue to change this,  but I didnt want to break the API in this pr.

danielballan:
Ah, right. I forgot that discussion. Good enough for now.

trackpy/motion.py
```diff
@@ -184,7 +184,10 @@ def subtract_drift(traj, drift=None):
 
     if drift is None:
         drift = compute_drift(traj)
-    return traj.set_index('frame', drop=False).sub(drift, fill_value=0)
+    traj.set_index('frame', inplace=True, drop=False)
+    for col in drift.columns:
```
danielballan: This will cost some performance, as the subtraction is no long vectorized over columns. I think there is a way to do this using the new `DataFrame.assign` method, but that might require a pretty recent version of pandas. Thoughts?

caspervdw: I should actually time the performance: it's only a loop on 2 or 3 columns. Before, every column got changed to `float` suggesting that all the data was rewritten. Possibly my change made it even faster.

Else: `df.loc[pos_columns].sub(..)` might do the job faster.

danielballan: Ah, that's right, we don't need `.assign` for this. If you don't mind benchmarking I'll wait to merge. If you don't want to bother I'm fine with merging -- you are right that's a small loop.

caspervdw: I'll not have time soon so please merge to fix the build again!

danielballan:
Annnd we're back to a green badge. Thanks for the quick fix and for culling down the stream of warnings.

caspervdw:
Nice :)  there are still some warnings left, something with pytables and an unclosed file. Couldnt reproduce them locally and travis really doesnt give a lot of info. We should check on them sometime.

PR #293, opened 2015-10-11
FIX Gaps in frames are not ignored anymore

caspervdw:
Fixes #292 , but not yet for `link_df_iter`. There is an exception because I only changed the `levels` generator and in `link_df_iter`, this is zipped together with the original iterable, which is unchaged.

Do you agree on the approach?

Another approach would be `link_iter` checking on the framenumbers.

PR #333, opened 2015-11-25
PERF/API: Drop in-place link_df()

### LINKED ISSUE #294: Link_df runs very slow with large data sets
I'm attempting to link particle positions to form trajectories using link_df for a data set containing ~1,700 particles for 90,000 frames.  In attempting this, I noticed link_df was running very slow.  After 12 hours, the code had completed under 8,000 frames (I estimate that the entire data set would have taken maybe 4 days).  While I understand that that is a huge data set, I previously was able to complete the linking of the same number of particles (~1,700) for 24,000 frames, which took about 4 hours.  Naively, I was expecting the time to scale linearly, and so this 90k frame data set would take 12-16 hours (much more reasonable than 4 days).  But this was not the case.

My hypothesis is there is a memory issue.  The 90k frame data file is ~3 GB, and uses 10-15% of my machine's memory (linux server, found using "top" command).  Any suggestions on how this process can be sped up?

Some useful information, versions of various packages are: python 2.7.10, numba 0.12.2, numpy 1.8.2, trackpy 0.2.4rc7, and pandas 0.14.1.

Thanks for your help.

nkeim:
This drops the ability to link a DataFrame in-place. The only API change is to remove the `copy_features` option to `link_df()`. The corresponding test was removed as well. `link_df()` now wraps `link_df_iter()`, with some basic (and hopefully well-optimized) one-time Pandas operations on the input and output. This code is not very optimized, but it is lightweight and hopefully avoids the present _N_^2 performance bug.

Does this close #294 ? Probably, but some simple performance scaling checks would be wise. I'll see if I can get around to it.

### DIFF
--- a/trackpy/linking.py
+++ b/trackpy/linking.py
@@ -557,9 +557,16 @@ def link_df(features, search_range, memory=0,
                        ignore_index=(not retain_index))
 
     if not retain_index:
-        return pandas_sort(linked, ['particle', t_column]).reset_index(drop=True)
+        linked = pandas_sort(linked, ['particle', t_column]
+                           ).reset_index(drop=True)
     else:
-        return linked.reindex(features.index)
+        linked = linked.reindex(features.index)
+
+    # Order columns as in original DataFrame, then 'particle', then
+    # any other new columns (i.e. diagnostics)
+    return linked.reindex(columns=list(features.columns) + ['particle'] +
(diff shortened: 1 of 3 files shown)


nkeim:
I just realized that this reverses #293 and does not handle empty frames. Sorry I forgot about that. I think that means we need more tests.

caspervdw:
This is superseded by #442. Closing

PR #336, opened 2015-12-12
Pair Correlation

dwieker:
static.py contains functions to calculate the "Pair Correlation" function g(r) in 2 and 3 dimensions. pairCorrelation2D and pairCorrelation3D contain an optional parameter "handle_edge," which, if True, will force g(r) to compensate for the reduced volume around particles near the edge, at the cost of run time.


caspervdw:
Great, thanks for this contribution! I think pair correlations would be a great addition to trackpy.

A few months ago I finished my own code that does the same thing, but I never actually made the PR and now I am not using the code anymore.

I think your implementation is better because it uses KDTree, mine has a brute force approach of calculating every combination of particles and then make a histogram (or a bit experimental: a [kernel density approximate](https://en.wikipedia.org/wiki/Kernel_density_estimation)). But I did have algebraic expressions for all the edge collisions, you might get a large performance increase if you use those (especially in the 3D case)

I will put the code in a branch on my repo so that you can check it out.

danielballan:
Travis is failing only on the build with pandas 0.12 due to an indexing issue. Perhaps a `.loc` should been an `.iloc` there? I haven't looked too closely -- just trying to give you a head start. :- )

caspervdw:
@dwieker if you need any help please let us know! I could also propose some changes (basically copy over some of those algebraic functions) in a PR to your repo if you prefer.

dwieker:
Thanks! I'm on vacation at the moment and haven't worked on much, but I'll be sure to fix the pandas .12 issue soon and take a look at your edge handling functions. Ill let you know if I need any help.

dwieker:
@caspervdw can you point me to your edge handling algebraic functions? I'd like to try them out and see the performance increase.

caspervdw:
Yes I linked to the proper line in my previous post. It is `arclen_2d_bounded` that calculates the circumference of circles at `pos` with radius `dist` given a certain `box`. Similarily, there is `area_3d_bounded`. The code is a bit messy and it contains some experimental things that I tried. Please let me know if you can't make sense of it then I'll try it.

caspervdw:
@dwieker Did you try to use the edge handling code that I posted?
If not we could just proceed with the current code and update it at a later time.

dwieker:
@caspervdw I did try using your edge handling code and I had a hard time gaining any significant speed boosts. The code is under my "grfastedgehandling" branch, if you want to look. For systems where the majority of particles have to be treated with edge handling it is faster, but for systems where it only matters for a small fraction of the particles, it's slower. Let's just proceed with the current code. Maybe I can find a faster solution at some point.

caspervdw:
Do you have a benchmark code for your g(r) code?  I could check mine with that and see what exactly is the speed increase. Then we know if it makes sense to proceed.

trackpy/static.py
```diff
@@ -0,0 +1,264 @@
+from scipy.spatial import cKDTree
+import random
+import numpy as np
+from warnings import warn
+
+
+def pairCorrelation2D(feat, cutoff, fraction = 1., dr = .5, p_indices = None, ndensity=None, boundary = None,
```
caspervdw: pep8: no upper case in function name (make it `pair_correlation_2d` for instance). And remove spaces around =

trackpy/static.py
```diff
@@ -0,0 +1,264 @@
...
+    else:
+        xmin, xmax, ymin, ymax = boundary
+        # Disregard all particles outside the bounding box
+        feat = feat[(feat.x >= xmin) & (feat.x <= xmax) & (feat.y >= ymin) & (feat.y <= ymax)]
+
+    if ndensity is None:
+        ndensity = feat.x.count() / ((xmax - xmin) * (ymax - ymin))  #  particle packing density
+
+    if p_indices is None:
+        p_indices = random.sample(range(len(feat)), int(fraction * len(feat)))  # grab random sample of particles
```
caspervdw: it might be that `np.random.randint(low, high, shape)` is faster for this

trackpy/static.py
```diff
@@ -0,0 +1,264 @@
...
+    # For edge handling, two techniques are used. If a particle is near only one edge, the fractional area of the
+    # search ring r+dr is caluclated analytically via 1 - arccos(d / r ) / pi, where d is the distance to the wall. If
+    # the particle is near two or more walls, a ring of points is generated around the particle, and a mask is
+    # applied to find the the number of points within the boundary, giving an estimate of the area. Below,
+    # rings of size r + dr  for all r in r_edges are generated and cached for later use to speed up computation
+    n = 1000  # TODO: Should scale with radius, dr
+    refx, refy = _points_ring2D(r_edges, dr, n)
+
+    for idx in p_indices:
+        dist, idxs = ckdtree.query(points[idx], k=max_p_count, distance_upper_bound=cutoff)
```
caspervdw: You can `query` for multiple points at once. This is probably more efficient and it could allow you to write out the for loop.

`ckdtree.query(points[p_indices], ...)` returns a 2D array of dist and 2D array of indices

nkeim: This would return a result set that scales with N^2, which for large systems can exceed available memory. So it might give a performance boost, but it should not be the default.

caspervdw:
It is only N^2 in worst case, that is, all N particles are closer than max_dist to each other. The KDTree makes it N x k, with k the max number of neighbors. 
Do you ever encounter more than 10000 particles?

But it was only a suggestion. Mainly because you could do the histogram in one go on the flattened dist array. Saves you N-1 calls to histogram.

trackpy/static.py
```diff
@@ -0,0 +1,264 @@
...
+
+    if boundary is None:
+        xmin, xmax, ymin, ymax =  feat.x.min(), feat.x.max(), feat.y.min(), feat.y.max()
+    else:
+        xmin, xmax, ymin, ymax = boundary
+        # Disregard all particles outside the bounding box
+        feat = feat[(feat.x >= xmin) & (feat.x <= xmax) & (feat.y >= ymin) & (feat.y <= ymax)]
+
+    if ndensity is None:
+        ndensity = feat.x.count() / ((xmax - xmin) * (ymax - ymin))  #  particle packing density
```
caspervdw: it should be `count - 1` because each time you drop one particle: the one you are measuring from.

trackpy/tests/test_static.py
```diff
@@ -0,0 +1,146 @@
...
+            points[i:i+10, 1] = (r + epsilon)*np.sin(theta)
+            i += 10
+        points[:10] = 0
+
+        return pandas.DataFrame(points, columns = ['x', 'y'])
+
+
+    def _lattice3D(self, n = 20):
+        #Generates 3D lattice, spacing = 1
+        x,y,z = [],[],[]
```
caspervdw: np.indices will do this too

trackpy/tests/test_static.py
```diff
@@ -0,0 +1,146 @@
...
+
+
+    def test_correlation2D_lattice(self):
+        ### Lattice Test
+        # With proper edge handling, g(r) of the particle at the center should be the same as g(r) for all particles.
+        lattice = self._lattice2D()
+
+        # Calculate g_r on the center particle only (index 210)
+        edges, g_r_one = pairCorrelation2D(lattice, dr=.1, cutoff=8, p_indices=[210])
+        g_r_one /= np.linalg.norm(g_r_one) #We care about the relative difference of g_r in this case, so let's normalize both.
```
caspervdw: Bikeshedding: you might import `norm` in the root namespace (`from numpy.linalg import norm`) as it is used many times.

nkeim:
@caspervdw That is a VERY good point. `histogram()` has a surprising amount of overhead.

If someone has 30k particles and they are careless with the cutoff radius, they could easily bring down their computer. But maybe then they wouldn't do that a second time :wink:.

Since this is a non-critical performance matter, maybe it should be spun off as a separate issue. I'll let @dwieker weigh in.

Thanks for your line comments!!

dwieker:
@caspervdw Thank you for the comments!

Regarding the histogram: could there be some code that approximates the memory needed by g(r) and throws an assertion error / warning if it might crash the computer? Or it could automatically switch to the current for-loop version. 

The largest data set I've used g(r) for was dense packing of around ~50,000 particles. Picking a reasonable cutoff gives a set of ~1e7. Is this too much?

I don't think this is super critical; the code has been fast enough even on large data sets. 

As for a speed benchmark: Yes, I can post some numbers and code when I get home.  10,000 random particles in a box size 1000x1000x1000 in with a cutoff of 30 takes ~3.5 seconds (with edge handling on). Changing the box to 100x100x100 (so that most particles have to take into account edge handling)
takes ~20 seconds.

caspervdw:
Thanks for the numbers @dwieker I will see if I can compare your code to mine using these configurations.

dwieker:
@caspervdw 
I should also include the dr used: .1 

Here's the code:

from trackpy.static import pairCorrelation2D, pairCorrelation3D
import pandas
import numpy as np

def test_speed():
    import time
    x,y,z = np.random.random(size=(3,10000))*100
    df = pandas.DataFrame({'x':x, 'y':y, 'z':z})
    t = time.time()
    pairCorrelation3D(df,dr=.1,cutoff=30)
    print "Run time for speed_test: " + str(time.time() - t)

test_speed()

trackpy/static.py
dwieker: I'm picking a random set of indices, so there can't be duplicate numbers.

caspervdw:
@dwieker I looked into your code and by merging it with my old functions, I managed to get a 3x performance increase for the speedtest you sent.

Maybe you can look in to the performance and check if it is indeed faster in other tests. I made a PR to your branch with the changes.

dwieker:
@caspervdw  Thanks! It's 2-3x faster than the previous code for all of my benchmark tests

trackpy/static.py
```diff
@@ -0,0 +1,343 @@
...
+    # drop zero and infinite dist values
+    mask = (dist > 0) & np.isfinite(dist)
+    dist = dist[mask]
+
+    if handle_edge:
+        pos_repeated = pos[:, np.newaxis].repeat(max_p_count, axis=1)[mask]
+        arclen = arclen_2d_bounded(dist, pos_repeated,
+                                   np.array([[xmin, xmax], [ymin, ymax]]))
+    else:
+        arclen = np.pi*dist
```
caspervdw: Forgot a factor 2 here! My bet

caspervdw:
Great! I left one more line note, for the rest it is good to merge. I am very enthusiastic about this addition!

Maybe @nkeim can do a review of the current code.

Two more things:
- the tests should be made more PEP8 compliant. Mainly the line length, it should be < 80 characters.
- could you remove the large datafiles from git history?

danielballan:
What is the intended behavior when `handle_edge` is False?

caspervdw:
I can see the point in dropping this parameter alltogether and always handle the edges.

dwieker:
My thinking is that g(r) would only change by a minuscule amount for large sets of particles and a small cutoff radius (because the vast majority of particles would not be near an edge), so turning off edge handling would help speed things up. 

The "fraction" parameter serves essentially the same purpose, though, so I agree that "handle_edge" could be removed

trackpy/static.py
```diff
@@ -0,0 +1,343 @@
...
+
+    # Estimate upper bound for neighborhood particle count
+    max_p_count = int(np.pi * (r_edges.max() + dr)**2 *
+                      ndensity * max_rel_ndensity)
+    # Protect against too large memory usage
+    if len(pos) * max_p_count > MAX_ARRAY_SIZE:
+        max_p_count = int(MAX_ARRAY_SIZE / len(pos))
+
+    dist, idxs = ckdtree.query(pos, k=max_p_count, distance_upper_bound=cutoff)
+    if np.any(np.isfinite(dist[:, -1])):
```
dwieker: @caspervdw Can you explain this line to me? I just used pair_correlation_2d on a small set of data and this exception was thrown

nkeim: This checks if any of the particles were found to have `max_p_count` neighbors. If so, this is a sign that we probably made `max_p_count` too small, and there are actually more neighbors within `cutoff`.

In the below error message, I'm not sure how using `fraction` would address this condition. But `max_rel_ndensity` would. So maybe that should be changed.

caspervdw: We should probably be more concise over here, I was confusing 2 different errors here:

At line 86, the `max_p_count` (so indirectly `max_rel_ndensity`) is reduced automatically if the produced array would be larger than `MAX_ARRAY_SIZE`. This should produce an error instead of changing the parameter silently. This error should recommend lowering either `max_rel_ndensity`, `cutoff`, or `fraction` as they all reduce the array size.

At line 90, there is a different issue (as @nkeim described): `max_rel_ndensity` is too low. The exception should say that this should be increased. (Which could lead to the first exception).

dwieker:
I am fairly new to git, so I have a question. If I make multiple local commits on my machine and then push to github, will all those past commits show up? Or do I need to push to github after every commit I make? 

Thank you!

danielballan:
If you commits are all on one branch (it sounds like they are) only one
push is needed. Pushing once uploads all the commits on a branch.

dwieker:
@danielballan  Thank you!

I believe I removed all the large files from the git history (2Dpacking.csv ~6M). It should be good to merge!

nkeim:
@dwieker Fantastic! I'd like to take one last look before it goes in. (And someone else should probably do the merge.)

nkeim:
See https://github.com/dwieker/trackpy/pull/6 for my suggested changes to the docstrings. Those were the only issues I spotted; otherwise, it's good to go!

caspervdw:
Tests pass, merged!

Thanks @dwieker for pushing this great addition

As a side note: my docstrings of `arclen_2d_bounded` and `area_3d_bounded` are respectively absent of sloppy. Something to improve for later.

nkeim:
Congratulations, @dwieker ! And thanks, @caspervdw, for your contributions on this PR.

If I had to nominate the next addition to `static.py`, it would be a full g(r vector). But I suspect a lot of people will be very glad to have the g(r) we do now.

danielballan:
Bravo!

Issue #335, opened 2015-12-11
Faster MSD calculations


hadim:
I found a Python implementation to compute MSD with FFT which seems faster than conventional MSD algorithm because the algorithm is `O(N*log(N))`.

I did a quick and small benchmark here : https://github.com/hadim/public_notebooks/blob/master/quick_msd_calculations/quick_msd_calculations.ipynb

Two things : 

- the returned `DataFrame` is less detailled than in `tp.msd`.

- the msd array is sliced at the end of the function to take into account `max_lagtime` parameter which is really not efficient ! I didn't find a way to integrate it in the algorithm and I don't know wether it's possible.

Would you be interested by a PR to integrate it into trackpy ?

If you do, we also need to discuss how to inegrate it with `tp.imsd` (maybe with a `fft=False` default parameter ?).

Edit : I added `max_lagtime` into the loop. According to the author of the algorithm it's not possible do things faster with `max_lagtime`.

danielballan:
Interesting. That's a big speedup!

Yes, I'd be very interested in seeing this as a PR into trackpy. Since it has different advantages and limitations compared to the existing implementation, I'm guessing it will make the most sense to incorporate as a separate function that either replaces or sits alongside the existing one (rather than implementing a keyword argument to switch between them).

If it is not possible to compute the extra "detail" columns using the FFT implementation, perhaps we could make `detailed_emsd` (current implementation, `detail=True`) and `emsd` (FFT implementation).

In addition, I think the existing `emsd`/`imsd`' could be vectorized more effectively. I'll look at this in a separate issue/PR.

hadim:
I created a `msd_fft` function. For `emsd` and `imsd` I think it's cleaner to add a paramter `fft` to avoid code duplication.

What do you think ?

If you don't like it, tell me how should I duplicate the function please : `emsd`/`emsd_fft` and  `imsd`/`imsd_fft`

danielballan:
Any thoughts from the rest of the team of this?

hadim:
I'll make the PR the way I think it's fine and then we can discuss there and iterate to modify the API.

danielballan:
Great. Thanks a lot for looking at this!

hadim:
Discussion continue to #337

PR #337, opened 2015-12-13
MSD computation with FFT

hadim:
See this https://github.com/hadim/public_notebooks/blob/master/quick_msd_calculations/quick_msd_calculations.ipynb and #335

### DIFF
--- a/trackpy/motion.py
+++ b/trackpy/motion.py
@@ -7,31 +7,7 @@ from pandas import DataFrame, Series
 from scipy.spatial import cKDTree
 
 
-def _msd_N(N, t):
-    """Computes the effective number of statistically independent measurements
-    of the mean square displacement of a single trajectory.
-
-    Parameters
-    ----------
-    N : integer
-        the number of positions in the trajectory (=number of steps + 1)
-    t : iterable
-        an iterable of lagtimes (integers)
-
-    References
-    ----------
-    Derived from Equation B4 in:
-    Qian, Hong, Michael P. Sheetz, and Elliot L. Elson. "Single particle
-    tracking. Analysis of diffusion and flow in two-dimensional systems."
-    Biophysical journal 60.4 (1991): 910.
-    """
-    t = np.array(t, dtype=np.float)
-    return np.where(t > N/2,
-                    1/(1+((N-t)**3+5*t-4*(N-t)**2*t-N)/(6*(N-t)*t**2)),
-                    6*(N-t)**2*t/(2*N-t+4*N*t**2-5*t**3))
-
-
-def msd(traj, mpp, fps, max_lagtime=100, detail=False, pos_columns=['x', 'y']):
+def msd(traj, mpp, fps, max_lagtime=100, detail=False, pos_columns=None):
     """Compute the mean displacement and mean squared displacement of one
     trajectory over a range of time intervals.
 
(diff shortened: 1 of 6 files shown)


hadim:
`emsd` requires `N` to be computed but I ma not sure how I could get it from the FFT implementation. Any hints are welcome !

caspervdw:
Thanks for the PR! I have a function that calculates `N`, the effective number of independent measurements, for given lagtimes. I will make a PR to your fork.

hadim:
Cool. I just added you to my repo as collaborators so you can push into it.

caspervdw:
There it goes. This also is an enhancement to the previous impelemtation that used `2N/lagtime`. In this reference they derived an exact relation for the number of statistically independent measurements for msd in single trajectories:

http://www.ncbi.nlm.nih.gov/pmc/articles/PMC1260142/

caspervdw:
By the way, the whole `motion.py` should get some optimization, the current `msd()` is clearly suboptimal. But that is for another time!

hadim:
Very nice equation for the `N` computation !

Should I integrate `mdf_fft` to `emsd/imsd` ?

caspervdw:
I broke the tests... will forcepush a fix soon!

caspervdw:
Do you think it would be possible to extend `msd_fft` to `<x>` and `<x^2>` as well?
I am thinking of replacing the line:

    D = np.square(r).sum(axis=1)

If so, I think that `msd_fft` can completely replace `msd`. Then you could just rename:

`msd_fft` --> `msd` and `msd` --> `_legacy_msd`

hadim:
That would be nice. I don't have time this week to have a look.

Feel free to push more things.

Otherwise I'll try to work on it next week.

caspervdw:
I partly addressed the issue by vectorizing the code differently. Now it returns `<x^2>` columns. This costs some performance but the vectorizing I did gained some. Overall, I think slightly slower than previously.

How to return `<x>`? Maybe we can implement this in the non-fft way of `pos.diff(lagtime).mean()` because this seems to be a fundamentally different problem than `<x^2>`.

hadim:
Very cool ENH @caspervdw !

For the `<x>` column we could use the non-fft way but what do you think to make it optional ? I know it will break the API but people can then choose to not compute `<x>` to have better performance if needed.

caspervdw:
Maybe you can check out the performance, I managed to use a similar approach to your code.

Another question: how to deal with gaps in the trajectories? The original `msd` inserts `np.nan` at missing frames. If you do that with the fft-approach, the fft will return `np.nan` for all values of lagtime. If you don't do it, you make an error as the analysis expects evenly spaced timepoints. I see some options:

- You can split up the trajectories in small bits and run each bit through the function, but then you waste some data for the larger lagtimes.
- Try to patch the code with a non-conventional fft with unevenly spaced timepoints, other parts will need to have some patching as well.
- Make this code a `no_gaps` version and keep the old code for the `gaps` case.

caspervdw:
@hadim I finished the implementation of your code. In the latest commit, I optimized the old code that stays in use when there are gaps in the trajectory. The `msd` function chooses if whether uses `_msd_fft` or `_msd_gaps`. 

Performance gains:

    10 loops, best of 3: 48.8 ms per loop  # original code
    100 loops, best of 3: 11.3 ms per loop   # optimized code, original approach
    1000 loops, best of 3: 1.18 ms per loop  # new FFT-based code

Possibly `imsd` and `emsd` could be vectorized more efficiently leading to even larger performance gains.

hadim:
Very nice @caspervdw ! Thanks for the work. Sorry I wasn't able to check your code. I will be busy again for the next few weeks. But feel free to merge this version if it's working. We can open new PR in the future to improve it.

caspervdw:
The code should be checked by at least one other person before merging, because I had to much to do with it to merge it myself.

hadim:
I just tested `tp.msd` on my dataset and msd look the same as before. I also did a quick review of the code and all seems fine to me.

I also see the performance are better by a factor of 10 with `_msd_fft`.

Good to merge to me (I saw the travis error are caused by something not related to this PR).

caspervdw:
The new test uncovered an already existing bug in `emsd` for pandas 0.12. Something with pandas multiindexes.

nkeim:
I don't know the details of this bug, but you should seriously consider dropping pandas 0.12 in Travis (and issuing a warning in this function if the version is too old) as an alternative to fixing this. You can use the version check code in `utils.py` as a guide.

caspervdw:
@danielballan @tacaswell  Are you OK with bumping the pandas minimum version dependency to 0.13?

danielballan:
Fine by me. Should run this by @nkeim as well, who has weighed in on this in the past.

caspervdw:
I updated the dependency, @nkeim are you OK with this? Or do you want to keep the `>=0.12` in the `setup.py` and work with a condition in `emsd` only?

nkeim:
I'm fine with dropping 0.12 altogether. There don't seem to be any critical bugs to fix in the next minor release (fingers crossed), so folks will hopefully be able to upgrade at their leisure. And they can always override this in conda or pip.

On Jan 22, 2016, at 1:10 PM, Casper van der Wel <notifications@github.com<mailto:notifications@github.com>> wrote:


I updated the dependency, @nkeim<https://github.com/nkeim> are you OK with this? Or do you want to keep the >=0.12 in the setup.py and work with a condition in emsd only?

trackpy/motion.py
```diff
@@ -35,26 +36,120 @@ def msd(traj, mpp, fps, max_lagtime=100, detail=False, pos_columns=['x', 'y']):
...
+    http://stackoverflow.com/questions/34222272/computing-mean-square-displacement-using-python-and-fft#34222273.
+    The algorithm is described in this paper : http://dx.doi.org/10.1051/sfn/201112010.
+    """
+    if pos_columns is None:
+        pos_columns = ['x', 'y']
+    result_columns = ['<{}>'.format(p) for p in pos_columns] + \
+                     ['<{}^2>'.format(p) for p in pos_columns]
 
-def imsd(traj, mpp, fps, max_lagtime=100, statistic='msd', pos_columns=['x', 'y']):
+    r = traj[pos_columns].values * mpp
```
danielballan: This call to `.values` may be costly. That can cause pandas to copy the data, depending on the internal structure of the BlockManger. Fine with me if we investigate in a separate PR though.

danielballan:
I gave this a careful look and left one (optional) remark. Good to merge, @caspervdw. Thanks @hadim!

trackpy/motion.py
caspervdw: The DataFrame -> numpy conversion is currently necessary. It's also done in `_msd_gaps`. For instance at line 127 the code would fail if you don't do the conversion. (dataframes are aligned on index values)

We might try to do a full implementation on dataframes but it could be that the higher efficiency of numpy compensates for the copy on this line.

danielballan: Yes, that is believable. Also, assuming all the `pos_columns` are the same data type, copying should not be necessary and pandas is probably smart enough to know that.

PR #345, opened 2016-02-03
small compute_drift speed up

zoeith:
I suspect the list creation slows things down a bit. I didn’t do a full characterization but I got 5% speed up for 1000 frames and 15% at 10K frames. None of which is a big deal for something that takes <30s, but here we are.

### DIFF
--- a/trackpy/motion.py
+++ b/trackpy/motion.py
@@ -154,13 +154,14 @@ def compute_drift(traj, smoothing=0, pos_columns=['x', 'y']):
     >>> corrected_traj = subtract_drift(traj, drift) # Apply them.
     """
     # Probe by particle, take the difference between frames.
-    delta = pd.concat([t.set_index('frame', drop=False).diff()
-                       for p, t in traj.groupby('particle')])
+    delta = traj.groupby('particle').apply(lambda x :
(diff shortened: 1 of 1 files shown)


trackpy/motion.py
```diff
@@ -154,13 +154,14 @@ def compute_drift(traj, smoothing=0, pos_columns=['x', 'y']):
     >>> corrected_traj = subtract_drift(traj, drift) # Apply them.
     """
     # Probe by particle, take the difference between frames.
-    delta = pd.concat([t.set_index('frame', drop=False).diff()
-                       for p, t in traj.groupby('particle')])
+    delta = traj.groupby('particle').apply(lambda x :
```
tacaswell: minor comment, we follow pep8 so can you put spaces after the commas?  It seems silly, but having a consistent style really does help in code bases maintained by a group of people.

zoeith:
Cool. The pep8 linter gives a bunch of other issues too in motion.py (though mostly line length), so I wasn't really paying attention.

caspervdw:
Looks good to me. Similar approaches could be implemented for `imsd` and `emsd` I think. We should have a look some time. Good to merge, I leave it to @tacaswell .

Issue #297, opened 2015-10-18
ENH: Add least-squares feature identification


nkeim:
This is more of a long-term goal. It is possible to refine feature locations not by weighted averaging, but by least-squares fitting with a model of how all _n_ particles should appear. This was developed by Mark Shattuck's group for close granular packings; there is a tutorial (and Matlab code) [here](http://gibbs.engr.ccny.cuny.edu/technical/Tracking/ChiTrack.php). And you can also check out a [paper](http://link.aps.org/doi/10.1103/PhysRevE.75.051311) that uses the technique successfully.

I wanted to create an issue because I hope that someday a user will contribute this to trackpy. Maybe you!

caspervdw:
This would be a great addition! I already have a python code that uses `scipy.optimize.curve_fit` to fit a sum of N gaussians, using masking just as trackpy does. This works nicely for clusters of gaussian-like features (tested for N < 10).

A similar method called SFF: [article](http://www.sciencedirect.com/science/article/pii/S0001868607001200)

PR #350, opened 2016-03-04
Sum of Gaussian refinement

caspervdw:
This implements feature or clusters of feature refinement by least-squares fitting to sums of Gaussians. The two most important routines are:

- `find_clusters`: Identify 'Clusters' in a DataFrame of features. Clusters are groups of features of which each feature overlaps with at least one of the other features.
- `refine_gaussian`: Apply least-squares Gaussian fits to single particles or clusters.

There are a couple of helper functions and tests in there, too.

It should be easy to extend the code to feature shapes other than Gaussians, as long as it has only two shape parameters: `signal` and `size`. This partially addresses #297.

### DIFF
--- a/trackpy/api.py
+++ b/trackpy/api.py
@@ -17,9 +17,10 @@ from .linking import HashTable, TreeFinder, Point, PointND, \
 from .filtering import filter_stubs, filter_clusters, filter
 from .feature import locate, batch, percentile_threshold, local_maxima, \
            refine, estimate_mass, estimate_size, minmass_version_change
-from .preprocessing import bandpass
+from .preprocessing import bandpass, lowpass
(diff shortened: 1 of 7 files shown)


caspervdw:
Tests pass, but there seems to be an issue in Scipy 0.16: something with `libgfortran`. I raised the issue at scipy.

nkeim:
Awesome!! This may also be of interest to @dwieker.

caspervdw:
In the meantime, I greatly improved on this with constrained fitting and arbitrary feature shapes, so I am closing this in favour of a better (future) PR. Saves us an extra (tedious) review round.

We should actually decide how to (re)structure trackpy if we are going to add additions like these.

PR #351, opened 2016-03-06
FIX Pandas indexing error in subtract_drift

caspervdw:
By changing the indexing in #330, there was a bug introduced in `subtract_drift`:

    ValueError: cannot reindex from a duplicate axis

This solves the indexing problem by temporarily converting the dataframe to a MultiIndex using a context manager. I think this pattern might be used in other dataframe-related functions too.

### DIFF
--- a/trackpy/utils.py
+++ b/trackpy/utils.py
@@ -7,6 +7,7 @@ import functools
 import re
 import sys
 import warnings
+from contextlib import contextmanager
 from datetime import datetime, timedelta
 from distutils.version import LooseVersion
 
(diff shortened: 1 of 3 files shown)


caspervdw:
This clearly isn't working. Pandas is giving me a headache currently, I will look into it later. Unless there are suggestions from pandas experts ;)

danielballan:
I can take a look at this today / this weekend.

caspervdw:
That would be great!

trackpy/utils.py
```diff
@@ -288,3 +289,21 @@ def _pandas_sort_post_017(df, by, *args, **kwargs):
     pandas_sort = _pandas_sort_post_017
 else:
     pandas_sort = _pandas_sort_pre_017
+
+@contextmanager
```
danielballan: These are nice. You should consider contributing these upstream to pandas, @caspervdw.

caspervdw:
The tests for subtract drift where wrong! The problem was this: `add_drift` operated inplace and `subtract_drift` operated inplace:

     tp.subtract_drift(add_drift(self.dead_still, drift), drift)

There is never a new DataFrame created, so basically nothing is tested. Luckily the tested functions were OK (I checked on the master branch).

Because I changed the default behaviour of `subtract_drift` not work inplace (API break!) this problem appeared. Ready to merge before v0.3.1

caspervdw:
@danielballan This one can be merged, if you agree

caspervdw:
I think this should go in v0.3.1, could anyone merge this?

danielballan:
@caspervdw I agree. It needs a rebase first.

caspervdw:
Done!

caspervdw:
v0.3.1 is waiting for this, could you merge @danielballan?

Issue #349, opened 2016-02-28
Errorbars on ensemble MSD plot


vivarose:
I would like to see errorbars on the ensemble mean square displacement plot. There are a number of subtleties that could be taken into account. (For example, the coordinates that Trackpy assigns to each particle will have some error. For another example, the single particle MSD error must be calculated with the awareness that the data being averaged are not independent.) 

The easiest way to do it for an ensemble MSD is to calculate the standard deviation when taking the average of the single-particle trajectories.

I think that calculating and plotting errorbars should be built in, if not the default setting.

caspervdw:
Thanks for you request! I think that a simple error analysis can be based on the effective number of independent measurements, which is already calculated by `msd`. However, if you start to include tracking uncertainties, particle correlations etc., the error estimation becomes more difficult and very much user-specific.

I am not in favor of including those extra effects. A simple estimation would be fine, if we are clear about the simplified error model we use.

I do not have the time to look into this, but I can help you if you want to contribute to trackpy yourself!

vivarose:
I will definitely think about contributing to trackpy because I do think this would be helpful for me and for the community.

nkeim:
:smile: This would be a welcome addition! I agree that it should not be turned on by default, and it should get more than a few words in the documentation. Otherwise the possibility for a non-expert (i.e., me) to misinterpret the result could be too great.

vivarose:
My plan is to edit the emsd function so that it returns a standard deviation. This standard deviation shows how the single particle MSDs vary. Very simple, no subtleties.
Two initial questions (which can be changed later anyway):
1. Should emsd() always return the standard deviation? or only if detail=True?
2. What should the standard deviation be called? stdev, stddev, sigma, something else?

tacaswell:
I am strongly :-1: on kwargs changing the number/type of returned values.  If you need to break the API on `msd`, then either just break it and document it or (and I know @danielballan will disagree about this) make a new function (`msd_err`?, I am bad at naming things).

I would go with the numpy naming of `std`.

danielballan:
Welcome, @vivarose!

@tacaswell A reminder here that we already have this `detail=True` flag that @vivarose mentioned in (1). If I was doing `emd` over again, I would implement `emd_detailed` as a separate function. But I think (1) is probably the right way to go here with respect to consistency with the existing code. If we want to rethink that API, we should do it for *all* the detailed columns, in a separate pull request.

[...]

PR #352, opened 2016-03-08
ENH: Standard deviation calculation for emsd()

vivarose:
This adds the ability to calculate the standard deviation when calculating the ensemble mean square displacement with tp.motion.emsd(), as discussed at https://github.com/soft-matter/trackpy/issues/349


caspervdw:
Thanks @vivarose!  The code looks fine to me. The Travis fail is another issue that is not connected with this PR.

Just to be concise on a theoretical ground, do you want to show the standard deviation of the samples, or the standard deviation in the average of the samples? The code suggest the latter, but I am not completely sure if the approach is mathematically correct.

Do you have a reference to put in the Notes section of the emsd?

vivarose:
Hi @caspervdw, it is the standard deviation in the mean of the samples. 

I know it is possible to calculate the std for a single particle msd, but I have not familiarized myself with that yet. So this is the std of the mean over the single particle msds. 

And it's biased (without Bessel's correction). It would be straightforward to include Bessel's correction but the std would still be biased.

I have a notebook where I used the new emsd() function here: https://github.com/vivarose/using_trackpy/blob/master/MSD_standard_deviation.ipynb

I can get back to you about references.

caspervdw:
Thanks for the clarification. I think we have to get to the bottom of the theoretic part on this before we can merge this. I am afraid that by averaging the squared deviations of the msds, which are already averages, we underestimate the error. I hope some reference can clear that up.

In either case, the Bessel correction is only a minor one as we are talking about N >> 1.

tdimiduk:
I think there may be a Travis glitch on python 3.5, I grabbed @vivarose's code and it runs fine under anaconda python 3.5 on my local box.

vivarose:
I think the test failed because the new version of pandas, 0.18.0, is causing an error. I ran into this problem today when I updated pandas.

danielballan:
Hmm. The proximate cause is `np.round`, but that API has not changed.

```python
assert_series_equal(np.round(actual), expected)
  File "/home/travis/mc/envs/testenv/lib/python3.5/site-packages/numpy/core/fromnumeric.py", line 2786, in round_
    return round(decimals, out)
TypeError: round() takes from 1 to 2 positional arguments but 3 were given
```

Maybe something funny about how the Series objects `actual` and `expected` are being interpreted as arrays?

vivarose:
I think the soft-matter/master would also fail this travis build with Pandas 0.18.0.

danielballan:
@vivarose Definitely -- I agree these failures are unrelated to your PR. I'm just speculating about what could have changed in pandas....

danielballan:
I read the pandas release notes and nothing jumps out. I'll run these with `--pdb` tomorrow and try to learn more.

vivarose:
I have now integrated the ability to calculate the standard deviation when calculating the ensemble msd (emsd). 

Previously, I was pretending that each individual particle msd (imsd) has no error and calculating the standard deviation using frequency weights based on the N column, and @caspervdw was concerned that this would underestimate the standard deviation.

Now, in a separate branch (not yet pull-requested here), I have a new approach. I am calculating the imsd standard errors using the Flyvbjerg and Petersen approach, implemented by [pyblock](https://pypi.python.org/pypi/pyblock) and/or https://github.com/manoharan-lab/flyvbjerg-std-err. Then I can use the imsd stderr to calculate the emsd standard deviation, by using the inverse square of the imsd stderrs as reliability weights. 

I believe that by incorporating the imsd stderr in this way, we are obtaining a better estimate of the std of the emsd.

My branch for calculating the standard deviation is here:
https://github.com/vivarose/trackpy/tree/using_fp_stderr

I have only just arrived at this point so I'm not quite ready to pull-request it. In addition to testing it out a bit, there are a few more things I would like to do:
1) Learn more about _msd_fft() to see if I can calculate the fp_stderr in that function. Right now I can only do it in the _msd_gaps() function.
2) Make it more user-friendly to plot the imsd with error bars. (I have a way of doing it but it's not built-in yet.)

This does not account for errors in the tracking. Those errors are known to increase the msd, especially at small displacements.

Jmlannan:
Hello, sorry to post on such an old issue, but could the walkthrough be updated to show this functionality?
I believe it should looks something like this with the current version:

`
mpp = 100 #microns per pixel
fps =24
detailed_emsd = tp.emsd(t1, mpp, fps, detail=True)
fig, ax = plt.subplots()
ax.set_xscale('log')
ax.set_yscale('log')
ax.set(ylabel=r'$\langle \Delta r^2 \rangle$ [$\mu$m$^2$]', xlabel='lag time $t$')
plt.errorbar(detailed_emsd["lagt"].to_numpy(), 
    detailed_emsd["msd"].to_numpy(), 
     yerr = detailed_emsd["<y^2>"].to_numpy() + detailed_emsd["<x^2>"].to_numpy(), 
     )
plt.show() `

biased on: https://github.com/vivarose/using_trackpy/blob/master/MSD_standard_deviation.ipynb

vivarose:
@Jmlannan, no, that is not the correct errorbar. The official version of soft-matter/trackpy doesn't calculate the error. The "<y^2>" and "<x^2>" columns are not uncertainty, they are mean square displacement in y and x, respectively.

You are welcome to use my code, https://github.com/horowitz-lab/using_trackpy/blob/master/stdev_emsd.py, to calculate the standard deviation.

nkeim:
@vivarose Am I right that you or I could rebase this and it would be a viable PR? From this discussion and the one in #349 it was unclear whether you still recommend this code for general use.

PR #353, opened 2016-03-08
FIX Build issue with numpy 1.9.3 and scipy 0.16.0

caspervdw:
This fixes Travis build again, using the suggestion from https://github.com/scipy/scipy/issues/5917


caspervdw:
Merging own PR, as it is only Travis-build related

Issue #354, opened 2016-03-22
Upgrading to Pandas 0.18.0 causes trackpy errors


vivarose:
My travis build (https://github.com/soft-matter/trackpy/pull/352#issuecomment-199969738) failed.

Pandas 0.16.2 works with trackpy. I'm not sure whether Pandas 0.17 works.

caspervdw:
Thanks for the report! This is a pandas / numpy bug (https://github.com/pydata/pandas/issues/12600), luckily it doesn't break our code but only two of our tests: I think we can come up with a workaround for the tests.

PR #356, opened 2016-03-27
BLD: Avoid buggy Pandas 0.18.0

nkeim:
This is meant to address #354 and the problem in #352 by forbidding Pandas 0.18.0. See http://conda.pydata.org/docs/spec.html#build-version-spec


caspervdw:
Just to be clear here, 0.18 does not break the trackpy code, only the two tests using `np.round` on a `Series` object. Merging!

Issue #343, opened 2016-01-27
Upside down plotting


Resonanz:
I think that the trajectory plots flip my video frames so that the top is at the bottom and the bottom is at the top. e.g. [0,0] in the trajectory plot is bottom-left while [0,0] in my video frames is probably being taken as top-left.  This makes it confusing to follow which particles are doing what (yeah, sometimes this really is useful) so I am wondering if this is known or deliberate or just me? Thanks.

caspervdw:
There was an inconsistency between `annotate` and `plot_traj` in 0.2.4, which was solved in 0.3. In 0.2.4, annotate displays upside-down (origin at bottom left) and plot_traj displays correctly (origin at top left).

Are you using 0.2.4 and does upgrading to 0.3 help?

If not, could you try something like this and see if this works out for you:

    import matplotlib.pyplot as plt
    import trackpy as tp    
    ax = plt.gca()
    tp.plot_traj(...., ax=ax)
    ax.invert_yaxis(True)

caspervdw:
Closing through https://github.com/soft-matter/trackpy/pull/357

PR #357, opened 2016-03-28
FIX Plot fixes

### LINKED ISSUE #342: Trajectory plot
A difference between 0.2.4 and 0.3.0 is that in the latter a legend is placed over and obstructs a section of the trajectory plot. This is shown in Daniel's walkthru example also.  I'm not sure that the legend is necessary and certainly detracts from the plot.  Can it be removed or switched off?

caspervdw:
This should fix #342 and #343.
y-axis inversion is now done with a function, that first checks if it was inverted already. Legends are now not plotted by default.

@tacaswell, I just realized this crosses your older PR #178. Maybe we can work out a combination of both, yours has to be rebased anyway. There has to be some more cleaning in `plots.py`.

### DIFF
--- a/trackpy/plots.py
+++ b/trackpy/plots.py
@@ -104,7 +104,7 @@ def make_axes3d(func):
 def make_fig(func):
     """See make_axes."""
     import matplotlib.pyplot as plt
-    wraps(func)
+    @wraps(func)
     def wrapper(*args, **kwargs):
         if 'fig' not in kwargs:
             kwargs['fig'] = plt.gcf()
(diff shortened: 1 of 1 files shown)


tacaswell:
Don't worry about my existing PR, I'll rebase it once this is in.

trackpy/plots.py
```diff
@@ -46,16 +46,11 @@ def wrapper(*args, **kwargs):
...
-            else:
-                del kwargs['legend']
+            legend = kwargs.pop('legend', False)
             result = func(*args, **kwargs)
-            if not (kwargs['ax'].get_legend_handles_labels() == ([], []) or
-                    legend is False):
-                plt.legend(loc='best')
+            handles, labels = kwargs['ax'].get_legend_handles_labels()
+            if legend and len(labels) > 0:
+                plt.legend(handles, labels, loc='best')
```
tacaswell: should do `kwargs['ax'].legend` here.  Why is the legend only added if the user did not provide an explicit `ax`?

caspervdw: Agreed. I pushed the change. I am in for dropping the `legend` kwarg altogether. No reason to change the matplotlib API.
Maybe a DeprecationWarning for v0.3.1?

PR #358, opened 2016-04-12
Added particle labelling to plot.annotate().

hohlraum:
Particles can be automatically laballed by plot.annotate(). Also includes a style dictionary similar to that for the markers and background image.

### DIFF
--- a/trackpy/plots.py
+++ b/trackpy/plots.py
@@ -392,9 +392,9 @@ def plot_traj3d(*args, **kwargs):
 ptraj3d = plot_traj3d
 
 @make_axes
-def annotate(centroids, image, circle_size=None, color=None,
+def annotate(centroids, image, circle_size=None, label=False, color=None,
              invert=False, ax=None, split_category=None, split_thresh=None,
-             imshow_style={}, plot_style={}):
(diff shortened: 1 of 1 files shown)


danielballan:
Nice!

I notice this could be made more flexible by accepting a string argument for `label`. So, for example, if I wanted to label particles by their mass, I would pass `label='mass'`. But if you don't feel inspired to add that to this PR, we can consider this enhancement separately in a backward-compatible way.

trackpy/plots.py
```diff
@@ -392,10 +392,10 @@ def plot_traj3d(*args, **kwargs):
 ptraj3d = plot_traj3d
 
 @make_axes
-def annotate(centroids, image, circle_size=None, color=None,
+def annotate(centroids, image, circle_size=None, label=False, color=None,
```
tacaswell: please put new kwargs at the end of the list (think of the case where someone is using `annotate(cent, img, 5, 'b')` this is now silently (as in the code _runs_) but does something different).

caspervdw:
@hohlraum Are you planning to update the PR? It would be nice to have this addition in the next release.

danielballan:
@hohlraum Any chance you'll be able to revisit this?

PR #359, opened 2016-04-24
CLN: Moved empty-level check to wrap kdtree.query()

nkeim:
This moves the check for empty levels into `assign_candidates`, because the check addresses a shortcoming of cKDTree specifically. See https://github.com/soft-matter/trackpy/pull/293#issuecomment-214027913 and https://github.com/soft-matter/trackpy/pull/293#discussion_r43029351 for motivation.

### DIFF
--- a/trackpy/linking.py
+++ b/trackpy/linking.py
@@ -1040,9 +1040,8 @@ class Linker(object):
                 p.forward_cands = []
 
             # Sort out what can go to what.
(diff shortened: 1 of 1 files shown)


nkeim:
Please ignore the first failed Travis check. I accidentally pushed the branch to `soft-matter/trackpy`, then deleted it, which caused Travis's pull to fail.

caspervdw:
Looks good, merging

PR #361, opened 2016-04-26
PERF Grey dilation

caspervdw:
As discussed, here a different `local_maxima` that does a grey dilation with a rectangular kernel, instead of a square one. Directly after, it eliminates duplicates using a KDTree.

I also refactored the code a bit in view of #257  : the `find_maxima` is renamed and moved to `find.py`.

### DIFF
--- a/trackpy/feature.py
+++ b/trackpy/feature.py
@@ -547,10 +547,10 @@ def locate(raw_image, diameter, minmass=None, maxsize=None, separation=None,
     #   - Invalid output of the bandpass step ("smoothing_size")
     margin = tuple([max(rad, sep // 2 - 1, sm // 2) for (rad, sep, sm) in
                     zip(radius, separation, smoothing_size)])
-    # Find features with minimum separation distance of `diameter`. This enables
-    # detection of small features close to large, bright features using the
-    # `maxsize` argument.
-    coords = grey_dilation(image, diameter, percentile, margin, precise=False)
+    # Find features with minimum separation distance of `separation`. This
+    # excludes detection of small features close to large, bright features
+    # using the `maxsize` argument.
+    coords = grey_dilation(image, separation, percentile, margin, precise=False)
     count_maxima = coords.shape[0]
 
     if count_maxima == 0:
--- a/trackpy/feature.py
+++ b/trackpy/feature.py
@@ -82,20 +82,23 @@ def minmass_version_change(raw_image, old_minmass, preprocess=True,
 
 def local_maxima(image, radius, percentile=64, margin=None):
     """Find local maxima whose brightness is above a given percentile.
+    This function has been deprecated. Please use the routines in trackpy.find,
+    with the minimum separation between features as second argument.
 
     Parameters
     ----------
     image : ndarray
         For best performance, provide an integer-type array. If the type is not
         of integer-type, the image will be normalized and coerced to uint8.
-    radius : integer definition of "local" in "local maxima"
+    radius : the radius of the circular grey dilation kernel, which is the
+        minimum separation between maxima
     percentile : chooses minimum grayscale value for a local maximum
     margin : zone of exclusion at edges of image. Defaults to radius.
             A smarter value is set by locate().
     """
     warnings.warn("Local_maxima will be deprecated: please use routines in "
                   "trackpy.find", DeprecationWarning)
-    return grey_dilation(image, [r * 2 + 1 for r in radius], percentile, margin)
+    return grey_dilation(image, radius, percentile, margin)
 
 
 def estimate_mass(image, radius, coord):
@@ -544,7 +547,10 @@ def locate(raw_image, diameter, minmass=None, maxsize=None, separation=None,
     #   - Invalid output of the bandpass step ("smoothing_size")
     margin = tuple([max(rad, sep // 2 - 1, sm // 2) for (rad, sep, sm) in
                     zip(radius, separation, smoothing_size)])
-    coords = grey_dilation(image, separation, percentile, margin)
+    # Find features with minimum separation distance of `diameter`. This enables
+    # detection of small features close to large, bright features using the
+    # `maxsize` argument.
+    coords = grey_dilation(image, diameter, percentile, margin, precise=False)
     count_maxima = coords.shape[0]
 
     if count_maxima == 0:
--- a/trackpy/feature.py
+++ b/trackpy/feature.py
@@ -10,9 +10,10 @@ from scipy import ndimage
 from pandas import DataFrame
 
 from .preprocessing import (bandpass, convert_to_int, invert_image,
-                            scale_to_gamut, scalefactor_to_gamut)
-from .utils import record_meta, validate_tuple, cKDTree
-from .find import grey_dilation
+                            scalefactor_to_gamut)
+from .utils import record_meta, validate_tuple
+from .find import grey_dilation, where_close
+
 from .masks import (binary_mask, N_binary_mask, r_squared_mask,
                     x_squared_masks, cosmask, sinmask)
 from .uncertainty import _static_error, measure_noise
@@ -257,28 +258,10 @@ def refine(raw_image, image, radius, coords, separation=0, max_iterations=10,
     # Flat peaks return multiple nearby maxima. Eliminate duplicates.
     if np.all(np.greater(separation, 0)):
         mass_index = image.ndim  # i.e., index of the 'mass' column
-        while True:
-            # Rescale positions, so that pairs are identified below a distance
-            # of 1. Do so every iteration (room for improvement?)
-            positions = results[:, :mass_index]/list(reversed(separation))
-            mass = results[:, mass_index]
-            duplicates = cKDTree(positions, 30).query_pairs(1)
-            if len(duplicates) == 0:
-                break
-            to_drop = []
-            for p0, p1 in duplicates:
-                # Drop the dimmer one.
-                m0, m1 = mass[p0], mass[p1]
-                if m0 < m1:
-                    to_drop.append(p0)
-                elif m0 > m1:
-                    to_drop.append(p1)
-                else:
-                    # Rare corner case: a tie!
-                    # Break ties by sorting by sum of coordinates, to avoid
-                    # any randomness resulting from cKDTree returning a set.
-                    to_drop.append([p0, p1][np.argmin(np.sum(positions.take([p0, p1], 0), 1))])
-            results = np.delete(results, to_drop, 0)
+        positions = results[:, :mass_index]
+        mass = results[:, mass_index]
+        to_drop = where_close(positions, list(reversed(separation)), mass)
+        results = np.delete(results, to_drop, 0)
 
     return results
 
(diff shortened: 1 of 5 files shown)


nkeim:
Excellent.

Notes on the ongoing performance comparison with MATLAB, for sparsely-populated frames:

- Number of features found went from 546 to 554. The 2-norm metric used by KDtree ≠ metric in pixellated space. But a ~1% difference is probably acceptable.
- Feature-finding time went from 0.106 s/frame to 0.0517 s/frame (!). MATLAB's current best time is 0.0458 s. A full 10 ms of trackpy's time is still used by pims' `as_grey`.
- The final product *looks* worse with the same parameters. For example, I draw your attention to the lower right corner of each image:

Before

benchmarking_sparse_trackpy_after_e20c6c4cf2f46966d92a04e2f50acb62975049ff

After

sparse_trackpy

So there are some issues to address, but this is a very successful experiment!

trackpy/feature.py
```diff
@@ -284,28 +248,10 @@ def refine(raw_image, image, radius, coords, separation=0, max_iterations=10,
     # Flat peaks return multiple nearby maxima. Eliminate duplicates.
     if np.all(np.greater(separation, 0)):
         mass_index = image.ndim  # i.e., index of the 'mass' column
-        while True:
```
nkeim: The new `where_close()` implementation is not iterative. I think that we previously managed to convince ourselves that multiple iterations may be necessary to resolve all close pairs, but I don't recall the detailed argument. I am trying to think of a test we could write to find out.

nkeim:
Bad news: this slows down the dense benchmark by about 25x. Profiler output for 10 frames:

```
         23297539 function calls (23297159 primitive calls) in 85.399 seconds

   Ordered by: internal time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
       20   50.948    2.547   80.919    4.046 find.py:18(where_close)
       20   16.450    0.823   16.450    0.823 {method 'query_pairs' of 'ckdtree.cKDTree' objects}
 11638432    4.137    0.000    4.137    0.000 find.py:34(<genexpr>)
 11638432    4.068    0.000    4.068    0.000 find.py:33(<genexpr>)
       40    4.012    0.100   12.217    0.305 {numpy.core.multiarray.fromiter}
       10    2.247    0.225   82.594    8.259 find.py:53(drop_close)
       20    0.818    0.041    0.818    0.041 {method 'sort' of 'numpy.ndarray' objects}
       10    0.397    0.040   83.484    8.348 find.py:70(grey_dilation)
       58    0.346    0.006    0.346    0.006 {numpy.core.multiarray.where}
       20    0.324    0.016    0.324    0.016 {scipy.ndimage._nd_image.correlate1d}
       20    0.287    0.014    0.287    0.014 {scipy.ndimage._nd_image.uniform_filter1d}
```

Given the badness of this worst-case scenario, I think we should find a way to make this new method optional.

caspervdw:
Sorry about the unintended close.
That's the reason why I changed `where_close`. The iterative algorithm was even slower than this. I convinced myself that iterations are not necessary but tests are certainly a good idea.

Is the benchmark a real image or a random noise?

caspervdw:
@nkeim I discovered a bug that might be crucial for the benchmark. I can't manage to put in a line note on my phone. The call to `grey_dilation` in `locate` should have `diameter` as second arg.

You coulf even argue to put in `separation` as second arg, but that's a big change.

nkeim:
Ah! With that bugfix made, number of features now goes from 546 (old method) to 351 (corrected new method). So maybe there is an off-by-one error somewhere.

The dense benchmark is still 10x slower than before. This is an image of close-packed repulsive particles, with typical spacing just a little bigger than feature size. So it is a worst case for this algorithm, but one I would argue is not uncommon.

I am (slowly) looking at alternatives to cKDTree for this. The precision and isotropy it offers are not really necessary at this stage. I like the Blair-Dufresne code's approach of doing "sparse grey dilation" by using the previously-found peak coordinates. In our case we only need to check the pixels outside the rectangular `grey_dilation` footprint. Numba can make that very fast.

nkeim:
Also, I remember now that @danielballan and I originally decided to do multiple iterations because we were doing peak *merging* — taking the average position of nearby peaks, rather than rejecting the dimmer ones.

caspervdw:
Apart from the whole `where_close` performance thing (which can be better) I think I understand why you got 546 features with the current method and 351 with the new one.

In fact the current implementation gives unexpected results:

    tp.find.grey_dilation_legacy(im, radius=7)

Would find features upto a separation distance of 7, whereas `locate` expects 15. 
This is because of the following: if you want to find maxima that are spaced 15 from each other, you should do a grey dilation with a circular kernel with a *radius* of 15. It is easy to see in a drawing.

I got it wrong too in the new implementation, so that the KDTree had to do a lot of work.

caspervdw:
@nkeim, could you check performance again? I made the fixes and I turned off the KDTree filtering in locate, possibly it leads to an improvement even for the dense case.

I am also curious what would happen if you turn in back on, this would mean less features to refine at the expense of an extra `drop_close`.

nkeim:
Wow. The dense benchmark is now actually *faster* than the v0.3 code — 0.24 s/frame, down from 0.30. The sparse benchmark is still about twice as fast as v0.3, at 0.050 s/frame. Awesome!

I think what @danielballan and I have learned previously is that when using the numba code, characterizing extra features costs almost nothing. So I think it is right to do only the final `drop_close()`. Note, however, that this introduces a new coupling between the peak-finding and refinement steps at a time when we are trying to decouple them. So maybe the default should be to perform KDTree filtering in `grey_dilation`, with an option to turn it off if you promise to do it later?

There is still the disagreement in the results. In the figure, MATLAB is on the left and this new trackpy code is on the right.

unknown

It looks like I can bring these back into agreement by changing the `percentile` parameter (another parameter for peak-finding). But I first want to make sure I understand what has changed. Am I right that

1. In v0.3, trackpy uses the feature *radius* for grey dilation, so that the overlap between feature ROIs can be as much as one radius; i.e., the minimum separation is one radius. KDTree filtering is applied *after* refinement, and controlled by the `separation` parameter, to set the final separation.
2. In this new code, `diameter` sets the size of the grey dilation kernel. There is still only one KDTree step, applied after refinement.
3. In the Blair-Dufresne code, square kernels are used everywhere, with size specified by a `separation`-like parameter that is independent of `diameter`. Effectively, there is no KDTree step. (I am fairly certain of this based on my reading of the code.)

Is this more or less correct?

Should `separation` set the size of the grey dilation kernel in trackpy? Conceptually, ROI size (for refinement) and neighborhood size (for peak finding and final separation) can be independent, as they are in the Blair-Dufresne code. Being able to set `separation` < `diameter` is useful when it's more important to find all the particles than to know their positions accurately.

caspervdw:
@nkeim, I think your description is entirely correct (although I have not checked the Blair-Dufresne code in depth)

The most logical choice for the length scale in `grey_dilation` would be `separation`. I cannot come up with any practical use to do it differently. The only thing is that it may lead to issues when combined with `maxsize` filtering (see the failed test `test_minmass_maxsize`, https://travis-ci.org/soft-matter/trackpy/jobs/128360102).

Basically, small features next to large features with size > maxsize will not be detected in the new algorithm, while they would be deteced in the old algorithm. The issue here is a wrong choice of `separation`: when it is much larger than feature `diameter` (as in the test), this leads to the issue.

So I would still go for `separation` and document the slight API change.

danielballan:
That sounds reasonable to me. Users with particular needs (and a very detailed understanding!) can always roll their own locate, passing around length scales manually, right?

caspervdw:
I think this PR is ready to merge. We just need a check whether this PR changes anything in the result of `locate` (it should not, except in some edge cases where size filtering is used)

@nkeim is there a separation parameter in the blair-dufresne code?

nkeim:
There is a separation parameter in Blair-Dufresne but it is applied after peak-finding, not after calculating the centroids. So that is a minor difference. I think that it is better to apply it at the end (as in this PR), because if two "features" are part of the same extended feature, they will migrate toward each other during refinement, making them more likely to get culled.

I would like to do the comparison one more time before merge and observe how the output has/hasn't changed. I can probably do that this weekend.

So will the next release of trackpy be v0.4? I think that is an appropriate increment when a core algorithm is altered.

caspervdw:
Maybe we can do v0.3.1 first and bundle this with #257 in v0.4
About the separation, it looked as if the extra features in the matlab sparse test are close to other features. We should just make sure that the separation (= diameter + 1 by default in trackpy) parameter is the same in the matlab code.

nkeim:
The issues I was encountering in the benchmark appear resolved. The large reduction in the number of features had nothing to do with peak-finding, but instead came from the way I used `np.mean()` to do RGBA ➔ grayscale conversion quickly (like in MATLAB).

After switching both MATLAB and trackpy to use `uint8` pixels, I'm getting 49.0 ms for trackpy and 40.1 ms for Blair-Dufrense. So this is very good performance. I cannot get the features to match, and it seems to have to do with square footprint vs. circular effective footprint for finding neighboring peaks.

Just as importantly, dense feature-finding is slightly faster than before (still about 5x MATLAB).

caspervdw:
Thanks for the work, it is looking very promising indeed. About the differences with master.. I am sure I tested the new `where_close` code against the old one on random coordinates, it was fine. I think that the few more peaks that are found in `grey_dilation` result in different peaks that are dropped eventually. This because the current (PR and master) peak-merging code leads to behavior that is sensitive to slight changes in `grey_dilation`. An example:

    peak 1, mass 20  --- peak 2, mass 40 -- peak 3, mass 30

Results in peaks 1 and 3 that are dropped, while:

    peak 1, mass 20 ---- (no peak 2)    -- peak 3, mass 30

Will result in no peak dropped.

Maybe we should try to define the expected result of `where_close` more accurately.

caspervdw:
I will roll a rebase after #400 .

caspervdw:
Did the rebase!

caspervdw:
@tacaswell  Do you understand why matplotlib is failing in the Py3.5 test? Is there something wrong in the conda dependencies?

    ImportError: No module named 'PyQt4'

caspervdw:
I think everything is fine for merge with this PR.

@nkeim, if you still have the code around, maybe you can do the comparative test with master again. The rebase was quite messy.

I think that we need a test that compares the results of a few frames in the demo image (like the `bulk_water` test that Nathan did) with a known table of feature coordinates. Just to consciously make the version-to-version breaks in the output results (like this one and #377)

nkeim:
OK. I think I can check this out today.

nkeim:
The rebase did not change the output. And performance is still excellent. 😌 

Before merge, I want to re-emphasize our above discussion, that this will change the output rarely but a little unpredictably. We are definitely drawing a bright line between v0.3 and v0.4.

@caspervdw I'd like to understand your proposal better. Basically,

1. Take a "messy" image (like from `bulk_water`) and add it to the repo.
2. Add a script to find features and output to a file (CSV is probably best).
3. Add the file to the repo. Any output-breaking changes will result in a failing test and require re-running the script.

Is that what you mean? It wouldn't be comprehensive but it would help us catch mistakes early.

caspervdw:
@nkeim Thanks for merging. The test you describe is indeed what I had in mind. The image data is already on the repo: we would only need a csv file of coordinates to test against.

PR #362, opened 2016-04-29
DOC Automatic doc build

caspervdw:
Implementation the same as for pims, see https://github.com/soft-matter/pims/pull/231 and https://github.com/soft-matter/pims/issues/232 . It should work (only) when it is merged. There is an encrypted deploy key for the doc repo that only Travis can decrypt.


tacaswell:
I am slightly concerned about the effect this will have on the size of the doc repo.  We do not really care about the history of the built docs.

caspervdw:
Totally agreed, but I first like to see it working before wiping the history automatically. So that is for a next step.

tacaswell:
:+1: I will let @danielballan push the merge button because he :heart_eyes: docs so much.

danielballan:
Now we'll see if Travis is happy....

caspervdw:
Not yet... https://travis-ci.org/soft-matter/trackpy/jobs/127663350

1) I apparently got the doc build folder wrong
2) There is a large number of warnings in the doc build. Do the tutorials have to be copied in manually?

caspervdw:
Fixed (hopefully) #363

PR #364, opened 2016-05-04
BLD Adjust paths in doc build

caspervdw:
This will hopefully solve the persistent issues with the automatic doc build.


caspervdw:
Ok, still not there, we will need pandoc on Travis to get the nbconvert working.

tacaswell:
If you are only touching travis / docs feel free to self merge.

PR #366, opened 2016-05-06
BLD Forced doc update

caspervdw:
This lets Travis wipe the doc history on each new build and finishes the automatic doc builds.

Generating the tutorials was a pain, I succeeded doing it by adapting the makefile for the notebooks (@danielballan). It now generates an html instead of an rst, together with a new rst that includes the html (see #364). In this way, the notebook gets converted to html in one step. I think this is better, but additionally it removes the need for pandoc on travis, which was problematic.

See for instance http://soft-matter.github.io/trackpy/dev/tutorial/walkthrough.html of the automatically generated notebook. The annoying thing is this `-e walkthrough` at the top, I can't get rid of it because Sphinx requires a title for each document.

I had to manually insert neat titles for each tutorial notebook, please check http://soft-matter.github.io/trackpy/dev/ and say if you agree on the titles.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -72,6 +72,9 @@ after_success:
         git checkout --orphan temp_branch
         git rm -rf ./dev
         mv $TRAVIS_BUILD_DIR/doc/_build/html ./dev
+        if [ -n "$TRAVIS_TAG" ]; then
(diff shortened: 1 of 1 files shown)


danielballan:
Titles looks good. Thanks, @caspervdw.

PR #375, opened 2016-06-07
DOC New zenodo badge and doc reference

caspervdw:
We have to start doing before the release or else it will not be incorporated in the doc build. I am actually in favor of doing this manually anyway because it allows for correctly mentioning all the contributors. I turned the zenodo webhooks off for now.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 trackpy
 =======
 
-[![build status](https://travis-ci.org/soft-matter/trackpy.png?branch=master)](https://travis-ci.org/soft-matter/trackpy) [![DOI](https://zenodo.org/badge/doi/10.5281/zenodo.34028.svg)](http://dx.doi.org/10.5281/zenodo.34028)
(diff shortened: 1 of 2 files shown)


tacaswell:
There is a boot-strap problem here isn't there?  You can not get a DOI until you tag, but if you want the DOI to be in the tag you have to get it before you tag...

The way we deal with this on MPL is to have an additional `tag-doc` branch which you checkout, update docs, install the tagged version and then build the docs.

caspervdw:
Exactly.

I don't completely understand how the `tag-doc` approach works. Do you have a recipe around?

Anyway, it might be the fastest to do the Zenodo manually (reserve the DOI before the release, add it via a PR a la this one, do the release, finalize the Zenodo entry)

tacaswell:
Oh, you can reserve a DOI ahead of time?  That does seem like a better tactic.

danielballan:
Definitely. I remember asking Arfon whether that was possible. Is it possible now?

caspervdw:
Yes it is, but you'll have to do it manually

PR #377, opened 2016-06-13
FIX Major bugfix in python refinement

### LINKED ISSUE #376: Refinement mismatch of python engine
Because I was running some tests on a Gaussian-refinement method, I was testing the rms error in the refinement on artificial images. There are some worrying results here: the python engine performs much worse than the numba engine! Also, I showed that bandpassing really helps.

Parameters are:
 -  feature shape is hat size, which is a solid disc with gaussian shaped borders
 -  `size`, the value of sigma in the gaussian function, default 4
 -  `disc_size`, the size of the solid disc, relative to `size`, default 0.3 (0 will make it exactly gaussian)
 -  `signal`, maximum value of feature, fixed at 100
 -  `noise`, added noise is uniformly distributed (0, `noise`), default 10
 -  `diameter`, always large enough (`int(4*size) + 1`)
refine_trackpy_tests

The notebook that I used to generate this can be found here (note to self: make a PR of the better artificial image codes):
https://gist.github.com/caspervdw/b439e5ecd62662582b97307dec79d1ad

caspervdw:
This removes any discrepancy between the python and numba engines:

 - the neighborhood should be calculated from rounded coordinates
 - the ndimage interpolation makes the code different from the numba engine and appears to add nothing in my tests

Fixes #376

### DIFF
--- a/trackpy/feature.py
+++ b/trackpy/feature.py
@@ -316,7 +316,7 @@ def _refine(raw_image, image, radius, coords, max_iterations,
 
     ndim = image.ndim
     isotropic = np.all(radius[1:] == radius[:-1])
-    mask = binary_mask(radius, ndim)
+    mask = binary_mask(radius, ndim).astype(np.uint8)
     slices = [[slice(int(round(c - rad)), int(round(c + rad + 1)))
                for c, rad in zip(coord, radius)]
               for coord in coords]
--- a/trackpy/feature.py
+++ b/trackpy/feature.py
@@ -317,7 +317,8 @@ def _refine(raw_image, image, radius, coords, max_iterations,
     ndim = image.ndim
     isotropic = np.all(radius[1:] == radius[:-1])
     mask = binary_mask(radius, ndim)
-    slices = [[slice(c - rad, c + rad + 1) for c, rad in zip(coord, radius)]
+    slices = [[slice(int(round(c - rad)), int(round(c + rad + 1)))
+               for c, rad in zip(coord, radius)]
               for coord in coords]
 
     # Declare arrays that we will fill iteratively through loop.
@@ -361,18 +362,19 @@ def _refine(raw_image, image, radius, coords, max_iterations,
                 upper_bound = np.array(image.shape) - 1 - radius
                 new_coord = np.clip(new_coord, radius, upper_bound).astype(int)
                 # Update slice to shifted position.
-                rect = [slice(c - rad, c + rad + 1)
+                rect = [slice(int(round(c - rad)), int(round(c + rad + 1)))
                         for c, rad in zip(new_coord, radius)]
                 neighborhood = mask*image[rect]
 
             # If we're off by less than half a pixel, interpolate.
             else:
-                # Here, coord is a float. We are off the grid.
-                neighborhood = ndimage.shift(neighborhood, -off_center,
-                                             order=2, mode='constant', cval=0)
-                new_coord = coord + off_center
-                # Disallow any whole-pixels moves on future iterations.
-                allow_moves = False
+                break
+                # # Here, coord is a float. We are off the grid.
+                # neighborhood = ndimage.shift(neighborhood, -off_center,
+                #                              order=2, mode='constant', cval=0)
+                # new_coord = coord + off_center
+                # # Disallow any whole-pixels moves on future iterations.
+                # allow_moves = False
 
             cm_n = _safe_center_of_mass(neighborhood, radius, ogrid)  # neighborhood
             cm_i = cm_n - radius + new_coord  # image coords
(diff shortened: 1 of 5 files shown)


tacaswell:
:+1:  makes sense to me.  @danielballan or @nkeim should merge this as I am only passingly familiar with this code.

caspervdw:
In view of the API change this will give, some more testing data below. Apparently, the numba code was made more accurate by #242 without us realizing it. Now the python code is also fixed.

I guess this pure python refine could use some streamlining, but that is for another time.

this PR:
refine_trackpy_tests_pr

v0.3.0:
refine_trackpy_tests_master

v0.2.4:
refine_trackpy_tests_v0 2 4

nkeim:
@caspervdw This is great!! Thank you for tracking this down.

Your test results suggest that there are reliable upper bounds on positional error. Do you think this would work as an automated test? At this point I think that creating an issue would suffice.

caspervdw:
I was thinking the same. There is a test already, `test_ep`. I wonder why it passed. Maybe we need to sharpen the tolerances.

danielballan:
Probably. If you have the code that generated those plots handy, maybe we should start a stash of "human-in-the-loop" tests to be run before releases and on PRs that touch critical code.

caspervdw:
Shall we add this to v0.3.1?

danielballan:
Is the signal-to-noise ratio on the plots actually a noise-to-signal ratio? The trends are the opposite of what I would expect, unless I am confused.

My vote is that this is a bug fix, not an API change, and thus can go into v0.3.1 (not v0.4.0).

tacaswell:
I am also :+1: on this being a bugfix, not an API change.

caspervdw:
@danielballan  Yes it should be noise-to-signal ratio... I will have a quick look at `test_ep` today so that we can merge soon.

caspervdw:
Done, ready to merge

nkeim:
Sorry for taking a closer look so late. It looks like this removes the shift code that was the principal difference between the python and numba engines, going all the way back to the beginning. I had always thought that this made the python engine slightly superior! But the data don't lie.

caspervdw:
I'll quickly run a test today to check if there is a difference @nkeim

caspervdw:
Apparently, it really doesn't make a difference whether you interpolate or not:

refine_trackpy_tests

Or, the differences with numba (negative = better, positive = worse):
refine_trackpy_iterpolation_tests

danielballan:
Interesting! I had wondered about that too. It's in the original Crocker-Grier implementation but, IIRC correctly, not in Blair's code. I wonder if Crocker tried it without interpolation first or what motivated that in the first place.

trackpy/feature.py
```diff
@@ -361,18 +362,19 @@ def _refine(raw_image, image, radius, coords, max_iterations,
...
-                new_coord = coord + off_center
-                # Disallow any whole-pixels moves on future iterations.
-                allow_moves = False
+                break
+                # # Here, coord is a float. We are off the grid.
+                # neighborhood = ndimage.shift(neighborhood, -off_center,
+                #                              order=2, mode='constant', cval=0)
+                # new_coord = coord + off_center
+                # # Disallow any whole-pixels moves on future iterations.
+                # allow_moves = False
```
danielballan: All references to `allow_moves` can be removed. It will always be `True`.

caspervdw: I'll look at it and refactor. Also the `shift_thresh` and `break_thresh` will be the same thing: if the code does not shift, it will break. Calculating the CoM without shifting makes no sense because the outcome will not change.

caspervdw:
Done the refactor at @danielballan suggestion. Because the interpolation went out, the refinement loop could be simplified substantially: `break_thresh` and  `shift_thresh` both resulted in stopping of the refinement loop. `break_thresh` basically did not have a function anymore. I dropped it as parameter for `refine`, I have added a deprecation warning.

I ran the tests again and the performance is better than before. Maybe I solved another bug: the `off_center` might have been subtracted one time too much, previously. But I am not sure of the origin of this difference. It is for sure much better at low noise values.

refine_trackpy_tests_refactored

caspervdw:
The test should be fixed now, the resulting tracking errors are now lower, agreeing much better (within 10%) with the theoretical ones.

caspervdw:
Could anyone check this and merge? The amount of code that is changed is substantial, but also fairly straightforward. I think that this increase of accuracy should be released asap so that people can make use of it.

Potentially as a v0.4, not v0.3.1? What do you think? This PR will result in different refined coordinates.

nkeim:
I suspect that the "python" engine is used by only a small minority. There is zero downside to this PR except that it changes output. And putting out a release is a fair amount of work. So I vote to include it in 0.3.1. I would like to focus on #361 (and maybe #333) for v0.4.

nkeim:
I've looked at the code and am happy with it. If I had more time I would test this with my application (1D feature finding). But if there is a problem with that very rare use case, I will find it soon enough.

So I also want to merge if we are OK with this being in v0.3.1.

danielballan:
Yep, this is a bugfix and can go into v0.3.1. Merging!

PR #380, opened 2016-07-06
Fixes before v0.3.1

caspervdw:
While checking the walkthroughs, I encountered some minor bugs, see commit messages.
I reverted most of the work in #351 (sorry @danielballan , my bad)


danielballan:
Travis failed passed with pandas==0.18 but failed on previous versions. My shift is ending here on the east coast, but I'll take a look tomorrow.

caspervdw:
Would be great! I am getting a bit fed up with this `subtract_drift` routine. It's probably the sorting by frame that triggers the test, curious that it does not fail on 0.18.

danielballan:
See https://github.com/caspervdw/trackpy/pull/3

The worst part was having to download really old versions of everything to reproduce.

caspervdw:
Thanks! Ready for merge

Issue #381, opened 2016-07-11
Try pre-compiling numba code


danielballan:
I just learned that numba now supports ahead-of-time compilation, such that numba-generated code can be distributed in a package with only numpy as a dependency -- no numba or no LLVM for necessary for users.

I see two benefits (1) obviously, this makes it easier for non-conda users to enjoy numba-powered speedups and (2) I think it frees us to rely on bleeding-edge numba features.

zoeith:
Heads up though:
```python
...
~/anaconda/lib/python2.7/site-packages/trackpy/try_numba.py in <module>()
     33 else:
     34     v = numba.__version__
---> 35     major, minor, micro = v.split('.')
     36     if major == '0' and minor == '12' and micro == '0':
     37         # Importing numba code will take forever. Disable numba.

ValueError: too many values to unpack

In [2]: import numba

In [3]: numba.__version__
Out[3]: '0.27.0+0.g6ca78ce.dirty'
```

the current builds on conda (at least for mac osx 64 and linux 64) have a weird versioning.

short term workaround:
```
$ conda install numba=0.26
```

caspervdw:
@zoeith  I made a PR with a fix for your issue. Could you check and confirm that it is fixed?

@danielballan  Would this precompiling be platform-specific?  And what is the advantage of numba over cython if you precompile anyway?

danielballan:
Fair question. Numba can beat Cython performance -- sometimes on one core and especially with the multicore and GPU targets (which we should expose through our API...). And I'd argue that numba-optimized Python is more readable than Cython to a user who doesn't know either of those.

On the whole, Cython is still probably the better, more stable choice for an important project. If I recall correctly, @nkeim and I started the numba implementations on a whim to experiment with numba, and they stuck.

danielballan:
@zoeith I was going to follow up with the numba or conda devs as a courtesy -- they probably didn't intentionally publish a package with a dirty tag -- but I can't reproduce finding that weird build on OSX. 

`conda search numba | grep dirty` turns up nothing.

When I `conda install numba` and obtain version 0.27.0, the verison string looks fine:

```python
In [1]: import numba

In [2]: numba.__version__
Out[2]: '0.27.0'
```

Can you tell me what channel that badly-tagged numba package is coming from?

Of course, trackpy should support non-tagged builds of numba, so the fix is a good idea either way.

zoeith:
@danielballan I just did `conda install numba` ... that should be the default channel?
[...]

PR #382, opened 2016-07-12
FIX Numba version conditional

caspervdw:
Fixes the issue reported by @zoeith  in #381


zoeith:
Worked for me. Good to learn about LooseVersion.
Cool, thanks.

danielballan:
Thanks for the report, @zoeith.

Issue #355, opened 2016-03-24
Inconsistent behaviour with preprocessing(bandpass) of frames


rfhigler:
The following to code snippets give entirely different results even though functionally they are identical. The only difference is the whether I call tp.bandpass() before I do tp.locate() or whether I use tp.locate() to call tp.bandpass() for me.

`frame_bp = tp.bandpass(frames[0], [1, 1, 1], [13, 13, 13], threshold=None, truncate=4)`
`test_loc2 = tp.locate(frame_bp, diameter=(13,13,13), separation=(9,9,9), engine='numba',preprocess=False)`

`test_loc = tp.locate(frames[0], diameter=(13,13,13), separation=(9,9,9), engine='numba', noise_size=[1,1,1], smoothing_size=[13,13,13], threshold=None)`

The first method gives the expected results and find all features perfectly, while the bottom approach mainly tries to locate noise. All frames are 3D stacks with a shape of (307,307,101)(x,y,z), are uint16s and have a couple of large, bright white, and spherical features against a black background. Perfect locating material. [Example slice](http://imgur.com/r8vlWNz).

I believe the bug is located in file feature.py, line 567:
`if np.issubdtype(raw_image.dtype, np.integer):` which if corrected to `if np.issubdtype(image.dtype, np.integer):` does no longer produce these inconsistent results. 

Cheers,
Ruben

tacaswell:
@rfhigler Can you put in a PR making that change?

rfhigler:
@tacaswell Yes I could, but after looking a bit further I don't think my changes actually fixes anything. They only work in my very specific case where I start with uint16s, bandpass them into float64 and rescale them back to either uint8 or uint16. Going the default route (uint16) results in ~3300 local maxima while I only expect 86. The uint8 route works fine; but forcing a rescale of every input to uint8s seems a bit of a hack to me. 

I can get rid of all these spurious maxima hits by increasing the bandpass 'threshold' parameter from its default value of 1 (for integers) to about 2500. I have never had this issue with Maria Kilfoil's matlab/python code, but her code does use a different algorithm to find the local maxima (it doesn't use dilation).

Anyway, as far as I'm concerned this issue can be closed since I can solve it by increasing the threshold value manually. If anyone is interested in reproducing my findings or figuring out exactly what's what I would be more than happy to share my image-data and ipython notebooks.

Cheers,
Ruben

caspervdw:
Hi @rfhigler, thanks for your issue report.

[...]

Issue #141, opened 2014-08-12
Consider changing the default value of filter_before.


danielballan:
The `filter_before=True` parameter of locate is for performance. It eliminates candidate features based on mass (and size if `minsize` is set) to avoid refining the position of features that will be thrown away. When numba is enabled, it does not actually improve performance (as @nkeim recognized) because the computation is faster if performed in the refine step as a numba-optimized loop over the image.

It is possible that `filter_before` could still be helpful in some use cases, especially for images that are not 2D and thus cannot use numba, so it should remain available.

Should it be:
* `False` by default in all cases
* `False` by default if the numba engine is used
* `False` by default according to some more subtle criterion?

PR #400, opened 2016-08-24
Locate refactor: preprocessing and filtering

caspervdw:
This is a start to making feature-finding more modular (#257). 

API changes:
- minmass defaults to 0, not to 100 (integer images) or 1. (float images) (solves obscure behaviour such as in #355)
- filter_before is removed, as it does not increase performance #141.


trackpy/tests/test_feature.py
```diff
@@ -512,7 +512,7 @@ def test_minmass_maxsize(self):
         image = np.zeros(dims, dtype=np.uint8)
         pos1 = np.array([15, 20])
         pos2 = np.array([40, 40])
-        pos3 = np.array([25, 45])
+        pos3 = np.array([25, 50])
```
nkeim: Why did this have to change?

nkeim:
This is a great first step!

Since this is a complicated, multi-step refactor to address #257 in v0.4, do we want to merge now? I think that holding PRs like this one until further decisions have been made will cause suffering. Same goes for #361.

trackpy/tests/test_feature.py
caspervdw: This had to do with the removal of `filter_before`. Without `filter_before` the `maxsize` tests did not pass; probably because the locations of the features change slightly in refinement and therefore the `maxsize` threshold was met by different particles.

This had mainly to do with the fact that features were too close together. Therefore I increased the spacing so that the tests passed again.

caspervdw:
I like the approach more in splitting #257 in multiple PRs. Then we can work on it whenever someone can make the time. So this one can be merged directly, for my part.

Also goes for #361 indeed. Let's look at that soon!

trackpy/feature.py
```diff
@@ -518,48 +533,39 @@ def locate(raw_image, diameter, minmass=None, maxsize=None, separation=None,
...
         warnings.warn("I am interpreting the image as {0}-dimensional. "
                       "If it is actually a {1}-dimensional color image, "
                       "convert it to grayscale first.".format(dim, dim-1))
 
-    if minmass is None:
-        if np.issubdtype(raw_image.dtype, np.integer):
-            minmass = 100
+    if threshold is None:
+        if is_float_image:
+            threshold = 1/256.
```
tacaswell: Should this be `1/255`?

caspervdw: Yes, I guess so.

trackpy/preprocessing.py
```diff
@@ -52,19 +124,52 @@ def bandpass(image, lshort, llong, threshold=None, truncate=4):
...
-            uniform_filter1d(boxcar, 2*smoothing+1, axis, output=boxcar,
-                             mode='nearest', cval=0)
-        if sigma > 0:
-            correlate1d(result, gaussian_kernel(sigma, truncate), axis,
-                        output=result, mode='constant', cval=0.0)
-    result -= boxcar
+    background = boxcar(image, [x * 2 + 1 for x in llong])
+    result = lowpass(image, lshort, truncate)
+    result -= background
     return np.where(result > threshold, result, 0)
```
caspervdw: This probably should be `>=` . When threshold is 1, you would expect that everything below 1 becomes zero, but 1 itself does not.

caspervdw:
I fixed the comment by @tacaswell and by myself (`>=` instead of `>` in thresholding)

trackpy/feature.py
```diff
@@ -486,14 +492,22 @@ def locate(raw_image, diameter, minmass=None, maxsize=None, separation=None,
     .. [1] Crocker, J.C., Grier, D.G. http://dx.doi.org/10.1006/jcis.1996.0217
 
     """
+    if invert:
+        warnings.warn("The invert argument will be deprecated. Use a PIMS "
+                      "pipeline for this.", DeprecationWarning)
+    if filter_before is not None:
```
danielballan: I think this should raise if `filter_before` is True, or at least rephrase the warning. It says the feature "will be deprecated," which I read as meaning "the filtering is still being done now, but it won't be in a future release."

danielballan:
Looking good. I left a quibble about how to deprecate the kwargs, but I fully agree that now is the time to remove those unhelpful "features."

caspervdw:
@danielballan Done, I put a `raise` instead of the warning for the deprecation of `filter_before`. Also I sharpened the other ones (`DeprecationWarning` for `filter_after` and `break_thresh`, `PendingDeprecationWarning` for `invert`)

caspervdw:
If any of you has time, please merge, because I would like to start rebasing #361.

caspervdw:
Thanks @danielballan

PR #406, opened 2016-10-18
REF/API Refactor refine into submodule

caspervdw:
This breaks the `tp.refine()` function: `tp.refine` becomes a submodule containing `tp.refine.center_of_mass()`.

There are no real code changes here, but I would like to know if we are on the same track. Connected to #257 and brings #297 a step  closer.

My aim for the following PR is to provide a `tp.refine.center_of_mass` that takes and produces a DataFrame, for user-friendlyness, and a `tp.refine._center_of_mass` that takes and produces a numpy array, for performance and internal use in the preconfigured tracking algorithms (such as `tp.locate`).


danielballan:
Yep, I like this pattern. Merging to keep things moving along; I hope I'm not cutting discussion short by doing so.

PR #409, opened 2016-11-25
in Compute_drift, sort traj to get it robust

thierryBottaro:
compute_drift does NOT work if the trajectory Dataframe is not sorted by frame.
the line `delta = traj.groupby('particle').apply(lambda x :x.set_index('frame', drop=False).diff())` expect that the frames are sorted. if not, the "delta.frame" will be relatively random (could even be negative), and the next line "delta = delta[delta['frame'] == 1]" will discard lots of (and sometimes all) lines.
So here is my correction proposition.

I'm not sure if it is preferred to make a copy or a "inplace" sort. 
"trajc.sort_values('frame',inplace=True)"  would also work, but then the trajectory Dataframe passed by the user will be changed, which is perhaps unwanted by the majority

### DIFF
--- a/trackpy/motion.py
+++ b/trackpy/motion.py
@@ -258,7 +258,7 @@ def compute_drift(traj, smoothing=0, pos_columns=None):
         pos_columns = ['x', 'y']
     # the groupby...diff works only if the trajectory Dataframe is sorted along frame
     # I do here a copy because a "inplace=True" would sort the original "traj" which is perhaps unwanted/unexpected
-    traj = traj.sort_values('frame')
+    traj = pandas_sort(traj,'frame')
     # Probe by particle, take the difference between frames.
     delta = traj.groupby('particle').apply(lambda x :
                                     x.set_index('frame', drop=False).diff())
--- a/trackpy/motion.py
+++ b/trackpy/motion.py
@@ -258,9 +258,9 @@ def compute_drift(traj, smoothing=0, pos_columns=None):
         pos_columns = ['x', 'y']
     # the groupby...diff works only if the trajectory Dataframe is sorted along frame
     # I do here a copy because a "inplace=True" would sort the original "traj" which is perhaps unwanted/unexpected
-    traj = pandas_sort(traj,'frame')
+    traj = pandas_sort(traj, 'frame')
     # Probe by particle, take the difference between frames.
-    delta = traj.groupby('particle',sort=False).apply(lambda x :
+    delta = traj.groupby('particle', sort=False).apply(lambda x :
                                     x.set_index('frame', drop=False).diff())
     # Keep only deltas between frames that are consecutive.
     delta = delta[delta['frame'] == 1]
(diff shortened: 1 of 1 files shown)


caspervdw:
Hi @thierryBottaro , thanks for the contribution to trackpy. I think this approach is right. Only thing is that the tests fail because `df.sort_values` is not backwards compatible with all tested pandas versions. To fix this, we always use the `trackpy.utils.pandas_sort` for sorting, which is already imported in `motion.py`. Could you update the code making use of this function?

thierryBottaro:
Thank you Casper, your advice pointed precisely the solution!
now that "all checks have passed", does it means that my modification will be taken into the next Trackpy Version ?
thank you anyway:)

caspervdw:
Thanks for the update. Just a further thought: as `groupby` will now sort again on `'particle'` (preserving the `'frame'` sorting) it might be faster to do it in one go with `pandas_sort(traj, ['particle','frame']` and pass `sort=False` to `groupby`.

thierryBottaro:
Nope !
bad luck! I thought it was a good idea (to be honest, I did not know that "groupby" sort per default by the "group" key, because in SQL, it is not the case), but as good practice, I needed to test it.
here you are :
[compute_drift_modif.zip](https://github.com/soft-matter/trackpy/files/617371/compute_drift_modif.zip)

the current modification is faster than the one you propose.
But still, you hint a interesting point : groupby is sorting the result by particle... but who need that ?
we anyway take the mean along particle, so we don't care about the order, so I added the "sort=False" in groupby anyway.
and as expected, it is faster. (compute_drift_really_fast in my test file)

thierryBottaro:
Oh, and I corrected the description which was wrong about the default smoothing.

caspervdw:
Thanks for checking this out @thierryBottaro . Before merging, a few PEP8 code style things which I left in line comments

trackpy/motion.py
```diff
@@ -248,16 +248,19 @@ def compute_drift(traj, smoothing=0, pos_columns=None):
...
+    >>> compute_drift(traj, 0, ['x', 'y']).plot() # not smoothed, equivalent to default.
     >>> compute_drift(traj, 15).plot() # Try various smoothing values.
     >>> drift = compute_drift(traj, 15) # Save good drift curves.
     >>> corrected_traj = subtract_drift(traj, drift) # Apply them.
     """
     if pos_columns is None:
         pos_columns = ['x', 'y']
+    # the groupby...diff works only if the trajectory Dataframe is sorted along frame
+    # I do here a copy because a "inplace=True" would sort the original "traj" which is perhaps unwanted/unexpected
+    traj = pandas_sort(traj,'frame')
```
caspervdw: Please insert a space after the comma

danielballan:
Looks good to me. Thanks for the contribution, @thierryBottaro! Yes, this will be in the next release of trackpy.

PR #407, opened 2016-10-20
Refine leastsq

caspervdw:
This PR builds onto #406 . I implemented all the code for:
- combined linking & feature finding
- separating images in clusters of features
- least squares refinement of overlapping features, using any radial model function
- adapted `artificial.py` and coupled tests in `test_feature` and `test_find` to work with a better convention for artificial feature creation

The description of the algorithm is available online in https://arxiv.org/abs/1607.08819. A revised version is currently ``provisionally accepted'' by a peer-reviewed journal. Will keep you updated about how that is going.

I am referring to trackpy in the paper in the sense that it ''will be implemented in trackpy, but currently available on https://github.com/caspervdw/clustertracking''. It would be neater to change that to ''will be  available from trackpy v0.4'', as I think it might pull some more users to trackpy. Is this OK with you guys? I will probably have to supply the definitive version of the paper in the next week or so. It's not that this needs to be merged by then, just the intention of putting it inside v0.4 is enough.

PR #410, opened 2016-12-21
FIX Ignore features that are in the margins in `find_link`

caspervdw:
This fixes a rare exception that is raised when a feature that is outside the image margins is characterized.

Also rebased the `tp.feature.characterize` on `tp.refine.center_of_mass._refine`

This only touches code that was merged in #407, so I am inclined to self-merge. Please stop me if you disagree!

### DIFF
--- a/trackpy/feature.py
+++ b/trackpy/feature.py
@@ -10,11 +10,12 @@ from pandas import DataFrame
 
 from .preprocessing import (bandpass, convert_to_int, invert_image,
                             scalefactor_to_gamut)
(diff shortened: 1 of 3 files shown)


nkeim:
Merging. You make it sound so trivial, but it looks like it was actually a pretty difficult fix.

On a related note: `test_feature.py` already had both `test_maxima_in_margin` and `test_reject_peaks_near_edge`, which check this case comprehensively using `locate()`. It seems like it would be a good idea to port most of those tests to `find_link`. Do you think that would be easy?

Issue #292, opened 2015-10-09
Linking ignores empty frames


caspervdw:
I just came across a peculiar issue: if I have a single blinking particle and link this with `memory=0`, this still results in a complete track. I would expect the tracks to be separated at frames where the particle is not found.

Because `link_df` iterates over all unique framenumbers, it ignores empty frames. I think this is not the desired behaviour, but correct me if I'm wrong.

danielballan:
Agreed. That is not the correct behavior.

tacaswell:
That does not sound like the desired behavior at all.

nkeim:
trackpy has always been fairly agnostic about frame numbering. I'm pretty sure that except for the storage classes, frame numbers are never required to be integers, or evenly spaced. I know specifically that the prediction framework is written to accommodate arbitrary sequential real numbers, and that these are used in the unit tests. This is because prediction requires not just the order of frames, but their times as well.

Secondly, skipping frames is sometimes desirable. Momentary vibrations or camera glitches can make certain frames unusable. So sometimes the behavior Casper describes *is* desired, and the "correct" behavior may be a surprise to the user.

So we should tread carefully here. Whatever we do to `link_df()`, I think that `link_df_iter()` should not assume successive integers.

tacaswell:
Maybe we should crib the ndimage labeled image api and add an optional `expected_frames` kwarg to `link_df_iter` which lists the expected frame numbering.  If it is `None`, it just runs through them as-is, and if it is given will only link on those frames, both inserting missing frames and skip existing frames as needed.

caspervdw:
We should require `memory` to have the same units as the `frame` column, which narrows down its dtypes to integer and floats. Also most of the functions in `tp.motion` expect numbers (integers). For a momentary vibration or camera glitch, time does go on and linking trajectories through such a glitch (with `memory=0`) should still result in broken tracks, in my opinion. That's what the memory function is for right?

Consider one blinking particle and one stable particle, both isolated. If you track this (no memory), you will have broken tracks for the blinking particle. If you now remove the stable particle, you will suddenly have a single track for your blinking particle. This makes linking (in some cases) dependent on other particles that are arbitarily far away. Which is not good, it should be an intrinsic property of the isolated particle.

So I would argue for a default behaviour of `expected_frames` that checks for the minimum difference between `frame` values and inserts empty frames if one frame is missing.

nkeim:
[...]

PR #414, opened 2017-01-01
Add prediction to find_link

nkeim:
This makes prediction work with find_link. In the process I added the equivalent of `link_df_iter()`.

However, testing is incomplete, because my prediction tests use floating-point frame numbers. (Prediction needs to know about the time of each frame, not just sequential ordering.) This was discussed in #292, and at the time we decided that this was a potentially useful feature that we basically got for free. The status quo is that `link_df_iter` (alone) honors arbitrary frame numbers.

@caspervdw, do you foresee any major complications that would come from adding awareness of frame numbers to your `find_link` code? I may be able to do it myself.

Also needed are some basic tests of prediction with the full `find_link` pipeline, including feature identification. That will greatly enhance the overall value of prediction.


caspervdw:
Nice @nkeim! 
I think the new implementation now almost has the full functionality of the older.

About floating-point framenumbers: in #292 we concluded that framenumbers should be of integer type, as they refer to a discrete sequence of frames. This was also the rationale behind the new implementation of `Linker`: on each new frame, `next_level` is called.
`FindLinker` already is framenumber-aware, as it interprets `frame.frame_no`. The framenumbers are stored inside the `hash` objects (which are `TreeFinder` objects).

That said, the rest of `Linker` is (as far as I can see) dtype-agnostic. So if it suits you, I don't see any problem in leaving `frame` open to be of any type. The only thing is that `link_df` (and friends) have to be able to group and sort the dataframe correctly.

For the tests, I think it would make sense to adopt `frame` and `particle` columns that are of integer type. And (optionally) separately test with floating-point `frame` columns.

caspervdw:
@nkeim I have some time for coding today/tomorrow, so I could have a look at this framenumber-awareness and make a PR to your repo. Please let me know if that's OK so that I don't cross your work.

nkeim:
@caspervdw that would be great! Thanks.

caspervdw:
See https://github.com/nkeim/trackpy/pull/5

caspervdw:
With debugging I convinced myself that at [this line](https://github.com/soft-matter/trackpy/pull/414/files#diff-ef01253517b496f9c55faa898b68174bR693), `t` is indeed the (floating point) time of the current frame, which is the same as in the original implementation ([link](https://github.com/soft-matter/trackpy/blob/master/trackpy/linking.py#L1090)).

So that problem is solved now, but still the tests are failing! I did some debugging but I can't find the error. Do you have an idea @nkeim?

Three more thoughts
 - I think that a good way to go would be `**kwargs` for every linker function, which are passed through to the `Linker` initialization. So the call signature of `link` would be `link(f, search_range, **kwargs)`. Possibly, `linker_class=Linker` could be a default value.
 - For prediction, coordinates are repeatedly converted between `ndarray` and list of `Point`. Conversion goes through list comprehension and for larger amounts of particles this is way slower than ndarrays. It would increase peformance if the predictors would (optionally?) take an `ndarray` which is already accessible through the `hash.coords`.
 - The way prediction is implemented now feels a bit "hackish" to me now: basically the `prev_hash` is changed inplace so that everything works out later on. From a user perspective, `prev_hash.coords` and `prev_hash.coords_rescaled` now unexpectedly return *predicted* coordinates. It might be more clear to use the `predictor` at the point where it is needed, namely in the `Subnet`-generating code at [this line](https://github.com/soft-matter/trackpy/blob/master/trackpy/find_link.py#L405), and afterwards directly put `hash._clean = False`.

trackpy/find_link.py
```diff
@@ -25,60 +25,85 @@
...
-    linker.init_level(coords, t=0)
-    yield 0, linker.particle_ids
+    linker = Linker(search_range, memory, predictor=predictor)
+    linker.init_level(coords, t)
+    yield t, linker.particle_ids
 
-    for t, coords in enumerate(coords_iter, start=1):
+    for t, coords in coords_iter:
         linker.next_level(coords, t)
         for k, coord in enumerate(linker.coords):
```
caspervdw: These two lines are obsolete (I recall to test whether the coords had been changed shuffled), they should be removed for better performance

nkeim:
I found the problem: at [this line](https://github.com/soft-matter/trackpy/pull/414/files#diff-ef01253517b496f9c55faa898b68174bR526), the `Subnets.add_dest_points` method builds new KDTrees out of lists of `PointND` objects, rather than using the (predicted) coordinates stored in `prev_hash`.

@caspervdw how does this code fit into the new algorithm? Is there an equivalent implementation (without constructing KDTrees) in the old algorithm, or is this new functionality?

nkeim:
Also, you are right that the predictor implementation was definitely a hack, although to put a more positive spin on it, I wanted to make as few intrusions into `linking.py` as possible. I like your proposed solution, except that as evidenced by my previous comment, I think that we are still not clear on where the "true" particle positions should be stored.

caspervdw:
This `Subnets.add_dest_points` is actually only used in combination with this `relocate`, that adds extra features after construction of the subnets. So this can't explain the failed tests that do not make use of the `find_link` (which has relocate) but only of the `link_simple` (which is supposed to just reimplement `link_df` in the long run)

Now we are reimplementing this code anyway, a little refactor might be in place, the error might also turn up in that way.

caspervdw:
@nkeim, I think I found it! I used [y, x] convention for the dimension order in the position fields, while prediction uses [x, y]. That explains why all tests except for the `BaselinePredictTests` failed.

I would like keeping [y, x] as we use it for most of the trackpy code. We could stitch it together by swapping the columns before and after prediction. Or propagate the convention into the predict API.
What do you think?

About the 'hackish' implementation, I guess it is just another word for a clever implementation ;) What bothered me is the naming of the things (as I described), I think if we rename `TreeFinder.coords_rescaled` to `TreeFinder.coords_mapped` and put the prediction in there, it would be fine already.

I will make a PR for that and a quick fix for the dimension swapping. Hopefully we will have a happy Travis then.

caspervdw:
It seems I had push rights on your repo. I have added some commits and rebased on master.

trackpy/find_link.py
```diff
@@ -424,6 +422,117 @@ def __init__(self, t, pos, id=None, extra_data=None):
...
+            self._kdtree = cKDTree(self.coords_mapped, 15)
+        # This could be tuned
+        self._clean = True
+
+    @property
+    def coords(self):
+        return _points_to_arr(self.points)
+
+    @property
+    def coords_mapped(self):
```
nkeim: It looks like `coords_mapped` should be used only once per frame. Is that true? Depending on the predictor this can be an expensive operation.

nkeim:
Fantastic! Yes, your push rights are a new GitHub feature; they are for this branch only.

I think this is a very sensible refactor of how prediction works! Assuming we are going to make `find_link` the default linking algorithm, at that time we should swap all the axes in `predict.py`. That will also break any existing user-provided predictors.

See my comment about performance, which is not an issue at this time. So I think this is ready. Who is going to merge?

trackpy/find_link.py
caspervdw: Good point. Maybe better to put the call to `coord_mapping` (the predictor) directly in the computation-heavy `rebuild` function, and call `rebuild` from `coords_mapped` if `_clean == False`?

Then the `KDTree` is built the first time fi `coord_mapping` is accessed. It does need to be built at some point so I think that is fine.

caspervdw:
I tackled the linenote you placed. Also cleaned some unused `TreeFinder` code and changed the `predictor` kwarg to `**kwargs`, that are passed on to the `Linker`.

The prediction hasn't actually been implemented for `find_link`. That is a next step, let's do that too in this PR. I'll have a look as I probably know my way around the `find_link` code best.

caspervdw:
I think I fixed the implementation of prediction in `find_link`, although I can't get the prediction-findlink tests to work. But that is mostly because I don't fully get the workings of the tests in `test_predict`.

In some way, the linkers for find_link need to be wrapped so that `reader` can be supplied to `find_link`, instead of `features`. I put down some code that is pretty close to working. Maybe you could have a look how to reimplement `predictor.wrap` in combination with `find_link` @nkeim?

caspervdw:
The problem is that `find_link` does not work with non-integer framenumbers. The solution to that is not obvious. Is it possible for you @nkeim to make the prediction tests with integer-framenumbers?

caspervdw:
@nkeim What shall we do with this?

My opinion on this is to limit `frame` column to integer and allow an optional `fps` parameter into the prediction framework.

nkeim:
Agreed. I think that an `fps` parameter would get used very rarely, so I think that just switching to integers-only is appropriate. It may be a couple more weeks, however, before I can spend the time to make the tests integer-only. Sorry for the delay!

caspervdw:
OK! In the meantime I might take a look at the `adaptive_search` on this branch. That also needs to be reimplemented.

caspervdw:
No the tests are the same, I only needed to increase the particle spacing (and search range correspondingly) to be able to draw the features on an artificial image and feed that to `find_link`.

nkeim:
Awesome!

I'd like to hear from @danielballan or @tacaswell before we merge, just because we are messing with a core part of trackpy (the pure-Python subnet linkers), and this might trigger a specific memory for them.

nkeim:
Ack! Sorry, I got this mixed up with #430. That is already merged, but @tacaswell or @danielballan let us know if you have anything to add to that discussion.

nkeim:
I think I've fixed the remaining issues. The largest ones were:

- Non-integer frame numbers were part of the problem. That is fixed.
- For tests that supply an initial velocity guess, coordinates can't be rescaled within the linker (which was happening for the `find_link` tests.
- I chose to skip a few minor tests for `find_link` (in which coordinates are passed to the linker via an image). They were mostly redundant and would have required significant additional code to support.

nkeim:
(It looks like the remaining failing `find_link` tests are not related to prediction.)

trackpy/tests/test_predict.py
```diff
@@ -212,13 +248,159 @@ def setUp(self):
...
+    pass
+
+
+class FindLinkPredictTest(object):
+    def setUp(self):
+        super(FindLinkPredictTest, self).setUp()
+        self.linker_opts = dict(separation=10, diameter=15)
+        # Disable certain tests that are redundant here
+        # and would require more code to support.
+        self.coords_via_images = True
```
nkeim: This approach would allow you to skip the currently-failing `test_quadrature_distances` test for find_link

caspervdw:
Great that it is all passing now! Tthanks for the work @nkeim

About the `quadrature_tests`, I looked into this test and it seems to be not OK anyway. It should test for that distances are added in quadrature, by looking at 2 pairs of features that have two degenerate linking solutions. Then, by adjusting one coordinate slightly, it would give one or the other solution. However, the coordinates tested are degenerate for both quadrature and linear addition of distance.

I will see if I can come up with a better set of degenerate coordinate pairs.

caspervdw:
The new normal linker as well as the FindLinker pass with prediction!

The API differences between the `linking.link_df` and `find_link.link_simple` are now easy to oversee:

 - `linking.link` has no equivalent function in the new implementation, as it works on lists of `PointND` objects, which are non-user friendly 'internal' datatypes
 - `linking.link_iter` also
 - `find_link.link_simple_iter` takes numpy arrays instead of lists `PointND` objects
 - `find_link.link_simple` is equivalent to `linking.link_df`
 - `find_link.link_simple_df_iter` is equivalent to `linking.link_df_iter`
 - `neighbour_strategy`: the non-default BTree has not been implemented (yet) for the new linkers
 - `diagnostics`: not implemented
 - `verify_integrity`: migrated to separate function that checks a dataframe
 - `retain_index`, `copy_features`:  indices are always retained, dataframes are never copied

I am :+1: on making the new linker the default, and keeping the old code for a while as a legacy option.

nkeim:
I am confused about the revised quadrature test. The original test was introduced in #212, where I found that it correctly detected subnet distances that were not added in quadrature. The new test appears to be checking something different.

I reached this conclusion by changing line 1419 of `linking.py`

```py
        tmp_sum = cur_sum + cur_dist**2
```

removing the `**2` so that the non-recursive subnet solver adds distances linearly. With this change, 3 tests fail: 
- `test_penalty` in `test_link.py`
- the *original* `test_quadrature_distances` in `test_link.py`
- `test_penalty` in `test_link_new.py`

`test_quadrature_distances` in `test_link_new.py` (the subject of this PR) does *not* fail. So it is checking for something different that I don't yet understand.

Since `test_penalty` seems to also detect this bug, perhaps `test_quadrature_distances` is redundant anyhow?

caspervdw:
I think I misunderstood this test. With "not adding in quadrature" I thought it meant `x0 + y0 + x1 + y1` as apposed to `x0**2 + y0**2 + x1**2 + y1**2`, with `x0` the displacements of particle pair `0` in `x` direction etc.

The test from #212 was not able to discern these two cases (this is what I wrote previously). But apparently it wasn't supposed to, it was supposed to catch `sqrt(x0**2 + y0**2) + sqrt(x1**2 + y1**2)`. This is what I misunderstood.

The new test is not able to catch `sqrt(x0**2 + y0**2) + sqrt(x1**2 + y1**2)` but is able to catch `x0 + y0 + x1 + y1`. There is some coincidence here!

That `test_penalty` catches this is also a coincidence, but a fortunate one, we could add the description that it also tests for quadrature distances and then indeed drop `test_quadrature_distances`

nkeim:
Ah! That makes perfect sense.

Upon further reflection, I think it's best if we rename the original `test_quadrature_distances` to `test_quadrature_sum`, and then keep both the old and new tests. It is good to have tests for very specific bugs, as well as more general catch-alls. I've just pushed this change.

caspervdw:
That looks good. I like the `as in Pythagoras`. 

I was surprised that the `FindLinkTests` passed with the new `test_quadrature_distances`. An `epsilon` of `0.01` should not make a difference in the `FindLinker`, as it basically works with integer coordinates. So I checked the tests and apparently I mixed up the `np.allclose` with `np.testing.assert_equal`. Both tests fail now so we will need to revisit them..  For sure the coordinates of `test_quadrature_distances` need to be further apart.

caspervdw:
This should fix the build! I basically revised the `DataFrame` testing bit (which is always a pain) using the previously implemented `assert_traj_equal`. In this way, tests don't fail if the `FindLinker` returns the features in a different order.

Also, I pulled the coordinates of the `test_quadrature_sum` test farther apart. It now fails if you remove the `**2` at line `1419` in `linking.py`.

nkeim:
Excellent!! I propose that we merge this PR, and then start a new branch/PR to make this new linker the default, and make any required documentation updates, etc.

That branch would be appropriate for performance testing using `asv`, which I could probably take care of.

caspervdw:
I agree with proceeding as you propose! Do @danielballan @tacaswell  want to review this code before we merge?

danielballan:
I'd be glad to be another pair of eyes on this before it's merged. Give me a couple days to find a block of time.

And, @nkeim, I'd be excited to see asv finally running for trackpy!

caspervdw:
Ok sure take your time. Great if you can have a look too.

Just to avoid confusion: `find_link.py` now contains a sort of complete implementation of the linking functionality (`link_simple` as temporary name) as well as the `find_link` extension.

danielballan:
LGTM. What's the status of the commented-out tests like `test_easy_tracking`?

caspervdw:
Right, that is something to be looked at still. Tests that feed `PointND` objects into `link` have to be rewritten into tests that feed iterables of ndarrays or dataframes, as we dropped the `PointND` API. `link` and `link_iter` don't have `link_simple` equivalents in terms of data format of the coordinates.

caspervdw:
@danielballan I fixed the last tests, the only one I removed was the `test_box_size`, which is `BTree` specific and I did not reimplement the `BTree`.

caspervdw:
@danielballan @nkeim  I'll merge after you approve.

danielballan:
:+1:

nkeim:
:+1: !

Looking ahead to the switchover, we need a way to compare directly the output of the old and new linkers. I think that this is an excellent opportunity to add the regression tests alluded to at the end of the discussion in #361, but whatever we do, I would be a lot more comfortable making the change if we could show identical output.

PR #416, opened 2017-01-19
FIX find_link with refine=True

caspervdw:


caspervdw:
I fixed a couple of bugs inside the `find_link` routine. This is good to go, as it only touches the new `find_link` code, I will wait some days and self-merge.

nkeim:
Great! Does this affect the dtype issues that you addressed in #414? It looks like it does not, and just checks that `find_link()` honors mixed dtypes.

caspervdw:
It also coerces the output dtypes of `frame` and `particle` to int. Basically, it initializes the returned data frame with integer typed empty columns. I dont think this will lead to issues for non-int types frame identifiers.

caspervdw:
@nkeim I have extended the tests a bit. For the normal linking, it asserts that the `frame`, `x` and `y` dtypes are preserved (float or int) and the returned `particle` dtype is always integer. For the find-linking, the dtypes are more scrict, as features are added inside the routine.

PR #421, opened 2017-02-07
Replace deprecated `getmoduleinfo` by `getmodulename`

lschr:
`getmoduleinfo` has been deprecated since Python 3.3 and has been
removed in Python 3.6. Replace it by `getmodulename`.


caspervdw:
Thanks @lschr

Issue #423, opened 2017-02-16
DOC Automatic build of notebooks fails


caspervdw:
Something is wrong with the automatic notebook build. The dev docs contain empty notebooks (for example: http://soft-matter.github.io/trackpy/dev/tutorial/walkthrough.html)

Travis says during the docbuild:

    ImportError: No module named 'jupyter_client'

PR #425, opened 2017-02-17
DOC Build pin all versions of nbconvert

caspervdw:
Addresses #423  by pinning all versions that have to do with the doc build. Saves us also from further surprises.


nkeim:
Smart. Thanks!

PR #428, opened 2017-02-23
find_link was not properly doing the com refinement when refine=True

pfigliozzi:
Even after the PR from #416 the refinement was not done properly and after checking with tp.subpx_bias() the particle positions were confined to integer values in x and y. Now, the refinement is actually done and produces reasonable locations confirmed by tp.subpx_bias()


caspervdw:
@pfigliozzi  Great to learn that someone is actually using my `find_link` code! If you have any questions or suggestions, please let me know. This code really is in an experimental phase still.

I completely agree with the change you made, the image should be preprocessed before the `refine_com`.

PR #430, opened 2017-03-13
BUG Penalty for breaking links

caspervdw:
While implementing different linking strategies for the `FindLinker` in #414, I found this bug.

Apparently, only the numba subnetlinker handles the penalty for breaking links correctly. Both the recursive and nonrecursive strategies fail in this testcase. This testcase describes the following two source particles (o) and destination particles (x) with respective coordinates (0, 0), (8, 0), (7, 8), (15, 8).

```
o    o

    x    x
```

We have two likely linking results:
1. two links, total squared displacement = `2*(8**2 + 7**2) = 226`
2. one link, total squared displacement = `(8**2 + 1**2) + search_range**2`

Case 2 gets a penalty for not linking, which equals the search range squared. Depending on the value of search_range, you expect different results:

- With a penalty of 13, case 2 has a total cost of 234 and we expect case 1. as the result.
- With a penalty of 12, case 2. will have a total cost of 209 and we expect case 2. as the result.

I am not sure what is the desired behaviour of the nonrecursive linker, but the recursive linker should for sure reproduce the result, and it does not: it always produces case 1.

### DIFF
--- a/trackpy/linking.py
+++ b/trackpy/linking.py
@@ -1402,7 +1402,6 @@ class SubnetLinker(object):
 
 
 def nonrecursive_link(source_list, dest_size, search_range, max_size=30, diag=False):
-    #    print 'non-recursive', len(source_list), dest_size
     source_list = list(source_list)
     source_list.sort(key=lambda x: len(x.forward_cands))
     MAX = len(source_list)
@@ -1550,7 +1549,7 @@ def numba_link(s_sn, dest_size, search_range, max_size=30, diag=False):
 @try_numba_autojit(nopython=True)
 def _numba_subnet_norecur(ncands, candsarray, dists2array, cur_assignments,
                           cur_sums, tmp_assignments, best_assignments):
-    """Find the optimal track assigments for a subnetwork, without recursion.
+    """Find the optimal track assignments for a subnetwork, without recursion.
 
     This is for nj source particles. All arguments are arrays with nj rows.
 
--- a/trackpy/linking.py
+++ b/trackpy/linking.py
@@ -1513,8 +1513,8 @@ def numba_link(s_sn, dest_size, search_range, max_size=30, diag=False):
     # A source particle's actual candidates only take up the start of
     # each row of the array. All other elements represent the null link option
     # (i.e. particle lost)
-    candsarray = np.ones((nj, max_candidates + 1), dtype=np.int64) * -1
-    distsarray = np.ones((nj, max_candidates + 1), dtype=np.float64) * search_range
+    candsarray = np.ones((nj, max_candidates), dtype=np.int64) * -1
+    distsarray = np.ones((nj, max_candidates), dtype=np.float64) * search_range
     ncands = np.zeros((nj,), dtype=np.int64)
     for j, sp in enumerate(src_net):
         ncands[j] = len(sp.forward_cands)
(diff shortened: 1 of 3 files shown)


nkeim:
Wow! I based the numba linker on the nonrecursive linker, so that is really remarkable. All 3 are supposed to give the same result.

I may be able to look into this next week and have made a note to do so.


On Mar 13, 2017, 7:44 AM -0700, Casper van der Wel <notifications@github.com>, wrote:

While implementing different linking strategies for the FindLinker in #414<https://github.com/soft-matter/trackpy/pull/414>, I found this bug.

Apparently, only the numba subnetlinker handles the penalty for breaking links correctly. Both the recursive and nonrecursive strategies fail in this testcase. This testcase describes the following two source particles (o) and destination particles (x) with respective coordinates (0, 0), (8, 0), (7, 8), (15, 8).


o    o

    x    x



We have two likely linking results:

  1.  two links, total squared displacement = 2*(8**2 + 7**2) = 226
  2.  one link, total squared displacement = (8**2 + 1**2) + search_range**2

Case 2 gets a penalty for not linking, which equals the search range squared. Depending on the value of search_range, you expect different results:

  *   With a penalty of 13, case 2 has a total cost of 234 and we expect case 1. as the result.
  *   With a penalty of 12, case 2. will have a total cost of 209 and we expect case 2. as the result.

I am not sure what is the desired behaviour of the nonrecursive linker, but the recursive linker should for sure reproduce the result, and it does not.

________________________________
You can view, comment on, or merge this pull request online at:

  https://github.com/soft-matter/trackpy/pull/430

Commit Summary

  *   BUG Tests for subnet linking penalty

File Changes

  *   M trackpy/tests/test_link.py<https://github.com/soft-matter/trackpy/pull/430/files#diff-0> (40)

Patch Links:

  *   https://github.com/soft-matter/trackpy/pull/430.patch
  *   https://github.com/soft-matter/trackpy/pull/430.diff

caspervdw:
@nkeim Would be great if you dive into the subnetlinkers, that part is really a black box to me.

I confirmed that the tests fail in a similar way for v0.3.2, so it does not have anything to do with #407. In #407, did rescale the distances with `search_range` (for anisotropic search_range), so the subnet_linker always sees a search_range of 1 and distances `<=1`. But the story stays the same.

nkeim:
The penalty was being double-counted: once when the "null particle" (i.e. no link, distance `search_range`) was appended to the source particle's `forward_cands` list, and once by the subnet algorithm itself. Apparently the null forward candidate has been with trackpy for a few years. This only impacted trackpy users who do not use the numba linker (few if any), and only in this edge case, so it makes sense that it was not found until now.

Fixing this problem (with the non-numba linkers) created is causing the `find_link` test `test_shifting_string` to fail. I really don't know why at this point—maybe you could take a look, @caspervdw ?

This realization also led me to look at the numba code more closely. The fact that it did not have this bug was entirely unintentional on my part. There, the penalty was not counted twice, but it did take up an extra column in the distances array. I have removed it after careful testing, so that the code will be clearer to future readers.

Thanks to @caspervdw for finding this!

caspervdw:
Great that you solved the issue @nkeim.
The failing find_link test is actually the reason why I found this bug, because it expects the wrong outcome (I should have told you). I reduced the shift so that the shifting string is linked correctly and the test pass.

I'll merge as soon as Travis passes.

caspervdw:
Oh sorry about the fast merge @nkeim

nkeim:
No problem @caspervdw ! It's just that anytime we make a change to the "standard" linking code it makes me sweat a little.

PR #431, opened 2017-03-16
More verbose error messages in RefineExceptions

caspervdw:


caspervdw:
I'll self-merge this, it only gives some descriptions of RefineExceptions and fixes the `IndexError` when logging the description of the error. Travis has a HTTP issue on one build, which is typically intermittent

PR #436, opened 2017-04-03
BLD Append conda-forge channel, instead of adding

caspervdw:
So that it will not have a higher priority than the default channel.


caspervdw:
I have added Py36 tests. There is this warning appearing:

```
/home/travis/mc/envs/testenv/lib/python3.6/importlib/_bootstrap.py:205: ImportWarning: can't resolve package from __spec__ or __package__, falling back on __name__ and __path__
  return f(*args, **kwds
```

@sciunto  Do you have an idea what is causing this?

PR #442, opened 2017-06-20
REF: Refactor linking into submodule

caspervdw:
This is my first shot at refactoring the linking routines. Continues the work in #414. I put all the legacy linking code in `trackpy.linking.legacy`, and the corresponding tests in `test_legacy_linking`.

Basically, we are left with three linking functions:

 - `trackpy.linking.legacy.link` and `trackpy.linking.legacy.link_iter` have no equivalent function in the new implementation, as they works on lists of `PointND` objects, which are non-user friendly 'internal' datatypes
 - `trackpy.link_iter` takes iterables of numpy arrays instead of iterables of `PointND` objects
 - `trackpy.link` is (alias for `trackpy.link_df`) is equivalent to `trackpy.linking.legacy.link_df`
 - `trackpy.link_df_iter` is equivalent to `trackpy.linking.legacy.link_df_iter`

We might want to sharpen this API a bit: maybe `trackpy.link` should take and return numpy arrays?

And the following functionality is not reimplemented (yet?):
 - `neighbour_strategy`: the non-default BTree has not been implemented (yet) for the new linkers
 - `diagnostics`: not implemented
 - `verify_integrity`: migrated to separate function that checks a dataframe
 - `retain_index`, `copy_features`:  indices are always retained, dataframes are never copie


apiszcz:
It appears we'll need to modify existing code to use the .legacy API call?

For the linker API I'm wondering if a modern 'status' display of linking
runtime progress might be supported with tqdm or similar?
https://pypi.python.org/pypi/tqdm.

caspervdw:
@apiszcz The new implementation of `link_df` and `link_df_iter` gives the same results as the old, as confirmed by many unittests. However, there are some kwargs (listed in my first message) that are not present anymore in the new implementation.

If these are required, indeed changing the import statement will work:

    from trackpy.linking.legacy import link_df

@nkeim There is a strange error popping up in only one of the Travis builds, originating from the numba subnetlinker. Could you have a look? Could it possibily have to do with #414? I remember you changed all subnetlinkers except for the numba linker there.

caspervdw:
@nkeim, please correct me if I'm wrong, I have pushed a change in the numba subnetlinker at line 260

nkeim:
@caspervdw I think you are referring to #430 ? That included changes to the numba linker.

I have spent 20 minutes looking at your change and it makes sense. However apparently it may be introducing new problems, and apprently we have found an issue with dependencies. I am going to look at this some more.

If we are really confused we can just roll back some of the changes in #430, but obviously I would like to better understand what is going on.

nkeim:
I have found the problem. The change to the numba linker in #430 removed an (ostensibly) superfluous null candidate. But it turns out that the numba linker was relying on its presence. The root causes were:

1. The numba linker uses -1 to signify the null link (and distinguish it from an actual particle that can be used only once). This is different from the non-numba linkers, which signify a null link with `None` instead of a Point object.
2. The "legacy" Travis environment runs the numba linker without compiling it (i.e. in pure-Python mode). Apprently this enables bounds checking on arrays 😨. So @caspervdw your fix was completely appropriate and necessary!

I am submitting a PR against your branch.

nkeim:
Actually, I can't seem to submit PRs against your branch, @caspervdw. I put my changes on https://github.com/nkeim/trackpy/tree/refactor-linking

tacaswell:
The `Point` version of the functions let you do linking with wonky metrics (ex https://github.com/tacaswell/leidenfrost/blob/master/leidenfrost/infra.py#L109 ).  The refactored API should preserve some way of injecting either custom metric or a custom `Point` class.

caspervdw:
That would be a real pain to reimplement. Actually, I have done this linking myself on curved surfaces, and at some point I decided to convert the (2D) internal coordinates on the metric to (3D) eucledian coordinates just for the linking. In this way, the (fast) kdtree could still be used.

So the use case on a 1D circular metric would be:

    f['x'], f['y'] = np.cos(f['angle']), np.sin(f['angle'])
    f = tp.link(f, 1.)


This does however not implement custom distance functions. This will require a significant amount of work, but I think it can be done using the same approach as I just described, because eucledian distance is always the shortest.

caspervdw:
@nkeim Great that you solved the issue! Apparently you had push rights to my branch because the commits are there and the tests passed.

caspervdw:
@tacaswell  For reference, sklearn supports KDTrees with custom metrics:

* http://scikit-learn.org/stable/modules/generated/sklearn.neighbors.KDTree.html
* http://scikit-learn.org/stable/modules/generated/sklearn.neighbors.DistanceMetric.html

caspervdw:
@tacaswell We need to decide what to do. Are you OK with merging this as is and leaving the custom distance functions as an open issue? One could always use the legacy functions to access this feature in the meantime.

caspervdw:
@nkeim I added the reproducibility unittests that check the outcome of feature finding and linking versus v0.3.1. This will give us a continuous confirmation of the end results.

As shown by you in https://github.com/soft-matter/trackpy/pull/361, the feature finding now gives slightly different results, originating from the different gray dilation kernel. 77 of 1881 features found by v0.3.1 were not found in the current master, and vice versa, 72 of 1861 features in master were not found by v0.3.1. Because of the image quality, I think this is really a worst case scenario.

danielballan:
bump @tacaswell

caspervdw:
Thanks @danielballan. To me personally it would be great if v0.4 could be out in say two weeks, as I am printing my thesis after then and I'd like to list that as a publication.
But no need to tag before everyone agrees.

nkeim:
`pip install` was omitting the new `linking` directory.

caspervdw:
@tacaswell @nkeim @danielballan  I reimplemented the BallTree, based on the implementation in `sklearn` which supports non-Euclidean metrics. This adds an optional dependency. The tests in radial coordinate pass locally:

https://github.com/caspervdw/trackpy/tree/link-custom-metric

I rather keep this separate from this PR, so I'll wait until this is merged.

tacaswell:
Given how much I have checked out of this project :+1: on merging as is.  If I need to run my thesis again I'll use an old version of trackpy ;)

caspervdw:
Good to merge then. @danielballan @nkeim  I leave it up to you.

nkeim:
Merged. Thanks @caspervdw for initiating this project and making the changes!! 🎉

PR #445, opened 2017-07-12
Partial linking

caspervdw:
Following up on our offline communication, this implements two functions that enable the partial linking of a DataFrame. This is split in two steps: first, linking is done on a range of frames, either through the convenience function `link_partial`, or through other means.

Then, some administration needs to be done to reconnect the new trajectories to the surrounding ones. This is done in `reconnect_traj_patch`. I also added some basic testing.

The code builds on the refactor #442: only the 2 most recent commits are for this PR.
@nkeim @tacaswell @danielballan @alexlib


caspervdw:
Rebased

caspervdw:
@alexlib Do you have any results yet with this partial linking?

alexlib:
@caspervdw unfortunately I do not have any results yet. I'm not yet fully into the way it works and it will take me some time to understand the logic. I was also interested in relinking trajectories based on other properties, rather than position, for instance, on their velocity. If you have some clue how to use trackpy for this task, it can be great.

caspervdw:
@alexlib If you have a dataframe that includes the velocities of the particles at each time point in the track, linking based on velocity can be done by putting the kwarg `pos_columns` away from its default value `['x', 'y']`, as follows: `link_df(..., pos_columns=['vel_x', 'vel_y'])`

In that case, you would consider velocity space as an Euclidean space and compute distances accordingly. Custom distance functions are also within the possibilities after #448 is merged.

alexlib:
This is a great thing, I didn't think is possible.

Is it also possible to use link_df(..., pos_columns=['x', 'y', 'vel_x',
'vel_y'])
i.e. using 4 columns  (or actually 6 for 3d) for linking?

caspervdw:
@alexlib Yes it works for any number of dimensions. You can also 'weigh' the dimensions using an anisotropic search_range (e.g. `search_range=(10,10,5,5)`).

caspervdw:
@nkeim I merged in the master branch so that this PR is up to date. This feature should go into the next release I think. Could you have a look and push the merge button at some point?

trackpy/linking/partial.py
```diff
@@ -0,0 +1,155 @@
...
+
+
+def coords_from_df_partial(df, pos_columns, t_column, link_frame_nos):
+    for frame_no in link_frame_nos:
+        yield frame_no, df.loc[df[t_column] == frame_no, pos_columns].values
+
+
+def link_partial(f, search_range, link_range,
+                 pos_columns=None, t_column='frame', **kwargs):
+    """Link a DataFrame of coordinates into trajectories.
```
nkeim: This first line should distinguish this from other linking functions.

trackpy/linking/partial.py
```diff
@@ -0,0 +1,155 @@
...
+        Reduce search_range by multiplying it by this factor.
+    link_strategy : {'recursive', 'nonrecursive', 'numba', 'drop', 'auto'}
+        algorithm used to resolve subnetworks of nearby particles
+        'auto' uses numba if available
+        'drop' causes particles in subnetworks to go unlinked
+
+    Returns
+    -------
+    DataFrame with added column 'particle' containing trajectory labels.
+    The t_column (by default: 'frame') will be coerced to integer."""
```
nkeim: Should `"""` be on a new line to better separate docstring from code?

nkeim: Add something like: 

```
A large dataset can be divided into several patches (time intervals) and linked 
separately with this function, for e.g. parallel processing. The results can then 
be stitched back together with reconnect_traj_patch().

Notes
-----
The memory option cannot retain particles at the boundaries between 
patches. Likewise, if a predictor depends on past trajectories, it may 
work poorly at the start of each patch.
```

nkeim:
@caspervdw It's great to have a solid, tested way to perform these delicate operations! I agree that it should go in the next release and people can start experimenting with it. However we really need a tutorial to explain how you can use this to do parallel linking, both for a modestly-sized dataset, and for one that is too big to fit in memory.

trackpy/linking/partial.py
```diff
@@ -0,0 +1,155 @@
...
+        reconnect_traj_patch(f, (start, stop), '_old_particle', t_column)
+        f.drop('_old_particle', axis=1, inplace=True)
+
+    return f
+
+
+def reconnect_traj_patch(f, link_range, old_particle_column, t_column='frame'):
+    """Reconnect the trajectory inside a range of frames to the trajectories
+    outside the range. Requires a column with the original particle indices."""
+    # TODO Does not yet work fully in combination with memory
```
nkeim: I would remove this comment since I'm pretty sure this is impossible to solve generally.

trackpy/linking/partial.py
```diff
@@ -0,0 +1,155 @@
...
+    ----------
+    f : DataFrame
+        The DataFrame must include any number of column(s) for position and a
+        column of frame numbers. By default, 'x' and 'y' are expected for
+        position, and 'frame' is expected for frame number. See below for
+        options to use custom column names.
+    search_range : float or tuple
+        the maximum distance features can move between frames,
+        optionally per dimension
+    link_range : tuple
```
nkeim: This needs to specify the format of the tuple.

nkeim:
Upon further review, I'm thoroughly confused about how this feature is meant to be used. Why is `reconnect_traj_patch()` called by `link_partial`? Could you give a simple example?

caspervdw:
@nkeim  Thanks for reviewing. I will have a look at your comments!

Concerning your question about the use of `reconnect_traj_patch()`; this is for the case a dataframe is already linked, and a patch is linked again. Say we have two particles in a linking patch that 'cross over', and they were linked as follows initially:
```
   frame 3    a   b
   frame 4     ab
   frame 5    a   b
```
The `relink` routine starts at frame 3 with the already existing labels, but arrives at a different outcome:
```
   frame 3    a   b
   frame 4     ba
   frame 5    b   a
```

In that case, the function `reconnect_traj_patch` is necessary to adapt the labels (`a -> b`, `b -> a`) in the remainder of the dataframe.

There are also some tests that check this, I will have a look whether this crossover example is in there.

nkeim:
@caspervdw Thanks!! That makes sense and is very cool. Is it easy to include some pseudocode-type sketch of how to use the API in that case, and how to use it for parallel linking (assuming that is also a use case)? Do you think that a tutorial is required in order for users to appreciate how to use this feature?

caspervdw:
@nkeim  I pushed the docstring! Ready to merge

danielballan:
> It's great to have a solid, tested way to perform these delicate operations! 

Agreed! Very cool feature.

nkeim:
Looks good! I'm looking forward to a tutorial, perhaps even from someone else who uses this. I realized also that the next step for `reconnect_traj_patch()` is to support streaming data. Thanks @caspervdw for adding this feature!

PR #448, opened 2017-07-27
Link custom metric

caspervdw:
This reimplements the kwarg `neighbor_strategy`, but the BTree is now based on the BallTree from scikit-learn, allowing for custom metrics.

To define a custom metric, you can:
 - pass a function as `to_eucl` that converts an `(N, ndim)` array of coordinates to an `(N, ndim2)` array of Euclidean coordinates (fast)
 - pass a function as `dist_func` that takes two `(ndim,)` arrays of positions and returns a distance (`float`). See [here](http://scikit-learn.org/stable/modules/generated/sklearn.neighbors.DistanceMetric.html)

I had to change the default rescaling of coordinates with `search_range`, that is why this PR is rather extensive. We have some (maybe negligible) performance increase for isotropic search ranges, because coordinates are not rescaled.

### DIFF
--- a/trackpy/linking/linking.py
+++ b/trackpy/linking/linking.py
@@ -8,13 +8,13 @@ import itertools, functools
 
 import numpy as np
 
-from ..utils import (default_pos_columns, guess_pos_columns,
-                     validate_tuple, pandas_sort)
+from ..utils import (guess_pos_columns,
+                     validate_tuple, is_isotropic, pandas_sort)
 from ..try_numba import NUMBA_AVAILABLE
 from .utils import (Point, TrackUnstored, points_from_arr,
                     coords_from_df, coords_from_df_iter,
                     SubnetOversizeException)
-from .subnet import TreeFinder, Subnets, split_subnet
+from .subnet import HashBTree, HashKDTree, Subnets, split_subnet
 from .subnetlinker import (subnet_linker_recursive, subnet_linker_drop,
                            subnet_linker_numba, subnet_linker_nonrecursive)
 
@@ -47,10 +47,6 @@ def link_iter(coords_iter, search_range, **kwargs):
         t, coords = 0, val
     else:
         t, coords = val
-
-    #  obtain dimensionality
-    ndim = coords.shape[1]
-    search_range = validate_tuple(search_range, ndim)
 
     # initialize the linker and yield the particle ids of the first frame
     linker = Linker(search_range, **kwargs)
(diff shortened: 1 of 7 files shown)


caspervdw:
@tacaswell @danielballan @nkeim Bumping for the release of v0.4

trackpy/linking/linking.py
```diff
@@ -96,15 +92,23 @@ def link(f, search_range, pos_columns=None, t_column='frame', **kwargs):
         algorithm used to resolve subnetworks of nearby particles
         'auto' uses numba if available
         'drop' causes particles in subnetworks to go unlinked
+    neighbor_strategy : {'KDTree', 'BTree'}
+        algorithm used to identify nearby features. Default 'KDTree'.
+    to_eucl : function, optional
```
nkeim: Is this something that could be done by just processing the data before linking? As opposed to `dist_func`, which needs to be incorporated here. I am just thinking about our trajectory with data pre-processing in pims.

trackpy/linking/subnet.py
```diff
@@ -106,35 +89,147 @@ def coords_df(self):
...
+    @property
+    def coords_mapped(self):
+        if not self._clean:
+            self.rebuild()
+        if self._btree is None:
+            return np.empty((0, self.ndim))
+        else:
+            return self._btree.data
 
+    def rebuild(self):
```
nkeim: So great that prediction can work with this new code!

trackpy/tests/test_linking.py
```diff
@@ -653,9 +697,50 @@ def test_drop_link(self):
...
+        # link in normal (2D Euclidean) coordinates
+        expected = self.link(f, search_range)
+
+        # compute radial coordinates
+        f_radial = f.copy()
+        f_radial['angle'] = np.arctan2(f_radial['y'], f_radial['x'])
+        f_radial['r'] = np.sqrt(f_radial['y']**2 + f_radial['x']**2)
+        # leave x, y for the comparison at the end
+
+        def dist_func(a, b):
```
nkeim: This is monumentally awesome.

nkeim: Upon merge, we should create an issue to write a tutorial or a short example about how to do this.

nkeim:
As expected, no performance regression.

trackpy/linking/linking.py
caspervdw: Yes this `to_eucl` might be done as preprosessing as well. We do need the implementation anyway for the anisotropic-pixel size rescaling. Might as well expose it to the user.

trackpy/tests/test_linking.py
caspervdw: Yes this really would have helped me in my PhD! And a tutorial would be great to have.

caspervdw:
@nkeim I will have some time to push the release of v0.4 during the next couple of weeks only, could you please have another look so that we can finish this? I guess #445 can wait for a further release.

nkeim:
@caspervdw Thanks. I don't see anything significant that still needs addressing, but I will get back to you by Monday at the latest. (Classes just started here.)

PR #457, opened 2017-11-11
Fix rolling sum parameter in compute_drift.

charlesreid1:
This fixes an issue with how the rolling sum parameter was being used.

The issue:
The pandas rolling() function returns a Rolling object, which does not
have a cumulative sum function cumsum(). The cumsum() function is called
on the trajectory dataframe, which works fine if no rolling sum
parameter is passed, but if there is a rolling sum parameter passed,
the cumsum() function is being called on a Rolling object, which is
undefined, causing compute_drift() to crash with an exception
(Rolling object does not have a cumsum() function).

The fix:
This fixes the problem by calling sum() on the Rolling object,
which returns a dataframe. The cumsum() function can now be called
on the function.

To do:
Add a test for compute_drift() function that uses a rolling sum parameter.
This problem would have been caught with more comprehensive test
coverage of compute_drift().

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,9 +5,9 @@ matrix:
   include:
     # "Legacy" environments: oldest supported versions, without and with numba
     - python: "2.7"
-      env: DEPS="numpy=1.12 scipy=0.18 matplotlib=1.5 pillow==2.1 pandas=0.16 scikit-image=0.9 pyyaml pytables" BUILD_DOCS=false
+      env: DEPS="numpy=1.12 scipy=0.18 matplotlib=1.5 pillow==2.1 pandas=0.16 scikit-image=0.10 pyyaml pytables" BUILD_DOCS=false
     - python: "2.7"
-      env: DEPS="numpy=1.12 scipy=0.18 matplotlib=1.5 pillow==2.5 pandas=0.16.0 scikit-image=0.9 pyyaml numba=0.13.4 pytables scikit-learn" BUILD_DOCS=false
+      env: DEPS="numpy=1.12 scipy=0.18 matplotlib=1.5 pillow==2.5 pandas=0.16.0 scikit-image=0.10 pyyaml numba=0.13.4 pytables scikit-learn" BUILD_DOCS=false
     # "Recommended" environments: More recent versions, for Py2 and Py3.
     - python: "2.7"
       env: DEPS="libgfortran=1.0 numpy=1.12 scipy=0.18 matplotlib=1.5 pillow==2.9 pandas=0.16 scikit-image=0.11 pyyaml numba=0.20 pytables pyfftw scikit-learn" BUILD_DOCS=false
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,14 +5,14 @@ matrix:
   include:
     # "Legacy" environments: oldest supported versions, without and with numba
     - python: "2.7"
-      env: DEPS="numpy=1.8.2 scipy=0.12.0 matplotlib=1.3 pillow==2.1 pandas=0.13.0 scikit-image=0.9 pyyaml pytables" BUILD_DOCS=false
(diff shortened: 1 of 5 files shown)


charlesreid1:
The Travis build is breaking with the error below. I added a test that uses the `pandas_rolling()` function, which is breaking because the `if is_pandas_since_018` block is not working as it should. This is not due to the principal change, which adds `mean()` to the `df.rolling()` calls when computing a rolling average drift term.

```
======================================================================

ERROR: test_no_drift (trackpy.tests.test_motion.TestDrift)

----------------------------------------------------------------------

Traceback (most recent call last):

  File "/home/travis/build/soft-matter/trackpy/trackpy/tests/test_motion.py", line 77, in test_no_drift

    actual_rolling = tp.compute_drift(self.dead_still,3)

  File "/home/travis/build/soft-matter/trackpy/trackpy/motion.py", line 280, in compute_drift

    dx = pandas_rolling(dx, smoothing, min_periods=0)

  File "/home/travis/build/soft-matter/trackpy/trackpy/utils.py", line 332, in _pandas_rolling_pre_018

    return df.rolling_mean(window, *args, **kwargs)

  File "/home/travis/mc/envs/testenv/lib/python2.7/site-packages/pandas/core/generic.py", line 1602, in __getattr__

    (type(self).__name__, name))

AttributeError: 'DataFrame' object has no attribute 'rolling_mean'
```

I will continue to push to this PR to get this fixed, but if @casparvdw wants to jump in to shed some light on commit 16b642279056e5459f172ed5a4cbf1dd86515e1e or any of the other version-specific pandas functionality, that would be much appreciated.

charlesreid1:
Updating with more info:
* Travis uses pandas version 0.13.0 (which is from 2014, and I strongly suggest updating, since Pandas is actively developed and its API changes fast).
* Version 0.13.0 implements the [`rolling_mean()`](http://pandas.pydata.org/pandas-docs/version/0.13/generated/pandas.rolling_mean.html?highlight=rolling#pandas.rolling_mean) function from the module level, as in, `pd.rolling_mean(my_df)`, while the `_pandas_rolling_pre_018()` method in utils implements it as `df.rolling_mean()` (**this implementation has never been correct!**)
* Version 0.18.1, released in May 2016, did away with `pd.rolling_mean(df)` and updated it with `df.rolling().mean()`

More comprehensive test coverage FTW.

charlesreid1:
I guess I opened a can of worms by fixing these issues. I'm unable to replicate the travis test environment in a virtualenv:

```
$ pip2 install numpy==1.8.2 scipy==0.12.0 matplotlib==1.3 pillow==2.1 pandas==0.13.0 scikit-image==0.9 pyyaml pytables
```

Raises issues with scikit-image (due to missing Cython) and pytables (no matching distribution found).

charlesreid1:
I'm afraid I've opened a can of worms, I'm totally unable to replicate the Travis test environment in a virtualenv. Running the nose tests result in dozens of failures like this:

```
TypeError: The numpy boolean negative, the `-` operator, is not supported, use the `~` operator or the logical_not function instead.
```

To replicate the environment, I also needed to install Cython (not listed in the travis.yml file?):

```
$ pip2 install cython==0.17
$ pip2 install numpy==1.8.2 scipy==0.12.0 matplotlib==1.3 pillow==2.1 pandas==0.13.0 scikit-image==0.9 pyyaml pytables
```

I'm a bit puzzled, and wondering if it is even worth the effort to do this, since it's to solve a bug unrelated to the one I'm fixing, and these versions are years out of date. If any of the developers want to jump in here, I'd appreciate it...

charlesreid1:
Closing this pull request. I plan to split it into two parts:
* One PR to fix the rolling sum calculation, leaving the tests alone (so Travis CI passes)
* One PR to update tests

I would like to get in touch with one of the project devs before embarking on test updates.

Issue #401, opened 2016-08-24
Default value of smoothing_size


caspervdw:
The boxcar background subtraction introduces a bias for features that are closer than `smoothing_size` + feature diameter together. This is described in my paper on [arXiv](http://arxiv.org/abs/1607.08819), page 11, top paragraph.

In the crocker-grier paper, the boxcar size is defined as `2w+1`, in which w is one pixel larger than the feature radius. The same as our `diameter`. However, we use `2 * diameter + 1` as a default value for the boxcar subtraction. This is a clear difference with the CG algorithm.

I am not sure if this was a choice, or if this is just a programming bug. If we choose to change it, the bandpass function (over [here](https://github.com/soft-matter/trackpy/blob/master/trackpy/preprocessing.py#L59)), needs to be changed.

@nkeim do you know what the matlab code does exactly?

nkeim:
In the Blair-Dufresne MATLAB code, boxcar is done with a 1D kernel of size 2`lobject`+1, where

```
%               lobject: (optional) Integer length in pixels somewhat 
%                       larger than a typical object. Can also be set to 
%                       0 or false, in which case only the lowpass 
%                       "blurring" operation defined by lnoise is done,
%                       without the background subtraction defined by
%                       lobject.  Defaults to false.
```

The default is to not do the boxcar at all. So that example is no help.

caspervdw:
Then I am :+1: on sticking to the Crocker Grier paper. We absolutely need background subtraction, but the current value introduces a bias in closely spaced features.

nkeim:
This sounds like a good idea, but ideally this should be grouped with other major changes. 0.4 or 0.5?

caspervdw:
I think that together with the adapted gray dilation, the new linker, and the modular approach to `locate`, we are already heading for that. That would be v0.4.

nkeim:
Right! Thanks. Assigning that milestone now.

caspervdw:
Thanks!

PR #465, opened 2017-12-02
Smoothing size

caspervdw:
This tackles #401 for the v0.4 release

### DIFF
--- a/trackpy/feature.py
+++ b/trackpy/feature.py
@@ -191,8 +191,9 @@ def locate(raw_image, diameter, minmass=None, maxsize=None, separation=None,
         Width of Gaussian blurring kernel, in pixels
         Default is 1. May be a tuple, see diameter for details.
     smoothing_size : float or tuple
-        Half size of boxcar smoothing, in pixels
-        Default is diameter. May be a tuple, see diameter for details.
+        The size of the sides of the square kernel used in boxcar (rolling
+        average) smoothing, in pixels
+        Default is diameter. May be a tuple, making the kernel rectangular.
     threshold : float
         Clip bandpass result below this value. Thresholding is done on the
         already background-subtracted image.
@@ -459,8 +460,9 @@ def batch(frames, diameter, minmass=100, maxsize=None, separation=None,
         Width of Gaussian blurring kernel, in pixels
         Default is 1. May be a tuple, see diameter for details.
     smoothing_size : float or tuple
-        Size of boxcar smoothing, in pixels
-        Default is diameter. May be a tuple, see diameter for details.
+        The size of the sides of the square kernel used in boxcar (rolling
+        average) smoothing, in pixels
+        Default is diameter. May be a tuple, making the kernel rectangular.
     threshold : float
         Clip bandpass result below this value.
         Default, None, defers to default settings of the bandpass function.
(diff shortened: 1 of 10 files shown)


nkeim:
The changes to the code look good! It looks like some of the feature-finding tests need to be updated to match the new API. Is that true?

danielballan:
Any reason not to keep `minmass_version_change` around for users who pick up some old work from v0.2.x and migrate it straight to v0.4.x?

caspervdw:
@danielballan  You're right, we should keep the `minmass_version_change`. I now figured out what changed in this PR so that the original tests broke. Because of a reduced (and square! see #361) boxcar kernel, features are dimmer after the bandpassing, leading to lower `mass` values.

This makes sense: a rolling average with a smaller kernel is less smoothed out than one with a larger kernel.  If you subtract this from the original image, the smaller the kernel, the dimmer the features. I explained this in the docstring of the new `minmass_v04_change`.

Counterintuitively, the locate precision increased in this PR: `ep` is now always larger than the real precision (see last commit to `test_ep`). One reason for this is described in #401. I now added an explicit test for this. `ep` can be seen as an upper boundary to the precision.

caspervdw:
Tests pass! This turned out to be a significant change in the locate outcome. See the list in my previous post, which I updated.

trackpy/feature.py
```diff
@@ -73,7 +73,84 @@ def minmass_version_change(raw_image, old_minmass, preprocess=True,
 
     scale_factor = scalefactor_to_gamut(image, dtype)
 
-    return int(old_minmass / scale_factor)
+    return old_minmass / scale_factor
+
+
+def minmass_v04_change(raw_image, old_minmass, diameter, preprocess=True,
+                       old_smoothing_size=None, new_smoothing_size=None):
+    """Convert minmass value from v0.3 o v0.4.
```
nkeim: v0.3 **to** v0.4

trackpy/feature.py
```diff
@@ -73,7 +73,84 @@ def minmass_version_change(raw_image, old_minmass, preprocess=True,
...
+
+    From trackpy version 0.4.0, the default image preprocessing is changed.
+    Before version 0.4.0 a blurred out image (rolling or boxcar average) with
+    a circular kernel with radius `diameter` was subtracted from the image
+    before refinement and mass calculation. From version 0.4.0, this has
+    changed to a square kernel with sides `diameter`, more or less twice as
+    small. This increasing tracking accuracy, and reduces the mass.
```
nkeim: increases

trackpy/preprocessing.py
```diff
@@ -116,18 +118,23 @@ def bandpass(image, lshort, llong, threshold=None, truncate=4):
     See Also
     --------
     lowpass, boxcar, legacy_bandpass, legacy_bandpass_fftw
+
+    Notes
+    -----
+    The boxcar size changed in v0.4: before, one side of the square kernel
```
nkeim: Maybe note that the shape also changed?

caspervdw:
@nkeim Done and ready to merge.

caspervdw:
This seems to me a Travis hiccup. Let's see what it does tomorrow.

Issue #475, opened 2018-03-16
Using the installation notes for Linux on Win10


Resonanz:
This is a really great and easy way of installing trackpy on Linux on Windows 10 (requires a newish build of Win10):

https://github.com/soft-matter/trackpy/wiki/The-Ten-Commandments-for-configuring-trackpy-on-Anaconda-on-Debian-Linux-on-Windows-10-using-WSL

If you install pims or trackpy inside a user-created conda environment rather than in the root environment, then expect errors as I document below. After many attempts to get this right LOL..... my Wiki instructions **should** now work perfectly and be error-free (2018-03-17, trackpy 0.3.3, python 3.6).

UPDATE: OH DAMMIT, THIS STILL FAILS. SO THIS APPEARS TO BE HOW IT WORKS....

CASE 1: INSTALL PIMS AND TRACKPY IN THE ROOT ENV ---> CANNOT IMPORT PIMS OR TRACKPY FROM THE COMMAND LINE INSIDE A CREATED CONDA ENV.

CASE 2: INSTALL PIMS AND TRACKPY IN A CREATED CONDA ENV ---> CANNOT IMPORT PIMS OR TRACKPY INTO JUPYTER NOTEBOOK.

SIGH.

Resonanz:
One failure: `PermissionError(13, 'Permission denied')`

This is the fail (can see below):

certifi 2018.1.18: #####################################################################7 | 75%


**FULL OUTPUT**

(py36) rob@WIN10-ELITEDESK-i7:~$ conda install -c conda-forge trackpy
Solving environment: done


==> WARNING: A newer version of conda exists. <==
  current version: 4.4.10
  latest version: 4.4.11

Please update conda by running

    $ conda update -n base conda



## Package Plan ##

  environment location: /home/rob/anaconda3/envs/py36

  added / updated specs:
    - trackpy


The following packages will be downloaded:

    package                    |            build
    ---------------------------|-----------------
    ca-certificates-2018.1.18  |                0         140 KB  conda-forge
    lzo-2.10                   |                0         234 KB  conda-forge
    fontconfig-2.12.6          |                0         943 KB  conda-forge
    libxml2-2.9.7              |                0         5.0 MB  conda-forge
    slicerator-0.9.8           |             py_1          10 KB  conda-forge
    bzip2-1.0.6                |                1         310 KB  conda-forge
    jinja2-2.10                |           py36_0         183 KB  conda-forge
[...]

PR #476, opened 2018-03-18
DOC: Improve installation instructions

caspervdw:
Related to #475


nkeim:
I think this is a great idea and will hopefully prevent some headaches. And I didn't know about `nb_conda`! Thanks, @Rezonanz and @caspervdw . I am OK with merging.

Issue #472, opened 2018-01-26
BUG: Particle added to track twice


nkeim:
I tried out the master branch for my regular analysis workflow. I got as far as running linking on some of my smaller movies, in this case ~54k particles for 19200 frames. It looks like they were failing because of an exception in `linking.utils.Point.add_to_track`: `trying to add a particle already in a track`. This occurred at some arbitrary point in each movie, after hundreds of frames.

This was regular linking, not `find_link`. Prediction and `adaptive_step` were in use. The data are generally messy and are no doubt a good place to find edge cases. Linking works fine with the v0.3 implementation.

I won't have time to investigate for a few more weeks, but I thought I would report this now since (a) it obviously needs a closer look before we release v0.4, and (b) @caspervdw might brilliantly think of the cause and save me some work 😉 .

caspervdw:
@nkeim I seem to have missed your issue, sorry! The unittests have apparently missed an edge case, good that this came up before the release. 

For sure the prediction and `adaptive_step` are the newest and least tested parts. I suggest turning them off independently so that we might narrow down the issue to one of them (and hopefully not the core linking code!). Do you have `memory` on? That might be a third thing to turn off/on. Without these tests I have no clue where to start looking.

nkeim:
@caspervdw agreed, those are the likely culprit. I did not have memory turned on. I think the best way is to turn on prediction diagnostics and then run it on the real data, and hope that I can identify the anomaly with `pdb`, or at least extract enough state to make it reproducible quickly. 

This is creeping toward the top of my to-do list!

caspervdw:
@nkeim any results yet? Would be nice to release v0.4 soon!

nkeim:
@caspervdw Today I'm working on reproducing the bug efficiently.

caspervdw:
Closing through #477  :tada:

PR #477, opened 2018-03-19
Fix linking duplicate bug

nkeim:
This addresses #472. The problem was in the `split_subnet()` function that is called when oversize subnets are broken into smaller daughter subnets, as part of the adaptive search feature. In addition to creating smaller sets of source and destination particles, each source particle's `forward_cands` list must also be pruned to remove candidates that are beyond the new `search_range`. This is because `forward_cands` is used by the subnet linker. If `forward_cands` is not pruned, then effectively speaking, the same destination particle may appear in multiple daughter subnets, and so it may be added to multiple tracks.

I was unable to formulate a simple test for this problem, but I did add a comment at the relevant portion of the code, so that hopefully we will be sure to account for this detail in any future implementation 🙂.

This also restores status messages to the "traditional" linking functions (e.g. `link_df_iter`).

Some additional notes about the hunt for this bug:
1. I first encountered this bug when using prediction on a real data set. Unfortunately, in general prediction depends on all preceding frames, making it hard to reproduce the bug quickly. I used prediction diagnostics to obtain the predicted coordinates that actually triggered the bug. I then saved the predicted coordinates at *t*, and the observed coordinates at *t*+1, to a file so that I could reproduce the bug in seconds instead of hours.
2. The code uses `set` in many places so that the particles' `uuid`s and the subnet IDs are not deterministic. I used a recipe for an `OrderedSet` to fix this so that I could reliably identify particles.
3. I used the debugger built into PyCharm. A lot.

### DIFF
--- a/trackpy/linking/linking.py
+++ b/trackpy/linking/linking.py
@@ -55,6 +55,7 @@ def link_iter(coords_iter, search_range, **kwargs):
 
     for t, coords in coords_iter:
         linker.next_level(coords, t)
+        logger.info("Frame {0}: {1} trajectories present.".format(t, len(linker.particle_ids)))
         yield t, linker.particle_ids
 
 
--- a/trackpy/linking/subnet.py
+++ b/trackpy/linking/subnet.py
@@ -263,17 +263,30 @@ def assign_subnet(source, dest, subnets):
 
 
 def split_subnet(source, dest, new_range):
-    # Clear the subnets and candidates for all points in both frames
+    """Break apart a subnet by using a reduced search_range."""
     subnets = dict()
+    # Clear source particles' subnets, and prune their forward candidates
+    # according to new_range.
+    # The pruning step is crucial because some subnet linkers ignore
+    # the destination set and use the source particles' forward candidates
+    # exclusively.
     for sp in source:
         sp.subnet = None
+        new_fcs = []
+        for dp, dist in sp.forward_cands:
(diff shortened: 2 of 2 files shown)


caspervdw:
@nkeim  Sorry for this, must have been a big pain to identify the error. After a while I could wrap my head around this part of the code again and I now understand what has been going wrong.

And :+1: for PyCharm!  Merging.

nkeim:
It was kind of fun! Let's hope someone finds another linking bug soon, before we forget again how the code works 😉.

J-F-Louf:
Hello guys, I would first like to thank you for developing Trackpy, I really like it (and I cited it on a recently submitted paper). 

Last week, one of my student and I were running it on "fake" data, and we encountered the same issue - some points, when being very close, were detected 2 times. It seems you addressed this issue 3 weeks ago, but it is not clear to me wether this problem is supposed to be solved or if it's still being fixed? 

I'm more an experimental physicist than a numerical one, but if needed, I'd be happy to help.

All the best

nkeim:
@J-F-Louf I don't think this is the same issue that was fixed by this PR, but of course you should double-check that you are running either v0.3.3 or the latest version of trackpy.

Do you think this is an issue with finding features, or with linking? It would be great if you could create a new issue with the information to reproduce the bug. Thanks!

(Also, looking at my comment above: in hindsight, I got my wish! ✨)

Issue #452, opened 2017-09-17
DOC/API: Add all kwargs to user-facing linking functions?


nkeim:
Since the refactoring of the linking code for `find_link`, the various linking functions have mostly relied on `**kwargs`, and those options are only fully documented for `link()`; the other docstrings refer the user to the `link()` documentation.

This is wonderful from the maintainer's point of view. However I think there is a cost to the user, at least to users like me. Various autocompleters (especially IPython) and linters use the explicit kwarg declarations to provide a great convenience. And IPython's instant documentation is less useful when the docstring is incomplete. This is one maddening (yet forgivable! Sorry @tacaswell ) thing about matplotlib.

What are your opinions about this? I am willing to do the work of re-adding all the declarations and/or docstrings for the v0.4 release. I don't think that maintaining that state in the v0.2 and v0.3 series was particularly onerous, but it was hardly automatic and it's worth discussing.

caspervdw:
I know the issue @nkeim and I agree that simply copy-pasteing the kwargs and docstring is the best solution for us.

danielballan:
:+1: -- the benefit to users outweighs the cost to "developers"

nkeim:
OK, I  have tagged this for v0.4 and will make the PR.

caspervdw:
@nkeim  I updated the docstrings in the self-merged #478, making use of a nice Sphinx trick that still lets us use the `**kwargs` in the actual code, but the Sphinx documentation does display the right arguments:

http://soft-matter.github.io/trackpy/dev/generated/trackpy.link_iter.html#trackpy.link_iter

And IPython displays the following docstring:
```
Signature: tp.link(f, search_range, pos_columns=None, t_column='frame', **kwargs)
Docstring:
link(f, search_range, pos_columns=None, t_column='frame', memory=0,
    predictor=None, adaptive_stop=None, adaptive_step=0.95,
    neighbor_strategy=None, link_strategy=None, dist_func=None,
    to_eucl=None)

Link a DataFrame of coordinates into trajectories.
```

So for autocompletion, you still see the `**kwargs`, but the docstring directly displays the full list. Is this a workable solution?

nkeim:
This works for me!

PR #478, opened 2018-03-24
DOC: Update docstrings for API functions with kwargs

caspervdw:
This tackles #452 

I used an idea from https://matplotlib.org/devel/MEP/MEP10.html and set `autodoc_docstring_signature = True` in Sphinx so that the first line of the docstring might override the call signature. This saves us the error-prone hardcoding of kwargs and default values, while keeping the docstrings nicely informative.


nkeim:
Thanks for doing this! This is not a perfect solution but it is a great compromise. Making sure the call signatures were correct is tricky, and a mistake could break something.

PR #479, opened 2018-03-24
Update travis

caspervdw:
I updated the version pins in travis, see commits.

Also I have bumped `pims` to `v0.4.1`. Not actually necessary, but I think it will probably make solving issues a little simpler if pims-trackpy combinations are similar for most users.


caspervdw:
I adapted Travis slightly, pulling in `pims` from conda-forge now. This makes sure that `pip` does not silently adapt the versions we want to test. Also, I adapted the docbuild part a bit so that again the versions are what we think they are.

A small fix is included for the legacy numba versions.

caspervdw:
Funny thing: if you install PIMS from `conda-forge` and then `trackpy` with `python setup.py install`, you get the following package listing:

```
PIMS                      0.4.1                     <pip>
pims                      0.4.1                      py_1    conda-forge
```

caspervdw:
Travis-github integration is acting weird. The tests pass as can be seen here https://travis-ci.org/caspervdw/trackpy/builds/357783139

nkeim:
This is great!! I'm amazed that we can still test on such old dependencies, especially numba. It's definitely a strength of the project.

PR #480, opened 2018-03-24
Specify mathjax library url on documentation site

prashnts:
In Chrome, the equations written in MathJax format are not rendered since the default sphinx-mathjax plugin fetches it from non-https url. Setting this variable ensures an https library is loaded.

snip 2018-03-25 at 00 44 49


caspervdw:
Thanks  @prashnts

Issue #487, opened 2018-04-06
Linking performance 0.4 vs 0.3.3


apiszcz:
My observations for the same input data and link parameters indicate a performance difference in 'linking' between 0.3.3 and 0.4. 0.4 takes ~12.25x longer to complete than 0.3.3.

0.3.3: 2m31s (7082 trajectories)
0.4: 30m38s, (7060 trajectories)

nkeim:
What's the output of

```
from trackpy.diag import dependencies
dependencies()
```

apiszcz:
```
(Pdb) pp(dependencies())
OrderedDict([('six', '1.11.0'),
             ('numpy', '1.14.2'),
             ('scipy', '1.0.1'),
             ('matplotlib', '2.2.2'),
             ('pandas', '0.22.0'),
             ('scikit-image', None),
             ('pyyaml', None),
             ('pytables', None),
             ('numba', '0.37.0'),
             ('pyfftw', None),
             ('python', '3.5.4')])
```

caspervdw:
This is very odd! We ran reproducibility tests and linking should not change between 0.3.3 and 0.4 (at least for the numba subnetljnker; there was a bug in the pure python one). Also @nkeim measured a slight performance *increase* for the linking. 

I suspect this to be an OS-related issue. For linking, `cKDTree` from `scipy` and of course `numpy`, `numba`, and maybe `pandas` are likely culprits. Are you sure that these dependencies were compiled correctly? Did you observe any warnings at some point? Are you using Windows? If so, I really recommend the conda binaries in isolated environments for reproducible results.

I am curious to see the output of a profiler. In IPython, this is easily done with `%prun tp.link(...)`

apiszcz:
Libraries: Let's think about this, the same libraries work fine with 0.3.3.
Warnings: none.
Windows 10/64, again 0.3.3 is fine.
Thanks for giving more options to consider, in the mean time I'll stick with 0.3.3, and test 0.4 when possible,

caspervdw:
Thanks, that helps to isolate the problem! Next thing: is it possible to try the new `link_iter`? This is for sure better tested than the `link_df_iter`, which is a quite new reimplementation. `link_iter` takes iterables of numpy arrays.

In this way we could isolate any issues associated with `pandas`.

Something like: 

```
# transform to numpy iterable, in memory
coords_iter = [f[['x', 'y']].values for f in f_iter]

particle_ids = []
[...]

PR #488, opened 2018-04-07
Remove numba on/off toggle

caspervdw:
This attempts a quickfix for #487 by removing most of our custom numba-wrapping code.

I am not sure why we still have the `enable_numba` and `disable_numba`, because the user can choose between python and numba linkers / feature finding engines anyway.

Another thing: numba.autojit is not necessary from version 0.12, as we do not support this version anyway, I adapted it to numba.jit.

### DIFF
--- a/trackpy/api.py
+++ b/trackpy/api.py
@@ -26,7 +26,7 @@ from . import predict
 from . import utils
 from . import artificial
 from .utils import handle_logging, ignore_logging, quiet
-from .try_numba import try_numba_autojit, enable_numba, disable_numba
+from .try_numba import try_numba_jit, enable_numba, disable_numba
 
 
 # pims import is deprecated. We include it here for backwards
--- a/trackpy/linking/subnetlinker.py
+++ b/trackpy/linking/subnetlinker.py
@@ -13,7 +13,7 @@ from collections import deque
 import numpy as np
 
 from .utils import SubnetOversizeException
-from ..try_numba import try_numba_autojit
+from ..try_numba import try_numba_jit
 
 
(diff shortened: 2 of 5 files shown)


caspervdw:
Closing this as it doesn't fix anything

nkeim:
Let's not forget about this PR entirely! It might be advisable under a general housecleaning for v0.5.

@apiszcz As far as I know, `enable_numba()` was *never* required; numba is always enabled by default if it is available. The purpose of `disable_numba()` and `enable_numba()` was to allow testing of the un-compiled functions. This was back when numba was still exotic and sometimes temperamental.

apiszcz:
Thank you, will remove in my applications.

apiszcz:
Another observation is when we have two particles within the search range in the same frame the behavior in 0.3.3 was to terminate the trajectory and start a new one. When I get time I will review this situation 0.4

Not sure there are any better options in these cases.

caspervdw:
That is not the intended behavior: when one particle is faced with two linking options, its trajectory should not terminate. Instead it should link to the closest one and the other starts a new trajectory. (Except memory is involved)

apiszcz:
understand, that is what I observer. terminate track one and start track
two.

nkeim:
That is very interesting! When you get a chance please do look into this. In think that in `test_linking.py`, `test_two_nearby_steppers_one_gapped()` is *supposed* to check this scenario.

apiszcz:
The test appears to separate the detections by 'frame' i have two detections in the same frame both within range. Somewhere in the past I provided a sample dataset on github, it might still exist?

PR #486, opened 2018-04-06
FIX: Fix try_numba module name resolution on Windows

nkeim:
This attempts to fix #485.

PR #489, opened 2018-04-08
FIX: Adaptive search bug that could trigger "Too many forward candidates"

nkeim:
This partially addresses #487. When many iterations of adaptive search are required to reduce a subnet to solvable size, there is a "leak" that adds superfluous null particles to the list of forward candidates. This PR fixes that bug, but performance in this case is still much worse than the adaptive search implementation in v0.3.

### DIFF
--- a/trackpy/linking/find_link.py
+++ b/trackpy/linking/find_link.py
@@ -475,6 +475,9 @@ class FindLinker(Linker):
             else:
                 new_cands = set()
 
+            for sp in source_set:
+                sp.forward_cands.sort(key=lambda x: x[1])
+
             # link
             sn_spl, sn_dpl = self.subnet_linker(source_set, dest_set,
                                                 self.search_range)
--- a/trackpy/linking/find_link.py
+++ b/trackpy/linking/find_link.py
@@ -475,10 +475,8 @@ class FindLinker(Linker):
             else:
                 new_cands = set()
 
-            # sort candidates and add in penalty for not linking
-            for _s in source_set:
-                _s.forward_cands.sort(key=lambda x: x[1])
-                _s.forward_cands.append((None, self.search_range))
+            for sp in source_set:
+                sp.forward_cands.sort(key=lambda x: x[1])
 
             # link
             sn_spl, sn_dpl = self.subnet_linker(source_set, dest_set,
--- a/trackpy/linking/find_link.py
+++ b/trackpy/linking/find_link.py
@@ -475,8 +475,10 @@ class FindLinker(Linker):
             else:
                 new_cands = set()
 
-            for sp in source_set:
-                sp.forward_cands.sort(key=lambda x: x[1])
+            # sort candidates and add in penalty for not linking
+            for _s in source_set:
+                _s.forward_cands.sort(key=lambda x: x[1])
+                _s.forward_cands.append((None, self.search_range))
 
             # link
             sn_spl, sn_dpl = self.subnet_linker(source_set, dest_set,
--- a/trackpy/linking/linking.py
+++ b/trackpy/linking/linking.py
@@ -53,9 +53,9 @@ def link_iter(coords_iter, search_range, **kwargs):
         Reduce search_range by multiplying it by this factor.
     neighbor_strategy : {'KDTree', 'BTree'}
         algorithm used to identify nearby features. Default 'KDTree'.
-    link_strategy : {'recursive', 'nonrecursive', 'numba', 'drop', 'auto'}
+    link_strategy : {'recursive', 'nonrecursive', 'hybrid', 'numba', 'drop', 'auto'}
         algorithm used to resolve subnetworks of nearby particles
-        'auto' uses numba if available
+        'auto' uses hybrid (numba+recursive) if available
         'drop' causes particles in subnetworks to go unlinked
     dist_func : function, optional
         a custom distance function that takes two 1D arrays of coordinates and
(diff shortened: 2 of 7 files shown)


caspervdw:
This is looking great, I understand that now performance is even better than before?

apiszcz:
looking 3x faster compared to 0.3.3 and Friday ...

On Mon, Apr 9, 2018 at 2:04 AM, Casper van der Wel <notifications@github.com

trackpy/linking/linking.py
```diff
@@ -503,6 +500,9 @@ def next_level(self, coords, t, extra_data=None):
     def assign_links(self):
         spl, dpl = [], []
         for source_set, dest_set in self.subnets:
+            for sp in source_set:
+                sp.forward_cands.sort(key=lambda x: x[1])
+
```
caspervdw: Am I correct with assuming that not every subnetlinker requires the null particle? Else it's probably best done here?

trackpy/linking/subnet.py
```diff
@@ -274,20 +280,19 @@ def split_subnet(source, dest, new_range):
         sp.subnet = None
         new_fcs = []
         for dp, dist in sp.forward_cands:
-            if dist <= new_range or dp is None:
+            if dist <= new_range:
```
caspervdw: This removes the null candidates, right? Maybe add comment to be clear about it

trackpy/linking/subnetlinker.py
```diff
@@ -427,24 +427,41 @@ def subnet_linker_nonrecursive(source_set, dest_set, search_range, **kwargs):
...
         return [source_set.pop()], [dest_set.pop()]
-    elif len(source_set) == 1 and len(dest_set) == 0:
+    elif lss == 1 and lds == 0:
         # particle is lost. Not possible with default Linker implementation.
         return [source_set.pop()], [None]
 
-    # sort candidates and add in penalty for not linking
+    # Forward candidates were already sorted by Linker.assign_links()
     for _s in source_set:
-        _s.forward_cands.sort(key=lambda x: x[1])
```
caspervdw: The sort should be dropped in the other subnetlinkers as well

trackpy/linking/subnetlinker.py
```diff
@@ -427,24 +427,41 @@ def subnet_linker_nonrecursive(source_set, dest_set, search_range, **kwargs):
...
         _s.forward_cands.append((None, search_range))
```
caspervdw: Not sure if adding the null candidate should be done in the subnetlinker or before the subnetlinker. It seems cleaner to me to do it before? Are there any reasons for doing it like this?

trackpy/linking/utils.py
```diff
@@ -100,6 +100,8 @@ class Point(object):
     :py:meth:`Point.__init__`.  (See :py:class:`~trackpy.linking.PointND` for
     example. )
     '''
+    __slots__ = ['_track', 'uuid', 't', 'pos', 'id', 'extra_data', 
+                 'forward_cands', 'subnet', 'relocate_neighbors', '__dict__']
```
caspervdw: Neat, I looked it up and I didn't know this feature

trackpy/tests/test_linking.py
```diff
@@ -701,6 +701,12 @@ def setUp(self):
         self.linker_opts = dict(link_strategy='numba')
 
 
+class TestHybridLink(SubnetNeededTests):
```
caspervdw: Just a thought: what about making a mock subnetlinker and testing if the subnetlinker gets the expected subnets (sorted / not sorted, with / without null particle)?  In a separate PR of course

trackpy/linking/find_link.py
```diff
@@ -475,6 +475,9 @@ def assign_links(self):
             else:
                 new_cands = set()
 
+            for sp in source_set:
+                sp.forward_cands.sort(key=lambda x: x[1])
```
caspervdw: Same comment as previous: maybe we want to add null particles over here.

caspervdw:
Fantastic!! I left some minor line notes

caspervdw:
@nkeim I found some time to write the test I commented about, also I made the null particle addition more consistent (and tested that)

nkeim:
@caspervdw Thanks! Sorry I have been absent this week. The mock subnet linker is a great idea. I just added a check that would have caught the original bug for this PR!

I added a reply to your comment about where in the code we should add the null candidate. Would it be OK if I reversed your commit? I think that the conceptual separation *and* slightly greater performance are worth the extra lines of code and duplication.

I'm sure you agree that this PR is getting messy! Once we are happy with it, I would like to rebase before merge.

caspervdw:
@nkeim sorry for causing the mess.. I am however not sure if I understand the performance difference... null particles are removed and added anyway by the adaptive step, that I did not change. It’s ok if you revert, maybe keep the commits and don’t rewrite history here ?

nkeim:
@caspervdw On the contrary, I started this mess!

It looks like your rollback commit did restore the code in `split_subnet()` that re-adds the null candidate. I have implemented the alternate scenario in which the (real) candidates are sorted in `linking.py`, and then the null candidate is added as an implementation detail in `subnetlinker.py`.

@apiszcz I think this will definitely be 0.4.1. I'm hoping we can release in the next few days.

caspervdw:
Ok, agreed! Let's look at #486 and then release v0.4.1

Issue #491, opened 2018-04-14
Remove dependency on PIMS


caspervdw:
I noticed some code in `api.py` that imports pims. I think it would be good to really drop PIMS as a dependency of trackpy. It pulls is a significant amount of packages that are not necessary for `trackpy` and adds significant overhead to trackpy.

apiszcz:
+1 concur on this. For people only using the linker PIMS is unnecessary dependency drama.

nkeim:
Agreed. Do you think we should do something like [this](https://github.com/soft-matter/trackpy/blob/fdd8acac9f422f31cf0cf889980d960809d353ce/trackpy/motion.py#L591) for v0.4.1?

nkeim:
Oops, we already do issue a warning!

I still think we should wait until v0.5. We would have to revise the installation instructions.

tacaswell:
:+1:

caspervdw:
As there has been a DeprecationWarning for some time now, I would be in favour of dropping it altogether for v0.4.1.

nkeim:
Sounds good. It won't affect existing users because they (must) already have pims. I will make a PR.

apiszcz:
Congrats on dependency reduction, i look forward to testing 0.4.1 when ready.

magsol:
Just came across TrackPy--absolutely love the work you've done and are continuing to do!

As I've been navigating through the Jupyter notebooks, and I see that you're planning to deprecate PIMS in the next version, I'm curious how you're going to have users instantiate 2D or 3D image stacks? For example, what will replace `pims.ImageSequenceND`?

nkeim:
For recordkeeping purposes I renamed this issue. Thanks @magsol — it's awesome to have a new user taking interest in development. Please keep it up! 😃

magsol:
Ahhh, that actually highlights the exact point of tension @nkeim @caspervdw. Thanks for the clarification!

caspervdw:
Done, release v0.4.1 is on PyPI and conda will follow in the next couple of days!

apiszcz:
Passes my standard performance test on Windows 10/64. Thank you.

nkeim:
🎉 Thank you, @caspervdw for pushing this out, and thank you, @apiszcz for your testing and feedback!

PR #492, opened 2018-04-14
Drop pims dependency

nkeim:
This addresses #491.


conda-recipe/meta.yaml
```diff
@@ -32,7 +31,6 @@ requirements:
     - numexpr  # required for pandas DataFrame.query, used in examples
```
caspervdw: The conda recipe can be dropped altogether, it's an old version, the new one is maintained at conda-forge

nkeim:
Aha! @caspervdw I suppose we will take care of conda-forge when we do the release?

nkeim:
Ah! I wasn't looking for the other pims imports.

- The use of pims in `plots.py` is an enhancement, and the `ImportError` is already handled; we just need to update the error message.
- I don't see pims imported anywhere in `refine/least_squares.py`; there are only references to pims in the docstrings, and the use of pims objects as arguments is optional.
- In `artificial.py`, the only serious use of pims is the `CoordinateReader` class, which is used for many `find_link` tests. It seems like making a pims-free equivalent would be reinventing the wheel. Perhaps we could move this class to `tests/common.py`, and skip most/all `find_link` tests if pims is missing?

If we want to release v0.4.1 very soon, I think it would be reasonable to omit this PR.

caspervdw:
@nkeim  I did some work on this in the train this morning. I think `artificial.py` can function fine without PIMS. Also, for testing purposes, I created a `tests.common.TrackpyImageSequence` that reads images via matplotlib `imread`.

See https://github.com/nkeim/trackpy/pull/7

PR #494, opened 2018-04-20
Revise v0.4.1 release notes

nkeim:
This incorporates my suggestions, plus notes that the use of `__slots__` saves some memory.


caspervdw:
Thanks @nkeim! Merging and I'll do the release after this.

Issue #498, opened 2018-05-06
Index-handling in filter functions and other feedback


lagru:
Hello devs, first of all I'd like to thank you for your work on this small but promising library. I'm currently wrapping parts of your library to use with a command line tool intended for particle tracking at my university. Thanks to your great example section and generally well documented API that process is very easy and satisfying. 

However I also encountered some minor hiccups and annoyances. Maybe you find my feedback in that regard useful:

- Using `pandas.DataFrame` as the primary data container in your library is a choice I'm not sure about and certainly not used to when comparing trackkpy to other scientific libraries. 
  - I don't understand the index-handling for the functions in `filtering.py`. During grouping the index is dropped (why not move the index back to into a column?) and later "frame" is explicitly set as the index. In that last step the keyword `drop=False` is moved which doesn't make sense because any would-be index was dropped during grouping? Is there a reasoning behind this?
To extend on this, wouldn't it be clearer if all functions treated dataframes and indexes the same way: e.g. leave index untouched or at least be explicit about it in the docstring. I was initially confused several times why this or that function chose to set an index. To me that often seemed arbitrary and not intuitive. 
  -  I also think in many cases the docstrings could be more explicit about the required columns and especially the structure of returned dataframes. I think this is a disadvantage when using `pandas.DataFrame` as the primary data container: it obfuscates the in- and output compared to simple arrays provided through multiple well documented arguments. Maybe a type declaration like `df : pandas.DataFrame[frame, particle, x, y]` could be used within the docstrings to make this clearer.
- Your example on parallelized feature location got me thinking why you don't support something equivalent in `trackpy.batch`. Even multi-threading can speed up this processing step for large sets by magnitudes and can easily be accomplished without depending on `ipyparalell`. E.g.:

```python
from functools import partial
from multiprocessing.pool import ThreadPool

def batch_threaded(frames, locate_kw, threads=4, report_hook=None):
    if not report_hook:
        def report_hook(): pass
    func = partial(trackpy.locate, **locate_kw)
    with ThreadPool(threads) as pool:
        stats = []
        for i, frame_stats in enumerate(pool.imap(func, frames)):
            frame_stats["frame"] = i
            stats.append(frame_stats)
            report_hook()
    return pd.concat(stats, ignore_index=True)
```

- Although calculating the displacement of linked particles between consecutive frames isn't hard I think a function doing that might be an useful addition to this library. Would you agree?

[...]

PR #499, opened 2018-05-20
ENH: Parallel batched feature location

lagru:
As suggested in #498 I propose an optionally threaded version of `trackpy.batch`. The new function takes two additional arguments:

     threads : integer, optional
         If 0, no threading is used. An integer greater 0 is treated as the
         number of threads used to process frames in parallel. If `threads` is
         None then the number returned by ``os.cpu_count()`` is used.
     report_hook : function, optional
         Specify a custom function to log information for each processed frame.
         Must accept two integer arguments `frame_no` and `detected_features`.
         If not given the process is passed to trackpy's default logger.

Please let me know your thoughts about these changes.

### DIFF
--- a/trackpy/feature.py
+++ b/trackpy/feature.py
@@ -3,6 +3,8 @@ from __future__ import (absolute_import, division, print_function,
 import six
 import warnings
 import logging
+from functools import partial
+from multiprocessing.pool import ThreadPool
 
 import numpy as np
 import pandas as pd
@@ -456,7 +458,7 @@ def batch(frames, diameter, minmass=100, maxsize=None, separation=None,
           noise_size=1, smoothing_size=None, threshold=None, invert=False,
           percentile=64, topn=None, preprocess=True, max_iterations=10,
           filter_before=None, filter_after=None, characterize=True,
-          engine='auto', output=None, meta=None):
+          engine='auto', output=None, meta=None, threads=0):
     """Locate Gaussian-like blobs of some approximate size in a set of images.
 
     Preprocess the image by performing a band pass and a threshold.
@@ -521,6 +523,10 @@ def batch(frames, diameter, minmass=100, maxsize=None, separation=None,
         If specified, information relevant to reproducing this batch is saved
         as a YAML file, a plain-text machine- and human-readable format.
         By default, this is None, and no file is saved.
+    threads : integer or None, optional
+        If 0, no threading is used. An integer greater 0 is treated as the
+        number of threads used to process frames in parallel. If `threads` is
+        None then the number returned by ``os.cpu_count()`` is used.
 
     Returns
     -------
@@ -571,27 +577,46 @@ def batch(frames, diameter, minmass=100, maxsize=None, separation=None,
             # Interpret meta to be a file handle.
             record_meta(meta_info, meta)
 
-    all_features = []
-    for i, image in enumerate(frames):
-        features = locate(image, diameter, minmass, maxsize, separation,
-                          noise_size, smoothing_size, threshold, invert,
-                          percentile, topn, preprocess, max_iterations,
-                          filter_before, filter_after, characterize,
-                          engine)
-        if hasattr(image, 'frame_no') and image.frame_no is not None:
-            frame_no = image.frame_no
-            # If this works, locate created a 'frame' column.
-        else:
-            frame_no = i
-            features['frame'] = i  # just counting iterations
-        logger.info("Frame %d: %d features", frame_no, len(features))
-        if len(features) == 0:
-            continue
+    # Prepare wrapped function for mapping to `frames`
+    curried_locate = partial(
+        func=locate,
+        diameter=diameter, minmass=minmass, maxsize=maxsize,
+        separation=separation, noise_size=noise_size,
+        smoothing_size=smoothing_size, threshold=threshold, invert=invert,
+        percentile=percentile, topn=topn, preprocess=preprocess,
+        max_iterations=max_iterations, filter_before=filter_before,
+        filter_after=filter_after, characterize=characterize, engine=engine
+    )
+    # Choose threaded or normal map function
+    if threads == 0:
+        pool = None
+        map_func = map
+    else:
+        pool = ThreadPool(processes=threads)
+        map_func = pool.imap
 
-        if output is None:
-            all_features.append(features)
-        else:
-            output.put(features)
+    try:
+        all_features = []
+        for i, features in enumerate(map_func(curried_locate, frames)):
+            image = frames[i]
+            if hasattr(image, 'frame_no') and image.frame_no is not None:
+                frame_no = image.frame_no
+                # If this works, locate created a 'frame' column.
+            else:
+                frame_no = i
+                features['frame'] = i  # just counting iterations
+            logger.info("Frame %d: %d features", frame_no, len(features))
+            if len(features) == 0:
+                continue
+
+            if output is None:
+                all_features.append(features)
+            else:
+                output.put(features)
+    finally:
+        if pool:
+            # Ensure correct termination of ThreadPool
+            pool.terminate()
 
     if output is None:
         if len(all_features) > 0:
--- a/trackpy/feature.py
+++ b/trackpy/feature.py
@@ -4,7 +4,7 @@ import six
 import warnings
 import logging
 from functools import partial
-from multiprocessing.pool import ThreadPool
+from multiprocessing.pool import Pool
 
 import numpy as np
 import pandas as pd
@@ -454,12 +454,8 @@ def locate(raw_image, diameter, minmass=None, maxsize=None, separation=None,
     return refined_coords
 
 
-def batch(frames, diameter, minmass=100, maxsize=None, separation=None,
-          noise_size=1, smoothing_size=None, threshold=None, invert=False,
-          percentile=64, topn=None, preprocess=True, max_iterations=10,
-          filter_before=None, filter_after=None, characterize=True,
-          engine='auto', output=None, meta=None, threads=None,
-          report_hook=None):
+def batch(frames, diameter, output=None, meta=None, processes=1,
+          after_locate=None, **kwargs):
     """Locate Gaussian-like blobs of some approximate size in a set of images.
 
     Preprocess the image by performing a band pass and a threshold.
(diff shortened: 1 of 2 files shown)


lagru:
The test failures seem unrelated to this PR. Something in `trackpy.tests.test_legacy_linking` seems to be the culprit but only for Python 3.6.

caspervdw:
Thanks for this contribution @lagru and sorry for taking so long to review. In general this is an absolutely great addition, multiprocessing has been on our list for some time and I am already happy to merge this as is. But now I am looking at it, I think it can be even faster if we go through some details. I will leave some line notes.

trackpy/feature.py
```diff
@@ -571,27 +582,51 @@ def batch(frames, diameter, minmass=100, maxsize=None, separation=None,
...
+        percentile=percentile, topn=topn, preprocess=preprocess,
+        max_iterations=max_iterations, filter_before=filter_before,
+        filter_after=filter_after, characterize=characterize, engine=engine
+    )
+    # Choose threaded or normal map function
+    if threads == 0:
+        pool = None
+        map_func = map
+    else:
+        pool = ThreadPool(processes=threads)
```
caspervdw: Is there a reasoning behind why you chose `ThreadPool`? I switched it to `Pool` and got much better performance (7 sec for `Pool` instead of 14 sec for `ThreadPool`, 16 sec with no threading at all)

nkeim:
This is great! I like that this approach uses the standard library. It seems that @caspervdw and I were looking at this at the same time.

Functions like `batch()` already have so many kwargs that we need to be careful about adding more. Could the job of `report_hook` be done via calls to `logging`? Or, is there a more general way we could build that into trackpy? (See e.g. the `quiet()` and `handle_logging()` functions in `trackpy.utils`)

Re: test failures see #501. That issue is still in search of a hero :)

trackpy/feature.py
```diff
@@ -571,27 +582,51 @@ def batch(frames, diameter, minmass=100, maxsize=None, separation=None,
...
+        all_features = []
+        for i, features in enumerate(map_func(curried_locate, frames)):
+            image = frames[i]
+            if hasattr(image, 'frame_no') and image.frame_no is not None:
+                frame_no = image.frame_no
+                # If this works, locate created a 'frame' column.
+            else:
+                frame_no = i
+                features['frame'] = i  # just counting iterations
+            report_hook(frame_no, len(features))
```
caspervdw: I would go for passing `features` alltogether and doing the `len` inside the `report_hook`. It is in the same process anyway, so we don't pay any price for doing that.

trackpy/feature.py
```diff
@@ -521,6 +524,14 @@ def batch(frames, diameter, minmass=100, maxsize=None, separation=None,
         If specified, information relevant to reproducing this batch is saved
         as a YAML file, a plain-text machine- and human-readable format.
         By default, this is None, and no file is saved.
+    threads : integer, optional
+        If 0, no threading is used. An integer greater 0 is treated as the
+        number of threads used to process frames in parallel. If `threads` is
+        None then the number returned by ``os.cpu_count()`` is used.
```
caspervdw: FYI I tested, seems to be only slightly faster than `os.cpu_count() - 1`. But this seems the most sensible default.

trackpy/feature.py
lagru: No, there is no specific reason. It was just an approach that worked for another use case in another project and I implicitly thought that threading would be enough to take advantage of NumPy's and Numba's GIL-releasing features while still being lightweight. 

My understanding is that multiprocessing increases the memory footprint and has a larger overhead compared to threading? But a speedup of 2 is certainly worth it in my opinion.

trackpy/feature.py
lagru: Yeah, I just mirrored the behavior of `Pool`'s input argument `processes`.

trackpy/feature.py
lagru: Sounds like a good idea.

lagru:
@nkeim  Thanks for you feedback.

> Functions like batch() already have so many kwargs that we need to be careful about adding more.

Why not collect all arguments that are passed through to `locate` in an argument `locate_kws` or  as `**kwargs` (not as explicit but wouldn't break the API)? I mean `batch` is nothing but a convenient wrapper for `locate` anyway and this would even follow the DRY principle concerning the docstring of `batch` which repeats the one from `locate`.

> Could the job of report_hook be done via calls to logging? 

Yeah it could but would be the inferior solution in my opinion. While wrapping trackpy for a command line tool I was a bit annoyed that I couldn't access the progress of `batch` (a really long running function) in a non-hacky way. Using the logging module wasn't enough, because I wanted to control the output in a more sophisticated way, in this case a progress bar. Besides the logging approach in trackpy only allows to configure logging for the whole library (at least that's what I think; I honestly find the logging module a little complex and difficult to grasp). So the best solution I could think of was to wrap `locate` in my own version of `batch` with a `report_hook`. Then I added threading and finally arrived on this PR. 

I think, especially for long running functions, there should be a way to overwrite the default behavior in a well defined way. With this approach the user is free to suppress all logging, if unwanted, but can still access the progress of long running functions if needed. 

Another advantage in this case might be, that `report_hook` could be used (abused?) for more complex tasks right after feature detection. But I can't really think of any right now. :smile:

> Or, is there a more general way we could build that into trackpy? (See e.g. the quiet() and handle_logging() functions in trackpy.utils)

I didn't even know those existed. It might be worthwhile to figure out whether these or `report_hook` enable the best solution. And I agree, library wide consistency is the more important point.

caspervdw:
@lagru @nkeim  Good point about the kwargs. In my opinion, the best would be

 - using `**kwargs` for all `locate` kwargs (as proposed by @lagru)
 - keeping `report_hook`, but renaming it to `after_locate` for library consistency. We have a similiar argument in the function `find_link`. See here for the docstring: https://github.com/soft-matter/trackpy/blob/b459975d8351b9b00d8cad57609c64fce5f24f6e/trackpy/linking/find_link.py#L79

trackpy/feature.py
caspervdw: A factor 2 is absolutely worth it. Memory consumption is not an issue in this case.

trackpy/feature.py
caspervdw: I was thinking about this. For me, `threads=None` would logically be "no threading", and `threads=0` could be the `os.cpu_count()`.

In either case, please keep the default behaviour "no threading". We do not want to surprise our users too much on a minor release (`v0.5` is not coming soon).

trackpy/feature.py
lagru: I tested this myself and there was no noticeable difference between the threading and multiprocessing version. So I think this depends on the setup and processed data itself.  For now I went with multiprocessing.

trackpy/feature.py
```diff
@@ -532,71 +506,97 @@ def batch(frames, diameter, minmass=100, maxsize=None, separation=None,
...
+            image = frames[i]
+            if hasattr(image, 'frame_no') and image.frame_no is not None:
+                frame_no = image.frame_no
+                # If this works, locate created a 'frame' column.
+            else:
+                frame_no = i
+                features['frame'] = i  # just counting iterations
+            features = after_locate(frame_no, features)
+
+            logger.info("Frame %d: %d features", frame_no, len(features))
```
caspervdw: I intended this to be in the default `after_locate`. What do you think @lagru and @nkeim?

trackpy/feature.py
```diff
@@ -532,71 +506,97 @@ def batch(frames, diameter, minmass=100, maxsize=None, separation=None,
...
+        try:
+            source = frames.filename
+        except AttributeError:
+            source = None
+
+        meta_info = dict(
+            timestamp=pd.datetime.utcnow().strftime('%Y-%m-%d-%H%M%S'),
+            trackpy_version=trackpy.__version__,
+            source=source,
+            # Use default values of `locate`...
```
caspervdw: you missed `diameter` here

trackpy/feature.py
```diff
@@ -532,71 +506,97 @@ def batch(frames, diameter, minmass=100, maxsize=None, separation=None,
...
+            minmass=None, maxsize=None, separation=None, noise_size=1,
+            smoothing_size=None, threshold=None, invert=False, percentile=64,
+            topn=None, preprocess=True, max_iterations=10, filter_before=None,
+            filter_after=None, characterize=True, engine='auto'
```
caspervdw: This is risky, we are likely going to forget these if we would ever change the defaults.
As the defaults are determined by the trackpy version, these can be dropped in my opinion.
I would go for timestamp, trackpy_version, source, diameter + kwargs.

What do you think @nkeim

caspervdw:
@lagru  Thanks for this revision, it is looking great now. I left some minor comments on your changes.
Only to do is unittesting, but let's raise a different issue for that.

trackpy/feature.py
lagru: No, I didn't. `diameter` is added to `kwargs` further up (line 520).

trackpy/feature.py
lagru: I didn't like this either but wanted to keep as much as possible of the old behavior. I'd prefer 

> timestamp, trackpy_version, source, diameter + kwargs

as well.

trackpy/feature.py
caspervdw: Right, missed that sorry!

trackpy/feature.py
lagru: That was my original intention as well. But I think this approach is actually more flexible and has the least surprises. From the user perspective: its not intuitive that if I supply my own post-processing function suddenly unrelated behavior (logging) is changed. At least that's my gut feeling. This way, logging and post-processing can be configured independently.

Otherwise the docstring should explicitly mention that providing `after_locate` overwrites the logging functionality.

trackpy/feature.py
```diff
@@ -521,6 +481,20 @@ def batch(frames, diameter, minmass=100, maxsize=None, separation=None,
         If specified, information relevant to reproducing this batch is saved
         as a YAML file, a plain-text machine- and human-readable format.
         By default, this is None, and no file is saved.
+    processes : integer or "auto", optional
+        The number of processes to use in parallel. If <= 1, multiprocessing is
+        disabled. If "auto", the number returned by `os.cpu_count()`` is used.
```
lagru: @caspervdw Do you agree with the way this is handled now?

trackpy/feature.py
caspervdw: That makes sense to me. Users could always control the logging through the logging module.

nkeim:
Sorry I have been absent from the discussion. I will try to catch up, but first: #503 has fixed the Travis errors. It should be pretty straightforward to rebase this PR onto the new `master` branch before merging.

lagru:
I'm sorry for taking so long but I didn't have access to my notebook.

trackpy/feature.py
nkeim: I agree as well—I like what you propose, @caspervdw . We should try to keep trackpy low-maintenance, and wrong metadata is much worse than missing metadata.

nkeim:
I am very excited about this PR!! Thank you @lagru .

lagru:
@caspervdw Do you agree with the current state?

caspervdw:
Thanks a lot @lagru for this contribution. @nkeim and @danielballan  do we need a minor release for this and the pandas fixes?

lagru:
You're welcome!

nkeim:
Ah! @caspervdw I just saw your question. I think a minor release would be OK but it is not urgent. My impression is that using the latest Pandas with v0.4.1 could cause some odd quirks but no errors or bad results.

I think that the tests shoud definitely cover this code path before we release. I created #506 .

PR #500, opened 2018-05-29
DOC: Update Keim funding info

nkeim:
This just updates the "Support" section of the intro page to acknowledge current funding for my trackpy work. Also, to reduce the US-centeredness, all affiliations now specify the country.

Because I am closing out a reporting period, It would be great to merge this week if possible.


nkeim:
The "bleeding edge" environment now runs tests on Pandas 0.23.0, which apparently can change the order of DataFrame columns in a new way, resulting in many failed tests. Plus there are a ton of new warnings. So this PR is safe to merge (it doesn't change any code, after all), but we should work on the Pandas issues very soon.

Issue #501, opened 2018-05-29
TST/BLD: Compatibility with Pandas 0.23.0


nkeim:
Most of our testing is still with Pandas 0.18.0. The latest release raises many, many warnings on both Python 2.7 and 3.6, but on Python 3.6 it also causes many tests to fail. The errors all have the form

```
======================================================================
FAIL: test_two_nearby_steppers (trackpy.tests.test_legacy_linking.TestKDTreeWithNumbaLinkDiag)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/build/nkeim/trackpy/trackpy/tests/test_legacy_linking.py", line 394, in test_two_nearby_steppers
    assert_frame_equal(actual_iter, expected)
  File "/home/travis/mc/envs/testenv/lib/python3.6/site-packages/pandas/util/testing.py", line 1323, in assert_frame_equal
    obj='{obj}.columns'.format(obj=obj))
  File "/home/travis/mc/envs/testenv/lib/python3.6/site-packages/pandas/util/testing.py", line 843, in assert_index_equal
    obj=obj, lobj=left, robj=right)
  File "pandas/_libs/testing.pyx", line 59, in pandas._libs.testing.assert_almost_equal
  File "pandas/_libs/testing.pyx", line 173, in pandas._libs.testing.assert_almost_equal
  File "/home/travis/mc/envs/testenv/lib/python3.6/site-packages/pandas/util/testing.py", line 1018, in raise_assert_detail
    raise AssertionError(msg)
AssertionError: DataFrame.columns are different
DataFrame.columns values are different (75.0 %)
[left]:  Index(['frame', 'x', 'y', 'particle'], dtype='object')
[right]: Index(['x', 'y', 'frame', 'particle'], dtype='object')
```

caspervdw:
I will try to fix the failures, warnings are for later

Resonanz:
I am having some success with Pandas 0.23.4.  Version 0.24.2 was a show-stopper for me.
```
pip uninstall pandas
pip install pandas==0.23.4
```

PR #503, opened 2018-06-03
Pandas 0.23 compat in legacy linking tests

caspervdw:
Attempts to fix the test failures in issue #501

### DIFF
--- a/trackpy/tests/test_legacy_linking.py
+++ b/trackpy/tests/test_legacy_linking.py
@@ -10,15 +10,13 @@ from numpy.testing import assert_array_equal
 import pandas as pd
 from pandas import DataFrame
 import nose
-from pandas.util.testing import assert_frame_equal, assert_series_equal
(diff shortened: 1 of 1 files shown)


caspervdw:
@nkeim See last commit, I removed the courtesy copy of the DataFrame if the provided DataFrame is a view and copy_features = False. In the legacy linking.

nkeim:
:+1: Since the legacy code is deprecated anyhow this makes perfect sense.

@caspervdw Should I take one last look and then merge?

caspervdw:
Yes ready to merge @nkeim

nkeim:
Thank you @caspervdw !!

It looks like, at worst, this issue would result in some warnings and weird column orders for people using v0.4.1 with the latest Anaconda. So I think we can release v0.4.2 at our leisure. Let me know if you think I'm mistaken!

PR #507, opened 2018-07-02
PERF/WIP: Parallel subnet solving

nkeim:
Inspired by #499 , this uses the `nogil` option for numba so that the subnet solver can be multithreaded. Threading lets us avoid worrying very much about shared mutable state in the linker, and also has lower overhead. The current implementation uses the `concurrent.futures` module so it is not compatible with older Python versions, and that's why some of the tests fail.

I have confirmed that this lets linking use multiple CPUs. However I have not confirmed that this is a significant speedup for common workloads. So I invite interested users to try it out!! At some point I will try this out on my own data, which has lots of subnets.

Obviously this is also missing tests, and an API to control the new feature. Right now multithreading is always enabled.

### DIFF
--- a/trackpy/linking/linking.py
+++ b/trackpy/linking/linking.py
@@ -5,6 +5,8 @@ from six.moves import range, zip
 import warnings
 import logging
 import itertools, functools
+from concurrent import futures
+from multiprocessing import cpu_count
 
 import numpy as np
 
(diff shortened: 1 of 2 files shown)


apiszcz:
Possible data point, I have a small tracking reference task that produces
about 7,000 trajectories over 2,362 frames.

0.4.1, 49s

0.4.1 with PERF/WIP patch applied, 53s

nkeim:
@apiszcz Thanks! So for at least some workloads, the additional overhead is not worthwhile.

One option would be to send a subnet computation to the threads only if it contains many particles. That could be done via a shared queue.

apiszcz:
I did notice 86% CPU load for existing method vs. 50% with patch. However
more tests are needed.

speedymcs:
Is this still being considered? Yesterday I ran into a problem with large subnetworks and thought that using multiple CPUs might be smart. For my test coordinates with about 2700 trajectories and 480 steps, the current trackpy implementation is much faster than this code here. However, my 12 cores report only about 10-12% usage each.

nkeim:
Hi @speedymcs ! I'm really glad you tried this patch. I'm not actively working on this right now, in part because I don't have obvious ideas about why it is slower or how to speed it up. I guess it is more of a proof-of-concept (that subnets can be solved in parallel, not that doing so is faster!).

If you have slow, large subnets, right now the best things to try (besides reducing `search_range` and filtering out spurious features) are adaptive linking and prediction.

speedymcs:
@nkeim Okay thanks! In the past I used the multiprocessing package for my own (crude) tracking implementation. It spread the CPU load very well but there were stability issues. Anyway I'll look into adaptive linking next. 

I tried to find out where in the image the oversize error occurs and recently asked a [question](https://stackoverflow.com/questions/57415419/how-to-find-out-where-in-the-image-a-subnet-oversize-exception-occurs) on Stack Overflow about that--excuse the shameless plug :)

tacaswell:
I responded on SO, but very short version is you should find a way to not get large sub-nets, they will lead to both long run times and unreliable linking.

PR #512, opened 2018-07-26
Update indexes of DataFrame when dropping coordinates

### LINKED ISSUE #511: DataFrame returned by locate() sometimes has NaN values
NaN values can appear in the ep_x, ep_y, x, y... columns. This seems to happen when coordinates are dropped in locate(). The concatenation

`refined_coords = pd.concat([refined_coords, ep], axis=1)`

leads to a `refined_coords` DataFrame instance with unexpected increased number of rows, because of index mismatches in the old `refined_coords` (that has missing indexes) and `ep` (that does not have missing indexes). When increasing the number of rows, the "holes" are filled with NaN.

This is a bug, but is also related to some of the concerns raised in #498. I will submit a PR soon.

bruot:
Fixes #511.

### DIFF
--- a/trackpy/feature.py
+++ b/trackpy/feature.py
@@ -400,7 +400,7 @@ def locate(raw_image, diameter, minmass=None, maxsize=None, separation=None,
     if np.all(np.greater(separation, 0)):
         to_drop = where_close(refined_coords[pos_columns], separation,
                               refined_coords['mass'])
-        refined_coords.drop(to_drop, axis=0, inplace=True)
+        refined_coords.drop(to_drop, axis=0)
 
     # mass and signal values has to be corrected due to the rescaling
     # raw_mass was obtained from raw image; size and ecc are scale-independent


bruot:
I saw two ways to fix this: Either keep indexes numbered 0, 1, 2... without gaps in `refined_coords`, or change indexes of `ep` to the same as `refined_coords` . Since the second option also leaks the strangely numbered indexes to the output of  `locate()`, I have chosen the first solution for cosmetic reasons.

caspervdw:
Hi @bruot, thanks for your contribution! It is always great to see others taking the effort of making a PR.
I had a look at your edit, and this may fix your issue but it does break some other things. Because you left out the `inplace=True`, the `pd.drop` is effectively not done. And we want to do that to get rid of duplicates.

I guess the easy solution would be initializing the `ep` dataframe with an index that equals the `refined_coords` dataframe. Over here:

https://github.com/soft-matter/trackpy/pull/512/files#diff-8caca4dcbf45ed9519ea2c65b2c391e6L447

Adding `index=refined_coords.index` would probably solve the issue

bruot:
Thanks for being reactive.

I think it is better to fix indexes as soon as possible issues can appear, i.e. when coordinates are dropped. This keeps the numbering in the indexes clean, and columns that are first constructed as numpy arrays can then be added without trouble. At the moment, only `ep` or `ep_*` are added, so it is easy to fix at `ep = `, but if more columns are added in the future, it could become tedious (and easy to forget) to take care of indexing for each new column.

See the new PR #514 that replaces this PR. My understanding of `locate()` is that it acts always on a single frame. If I am wrong, or if elsewhere in the code multidimensional DataFrames are used, it may be necessary to also add a `level=X` option to `reset_index`.

PR #519, opened 2018-09-15
FIX Do not calculate noise with insufficient pixels

krrk:
`uncertainty.measure_noise` returns 0 for noise if there is only one pixel in the background region. This leads to 0 for `ep` which I think is rather strange and unexpected. My change would instead make the result `NaN` which I think is more understandable to the end user.

### DIFF
--- a/doc/releases/v0.4.txt
+++ b/doc/releases/v0.4.txt
@@ -4,8 +4,8 @@ Development version
 API Changes
 ~~~~~~~~~~~
 
-- "trackpy.locate" and associated functions now report "signal" (and noise) to be
-  NaN when there is only one usable background pixel to sample. (:issue:`519`)
+- "trackpy.locate" and associated functions now report "ep" to be NaN when
+  there is only one usable background pixel to sample. (:issue:`519`)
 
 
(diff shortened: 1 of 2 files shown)


nkeim:
Thanks for the pull request!! This looks good to me. Since NaN is already a possible return value (when there are zero usable background pixels) this will probably not break any existing code. I'm going to give @caspervdw a couple more days if he has time to weigh in, and then I'd like to merge this.

nkeim:
@krrk I added an entry to the changelog. Feel free to edit if you see an inaccuracy.

caspervdw:
@nkeim Looks good to me. I kicked Travis, one build had a hiccup. After that has passed: good to merge.

krrk:
@nkeim I revised the documentation slightly. I don't think this affects the `signal` calculation which is just the maximum of the neighborhood. It only affects the `ep` column of the DataFrame.

I also updated my change slightly to store the number of pixels in the background so it doesn't have to be calculated twice.

nkeim:
Great! Thank you @krrk and @caspervdw !

PR #527, opened 2019-02-12
New locate function for colloidal particles in brightfield images

rbnvrw:
Hi all,

I added a new locate function (`locate_brightfield_ring`) which we have used in the following paper to locate micron-sized colloidal particles in brightfield movies: https://doi.org/10.1039/C8SM01661E (also on ArXiV: https://arxiv.org/abs/1807.07354

The tracking procedure is explained in the paper (see also figure in the SI) and is in my experience more reliable than the standard TrackPy routines for these particles, because they are not Gaussian features, contrary to for example confocal images (see images in SI for examples).

Specifically, it uses the normal `locate` function to get a rough estimate of the particle position, then it refines this position by fitting the ring of dark intensity around the particle, which in my experience always occurs in the images, regardless of specific focal plane. The normal locate/refinement fails because it assumes that the particle is a Gaussian blob.

Another feature is that you can supply initial positions instead of the 'guess' by the `locate` function, allowing for faster linking in the case of multiple frames and/or initial selection of a subset of particles to track by the user.

I have added unit tests that test the validity of the functions for generated images of multiple noise levels and intensity profiles. I also test that the normal `locate` function fails in these cases.

I hope this will be useful to someone and am happy to hear any feedback you might have.
I tried to adhere to the coding style of the package, but if you want me to change anything, please let me know. Also let me know if you want me to add something to the documentation.

Best,
Ruben


rbnvrw:
Note that some CI tests are passing while others are failing, I checked the errors but as far as I can see, they don't seem to be connected to my edits.

nkeim:
@rbnvrw This looks fantastic! I'm looking forward to trying it on my own images. Thank you for thinking to contribute this to trackpy, and for writing tests!!

The CI errors are from a new version of Pandas. I created issue #529 .

I can review the code, though it might be a couple weeks before I have time, so it would be great if someone else is interested!

In terms of documentation, have you considered adding a brief tutorial to https://github.com/soft-matter/trackpy-examples ? That seems to be how a lot of users discover new features, especially because you can show sample images.

rbnvrw:
Thank you @nkeim for your feedback! I'm curious to see how it performs 'in the wild'. I mainly used the refinement code until now on particles where the rough initial position was selected by the user, but based on the unit tests I've added, it also seems to work when no initial positions are given beforehand.

I've added additional tests to check how it performs for touching particles, this is also passing on my computer at least.

Thank you for looking into the errors from the CI. When I've time for it, I'll add an example to trackpy-examples.

rbnvrw:
I have added a short tutorial notebook to the trackpy-examples repository: https://github.com/soft-matter/trackpy-examples/pull/50

caspervdw:
@rbnvrw  Great to see your code contributed to trackpy! I scanned through the code, and the tests look convincing so I am :+1: on merging.

Only one thing, might be good to mention our earlier work available at https://github.com/caspervdw/circletracking , I recognise the `unwrap_ellipse` code.

rbnvrw:
Hi Casper, thank you! Yes, most functions are based on our code in https://github.com/caspervdw/circletracking
I've added it in the docstring (see commit) do you think I should add it anywhere else?

rbnvrw:
I have added a `**kwargs` to the `refine` function so that you can specify some parameters that are used in the `_min_edge` function. This is useful if you want to use this code for tracking bright rings with dark interiors, for example fluorescent particles with the dye only on the surface.

rbnvrw:
The unit tests are failing because of issue #529 which is fixed in PR #538

rbnvrw:
@nkeim would it be possible to include this in milestone v0.4.2, if the CI errors are indeed fixed by PR #538?

caspervdw:
Thanks @rbnvrw for the contribution. Merging this

PR #566, opened 2019-09-06
Misc. fixes for v0.4.2

nkeim:
Mostly documentation fixes. See commit list. Fixes #540, fixes #542, fixes #522

Also fixes some new Pandas warnings. CI tests will reveal whether there are unintended consequences.

PR #569, opened 2019-09-10
API: Improve dependencies report

nkeim:
This corrects some of the package names in `diag.dependencies()` and adds trackpy as a "dependency" to aid in diagnostics. It also puts `dependencies()` into the root namespace, so it will eventually be easier to ask people to run it.

This was originally part of #566 but I'm making it a separate PR.

### DIFF
--- a/trackpy/diag.py
+++ b/trackpy/diag.py
@@ -31,7 +31,7 @@ def dependencies():
         optional dependency is not installed, None
     """
     packages = ['six', 'numpy', 'scipy', 'matplotlib', 'pandas',
-                'skimage', 'pyyaml', 'tables', 'numba', 'pyfftw']
+                'skimage', 'sklearn', 'pyyaml', 'tables', 'numba', 'pyfftw']
     result = OrderedDict()
 
     # trackpy itself comes first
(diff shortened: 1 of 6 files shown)


nkeim:
I changed my mind on this: adding `trackpy.dependencies()` means extra short-term confusion for a small long-term benefit. So it is staying as `trackpy.diag.dependencies()`. In the meantime I realized that it should also report the `sklearn` version.

This should be ready to merge.

trackpy/diag.py
```diff
@@ -30,8 +31,12 @@ def dependencies():
         optional dependency is not installed, None
     """
     packages = ['six', 'numpy', 'scipy', 'matplotlib', 'pandas',
-                'scikit-image', 'pyyaml', 'pytables', 'numba', 'pyfftw']
+                'skimage', 'sklearn', 'pyyaml', 'tables', 'numba', 'pyfftw']
```
caspervdw: I think pyfftw can be dropped now that we moved pyfftw to the `legacy_bandpass` function.

Also, the exception for pyfftw below can be cleaned up then

nkeim: Good call! I dropped pyfftw and added pims, since that seems to come up often when troubleshooting trackpy.

nkeim:
Thanks, @caspervdw !

Issue #570, opened 2019-09-17
DOC: Zenodo integration


nkeim:
Back in 2016, #375 switched our Zenodo versioning to a manual process, so that we could include the up-to-date DOI with each release, and fully customize the record. Thus the records for the last 2 releases have been owned by @caspervdw and did not use Zenodo's GitHub integration. This gives us lots of control but creates extra steps, and makes it harder for multiple maintainers to keep a continuous chain of versions.

GitHub versioning support on Zenodo has advanced since then and uses better metadata. (Alas, still no ORCiDs.) As before, Zenodo has a "concept DOI" that always resolves to the latest release, as well as a [search that gives a reverse-chronological list of all versions](https://zenodo.org/search?page=1&size=20&q=conceptrecid:%22591577%22&sort=-version&all_versions=True).

I propose
* Re-enable the Zenodo web hooks on the repo (@caspervdw may be the only one who can do this)
* Change the Zenodo badge to point to the [GitHub-linked records](https://zenodo.org/badge/latestdoi/4744355) (as it was before 2016). It will always resolve to the latest release.
* To the version table in `introduction.rst`, add a "v0.4.2 and later" entry that points to the [complete list on GitHub](https://zenodo.org/search?page=1&size=20&q=conceptrecid:%22591577%22&sort=-version&all_versions=True)

**The alternative** is that the person doing the release creates the Zenodo record, or gets @caspervdw to do it. But since Casper is busy, that leads to breaks in continuity and means that we continue to manually update the DOIs in `introduction.rst` with each release.

Any thoughts?

caspervdw:
Creating a Zenodo entry should definitely become possible for all contributors. 

If re-enabling the Zenodo hook would solve this issue and at the same time ease the release process, I agree that this should be done.

If you need any help in changing the current records (that are apparently owned by me) please let me know.

nkeim:
Thanks, @caspervdw ! If it's easy for you to re-enable the web hooks, please do. If it's hard, don't worry about it—I've already manually prepared a record for v0.4.2.

caspervdw:
OK, I had a look:
 - v0.2.4 (https://zenodo.org/record/12255) is a versioned record. My account says that this record is "managed by another user". Maybe @danielballan ?
 - v0.3.0 (https://zenodo.org/record/34028) is manged by me and is not a "versioned record"
[...]

PR #576, opened 2019-10-10
Zenodo v042

nkeim:
This adds the reserved Zenodo DOI for v0.4.2, and corrects a typo. It should be the last PR before the release.

I have verified that the html docs build correctly with this PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 trackpy
 =======
 
-[![build status](https://travis-ci.org/soft-matter/trackpy.png?branch=master)](https://travis-ci.org/soft-matter/trackpy) [![Build status](https://ci.appveyor.com/api/projects/status/bc5umcboh3elm8oh?svg=true)](https://ci.appveyor.com/project/caspervdw/trackpy) [![DOI](https://zenodo.org/badge/doi/10.5281/zenodo.1213240.svg)](http://dx.doi.org/10.5281/zenodo.1213240)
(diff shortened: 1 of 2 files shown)


nkeim:
Ready to merge! Then I will finish the release.

rbnvrw:
Awesome!

I can't see the record, so just checking: can you also add me with my orcid id? It's http://orcid.org/0000-0003-3925-5732

Thanks!

nkeim:
Hi! Sorry for the wait. That DOI is reserved (it cannot be assigned to another record), but it is not yet published. Unfortunately there is no way for me to share ownership of the record, as we learned in #570.

Once I have made a zip file with `python setup.py sdist` I can publish the DOI. Then we can check that the record is public and valid, before publishing on PyPI and conda-forge. So I think you should go ahead and merge. Thanks!

nkeim:
BTW, in case you have not seen it yet, I am following (and revising) the [trackpy release checklist](https://github.com/soft-matter/trackpy/wiki/Release-Checklist).

rbnvrw:
Ok, great! I've merged it.

caspervdw:
@nkeim   I have been tracking the versioned record at Zenodo, and still haven't been able to re-enable the GH integration... I'll let you know when that is done

In the meantime, I suggest just using https://zenodo.org/badge/latestdoi/4744355 . I can make new versions under that record.

caspervdw:
Ah @rbnvrw were almost at the same time. I don't think it is good to start adding more Zenodo records that are unrelated to the main versioned record.

To make this process much easier, I suggest only keeping the "versioned" DOI in the repo. This will be a constant. After you do the release @nkeim, just tell me then I'll make and upload new version manually to Zenodo.

I'll make a PR in a minute

PR #577, opened 2019-10-14
Refer to versioned Zenodo record

caspervdw:
Following up #576 
Sorry about this double work @nkeim @rbnvrw

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 trackpy
 =======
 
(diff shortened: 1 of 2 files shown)


nkeim:
Thanks @caspervdw ! I like your plan for this release; I was just reluctant to commit you to doing the work :). I will let you know when the .zip is available.

rbnvrw:
OK, great! So we'll just have one Zenodo record from now on that we can update when there is a new version?

nkeim:
> So we'll just have one Zenodo record from now on that we can update when there is a new version?

That will work but it relies on @caspervdw to help out with each release (thanks!). Zenodo already has a GitHub integration and we would like to switch back to that, perhaps for v0.5. See #570 .

PR #580, opened 2019-10-19
ENH: Handle non-list pos_columns in compute_drift()

### LINKED ISSUE #579: Bug report via email
I received the following via personal email (also addressed to @tacaswell) back in August. Copying here so it doesn't get lost:

Part 1:

I just wanted to flag up a bug with the latest version of trackpy. 
I was trying to overcome the error triggered by the latest version of pandas when computing the drift and so I downloaded and copied the latest trackpy version into the ~/anaconda3/lib/python3.7/site-packages/trackpy folder. When I get to the compute drift walkthrough I get the following error:
TypeError: can only concatenate str (not "list") to str

Can you guys help/advise? 

Many many thanks! 

Part 2:

I  think I found the mistake.
In the line (in motion.py)

 dx = f_diff.loc[mask, pos_columns + ['frame']].groupby('frame').mean()

It should be 

 dx = f_diff.loc[mask, list(pos_columns) + ['frame']].groupby('frame').mean()

instead. Now it works for me. 
Thanks.

nkeim:
Fixes #579 .

### DIFF
--- a/doc/releases/v0.4.txt
+++ b/doc/releases/v0.4.txt
@@ -1,4 +1,13 @@
 
+v0.4.3 (in development)
+-----------------------
+
+Enhancements
+~~~~~~~~~~~~
+
+- compute_drift() now accepts list-like pos_columns (e.g. tuples). (:issue:`579`)
+
+
 v0.4.2
 ------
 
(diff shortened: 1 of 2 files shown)


nkeim:
Thanks @rbnvrw !

PR #585, opened 2019-10-26
Update deprecated import to collections.abc.Hashable

lagru:
Importing ABCs from `collections` instead of from `collections.abc` is deprecated and will stop working in Python 3.8. Therefore this will block trackpy from working with the new Python version.

Also im curious. Do you expect any other hiccups when using trackpy with 3.8? It looks like it should work out of the box seeing that the package is shared as a source distribution on PyPI.


tacaswell:
Thanks @lagru !  I would not expect any issues, but you can never be sure until you try....

PR #586, opened 2019-12-17
Accelerate linking by avoiding DataFrame.groupby.

anntzer:
On a personal dataset, this patch accelerates trackpy.link_df ~3-fold.
(I haven't managed to run the vbench benchmarks locally -- Py2 environments are too far away for me.)

Edit: This bumps the oldest supported numpy from 1.8.2 (released Aug. 4 2014) to 1.9.0 (released Sept. 7 2014, added return_counts), and thus also matplotlib from 1.3 to 1.4, numba from 0.13 to 0.14, scikit-image from 0.9 to 0.10, scikit-learn from 0.14 to 0.15, scipy from 0.12 to 0.14, and pandas from 0.13 to 0.15 (all because of availability of conda packages, except pandas 0.15 for some obscure compat with pytables problem?).  Hopefully that's fine with the devs; I can do without but the code would be a tiny bit more complicated.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,9 +5,9 @@ matrix:
   include:
     # "Legacy" environments: oldest supported versions, without and with numba
     - python: "2.7"
-      env: DEPS="numpy=1.8.2 scipy=0.12.0 matplotlib=1.3 pillow=2.1 pandas=0.13.0 scikit-image=0.9 pytables scikit-learn=0.14.1 pyyaml" BUILD_DOCS=false
+      env: DEPS="numpy=1.9 scipy=0.14.0 matplotlib=1.4 pillow=2.1 pandas=0.15 scikit-image=0.10 pytables scikit-learn=0.15 pyyaml libgfortran=1" BUILD_DOCS=false
     - python: "2.7"
-      env: DEPS="numpy=1.8.2 scipy=0.12.0 matplotlib=1.3 pillow=2.1 pandas=0.13.0 scikit-image=0.9 numba=0.13.4 pytables scikit-learn=0.14.1 pyyaml" BUILD_DOCS=false
+      env: DEPS="numpy=1.9 scipy=0.14.0 matplotlib=1.4 pillow=2.1 pandas=0.15 scikit-image=0.10 numba=0.14 pytables scikit-learn=0.15 pyyaml libgfortran=1" BUILD_DOCS=false
     # "Typical" environments: versions 2-3 years old
     # Python 2, circa 2016
     - python: "2.7"
--- a/setup.py
+++ b/setup.py
@@ -32,8 +32,8 @@ setup_parameters = dict(
     author = "Trackpy Contributors",
     author_email = "daniel.b.allan@gmail.com",
     url = "https://github.com/soft-matter/trackpy",
-    install_requires = ['numpy>=1.8', 'scipy>=0.12', 'six>=1.8',
-	                'pandas>=0.13', 'pyyaml', 'matplotlib'],
+    install_requires = ['numpy>=1.9', 'scipy>=0.14', 'six>=1.8',
+                        'pandas>=0.15', 'pyyaml', 'matplotlib'],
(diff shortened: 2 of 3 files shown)


nkeim:
@anntzer This is very exciting! I don't have time to review this in the next few weeks (moving to a different university), but perhaps someone else can (@rbnvrw ?) 

The `vbench` stuff should probably be removed. We have a working benchmark suite for `asv`, but the repo is private because it uses LFS so trackpy can run on a significant amount of real data. The instructions are at https://github.com/soft-matter/trackpy-bench/wiki

anntzer:
I can't access the wiki.  I guess it may be private as well?

nkeim:
Yes. @anntzer, I just gave you read access to the trackpy-bench repo.

anntzer:
thanks

anntzer:
A run of the benchmark suite shows that this PR accelerates WaterLinkingSuite.time_{python,numba} from ~130ms to ~107ms and has negligible effects on the other benchmarks.  I *guess* the difference in speedup may come from the fact that my datasets have very long movies (~20k frames) relative to yours, and the PR is really just getting rid of the overhead of splitting the dataset by frames.

anntzer:

setup.py
```diff
@@ -32,8 +32,8 @@
     author = "Trackpy Contributors",
     author_email = "daniel.b.allan@gmail.com",
     url = "https://github.com/soft-matter/trackpy",
-    install_requires = ['numpy>=1.8', 'scipy>=0.12', 'six>=1.8',
-	                'pandas>=0.13', 'pyyaml', 'matplotlib'],
+    install_requires = ['numpy>=1.9', 'scipy', 'six>=1.8',
+	                'pandas', 'pyyaml', 'matplotlib'],
```
nkeim: Did you remove the version requirements for scipy and pandas because `numpy>=1.9` is sufficient to pin those other packages? If not, I think they are useful, since some scientific users have really weird environments and installation methods.

.travis.yml
```diff
@@ -5,9 +5,9 @@ matrix:
   include:
     # "Legacy" environments: oldest supported versions, without and with numba
     - python: "2.7"
-      env: DEPS="numpy=1.8.2 scipy=0.12.0 matplotlib=1.3 pillow=2.1 pandas=0.13.0 scikit-image=0.9 pytables scikit-learn=0.14.1 pyyaml" BUILD_DOCS=false SKIP_SLOW=true
+      env: DEPS="numpy=1.9 scipy=0.14.0 matplotlib=1.4 pillow=2.1 pandas=0.15 scikit-image=0.10 pytables scikit-learn=0.15 pyyaml libgfortran=1" BUILD_DOCS=false SKIP_SLOW=true
```
nkeim: I am very happy with updating the "legacy" environment to late 2014 😄.

trackpy/linking/utils.py
```diff
@@ -38,23 +38,27 @@ def coords_from_df(df, pos_columns, t_column):
...
+    unique_times, time_counts = np.unique(times, return_counts=True)
+    pos_by_frame = np.split(pos, np.cumsum(time_counts)[:-1])
+
+    idx = 0
+    for time in range(unique_times[0], unique_times[-1] + 1):
+        if time == unique_times[idx]:
+            yield time, pos_by_frame[idx]
+            idx += 1
+        else:
+            yield time, np.empty((0, ndim))
```
nkeim: Nice! From my read through it looks like this is really equivalent in terms of results and data types, which is of course borne out by @caspervdw 's tests. Pandas helped trackpy avoid premature optimization, but it's good to see it replaced in one more algorithm.

setup.py
anntzer: I put back the version requirements to match the legacy env on travis, just to be safe.

nkeim:
Excellent. Thanks for doing the rebase! I'm so happy that trackpy continues to get faster. 🙏

PR #587, opened 2019-12-17
Fix for deprecation of np.asscalar in numpy 1.16.

anntzer:
Gets rid of a deprecationwarning in the test suite.


anntzer:

PR #588, opened 2019-12-17
Remove dependency of test suite on nose.

anntzer:
Nose is quite unlikely to be installed on a dev's machine these days.

The test suite can fairly easily be converted to plain unittest, except
for the "slow" attrs which would need pytest to be replicated.  I'm not
against making the test suite depend on pytest (which is quite common
nowadays), but I'm not really sure it's worth it either -- the slow
tests don't add *so much* runtime to the test suite either.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,26 +5,26 @@ matrix:
   include:
     # "Legacy" environments: oldest supported versions, without and with numba
     - python: "2.7"
-      env: DEPS="numpy=1.8.2 scipy=0.12.0 matplotlib=1.3 pillow=2.1 pandas=0.13.0 scikit-image=0.9 pytables scikit-learn=0.14.1 pyyaml" BUILD_DOCS=false SKIP_SLOW=true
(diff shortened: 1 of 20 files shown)


nkeim:
Thanks @anntzer ! Sorry about the accidental close/reopen. 

I'm going to hold off a bit on merging this until I/someone can look at the "slow" tests a little more closely and decide whether pytest is worth it. You are right that there does not seem to be a dramatic change in runtime.

Note that "nose" is still included in the default Anaconda install, which is probably why it got adopted in the first place.

tacaswell:
When we started trackpy `nose` was the standard and `pytest` was a twinkle in someone's eye, now `nose` has declared themselves obsolete and the community has mover more-or-less en-mass to pytest.

nkeim:
Thanks @anntzer for these PRs and for your patience!

Thanks @tacaswell for weighing in and merging the other PRs. My move is done and I am slowly coming back online! I personally use pytest and I didn't realize that nose was deprecated. TBH I don't think we will ever get around to skipping slow tests again, but it's a safe bet that CI servers aren't getting slower.

anntzer:
No hurries, thanks :)

PR #593, opened 2020-01-01
Fix deprecations in recent Pythons.

anntzer:
Iterable will be removed from collections in Py3.9; \d is an invalid
escape.


anntzer:

PR #596, opened 2020-02-12
Fix test suite with pandas 1.0.

anntzer:
pandas 1.0 broke backcompat in the format of their pickles, so the
various `.df` files in tests/data don't load properly anymore.  Delete
the unused ones and convert the rest to the more stable .npy format (as
struct arrays); this was done using
```
def convert(path):
    df = pd.read_pickle(path)
    np.save(path.replace(".df", ".npy"), df.view(df.dtype.fields, np.ndarray))
```
(TL;DR: this patch is necessary to run the test suite with the latest pandas.)


danielballan:
Thank you!

PR #597, opened 2020-02-12
PERF: Don't sort links before applying them.

anntzer:
This should not have any effects on the linking(?) given that the
sorting is by x/y coordinate, which is basically arbitrary.

This speeds up the simple example below by ~15%
```python
import time
import numpy as np
from pandas import DataFrame
import trackpy as tp; tp.ignore_logging()

n_particles = 50
n_frames = 10000
frame_size = 200
starting = np.random.random((n_particles, 2)) * frame_size
dxys = np.random.standard_normal((n_frames, n_particles, 2))
xys = starting + np.cumsum(dxys, axis=0)
frames = []
xs = []
ys = []
for f in range(n_frames):
    for p in range(n_particles):
        if np.random.random() < 0.5:  # Kill half of the localizations.
            continue
        frames.append(f)
        xs.append(xys[f, p, 0])
        ys.append(xys[f, p, 1])
df = DataFrame({"frame": frames, "x": xs, "y": ys})

__start = time.perf_counter()
tp.link_df(df, 3, memory=10)
print("elapsed", time.perf_counter() - __start)
```

On personal datasets the speedup goes up to 2x(!).

This PR goes on top of #596.

### DIFF
--- a/trackpy/linking/linking.py
+++ b/trackpy/linking/linking.py
@@ -307,13 +307,6 @@ def adaptive_link_wrap(source_set, dest_set, search_range, subnet_linker,
     return sn_spl, sn_dpl
 
 
-def _sort_key_spl_dpl(x):
-    if x[0] is not None:
-        return list(x[0].pos)
-    else:
-        return list(x[1].pos)
-
-
 class Linker(object):
     """Linker class that sequentially links ndarrays of coordinates together.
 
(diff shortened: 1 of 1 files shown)


anntzer:
Kindly bumping.

rbnvrw:
I think this is a good addition and I'll merge the PR. Thank you for the effort!

nkeim:
This is fantastic, @anntzer !! I am glad that this is merged —thanks @rbnvrw !

It seems like this will change the track numbering in some cases, even though nobody's scientific results will change. Is that true?

If so, we should be sure to explain clearly in the release notes, in case someone's code does break. (Or maybe it is time for v0.5?)

anntzer:
You are correct.  Let's keep the discussion on that other thread.

Issue #506, opened 2018-07-02
Unit tests for parallel batch


nkeim:
At least a smoke test is needed for #499 .

PR #602, opened 2020-03-18
Add optional Pool argument to locate_brightfield_ring to enable multiprocessing

rbnvrw:
**Add optional Pool argument to locate_brightfield_ring to enable multiprocessing**

- If not passed, nothing changes with respect to the "old" behaviour of the function.
- If `pool` is "truthy", a method `pool.map` is called for each of the particles that need to be refined

This allows you to do something like this:
```python
import trackpy as tp
from multiprocessing import Pool
with Pool() as pool:
    result = tp.locate_brightfield_ring(frame, diameter, pool=pool)
```

To speed up the refinement step.

### DIFF
--- a/trackpy/locate_functions/brightfield_ring.py
+++ b/trackpy/locate_functions/brightfield_ring.py
@@ -16,7 +16,7 @@ from ..feature import locate
 
 
 def locate_brightfield_ring(raw_image, diameter, separation=None,
-                            previous_coords=None, **kwargs):
+                            previous_coords=None, pool=None, **kwargs):
     """Locate particles imaged in brightfield mode of some approximate size in
     an image.
 
@@ -40,6 +40,9 @@ def locate_brightfield_ring(raw_image, diameter, separation=None,
     previous_coords : DataFrame([x, y, r])
         Optional previous particle positions from the preceding frame to use as
         starting point for the refinement instead of the intensity peaks.
+    pool : multiprocessing.Pool (optional)
+        If passed, use a pool for the refinement step (separate process for
+        each particle).
     kwargs:
         Passed to the refine function.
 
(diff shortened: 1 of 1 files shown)


nkeim:
Fabulous!

This is a different argument name and convention than [the ones used by `batch()`](http://soft-matter.github.io/trackpy/v0.4.2/generated/trackpy.batch.html#trackpy.batch). How hard would it be to make them work the same way? Could there be a `trackpy.utils._get_pool()` function that accepts an integer, `"auto"`, or a pool-like object and returns a pool?

rbnvrw:
Good point @nkeim to be honest I forgot about the `processes` keyword in `batch`. It makes sense to implement the same API. I'll make a `pool, map_func = trackpy.utils._get_pool(processes)` function that can be used in `batch` and in this function.

rbnvrw:
I accidentally pushed another commit to this PR, however I'd like to merge that one as well. It is a small improvement to the `refine_brightfield_ring` method I developed, where instead of the minimum value it uses the average of values lower than the 5th percentile for edge detection. I have tested it on my data and it performs a bit more robustly, especially for lower contrast images. Additionally, all unit tests I added for this function where synthetic data is tested are also passing.

nkeim:
The refactor looks great!! Sorry for the delay.

The algorithm tweak sounds good. It will need to be noted in `whatsnew.rst`. You can write that or just create an issue for v0.4.3.

One last thing: could you please add a test of this new feature to `test_brightfield_ring.py`? You could just make a copy of `test_multiple_simple_sparse()` that operates on a list of frames. This would also partially address #506 .

rbnvrw:
Good points @nkeim. I've added it to the doc file. The `processes='auto'` turns on multiprocessing by default, I have added one simple test case with `processes=0` to test the case where no multiprocessing is used. This one is also passing.

swissel:
OK, this is ready to merge! Thanks for this great enhancement, @rbnvrw . 

Your test code brought up one inconsistency I didn’t notice before: the default for `locate_brightfield_ring()` is `processes=‘auto’`, but the default for `batch()` is `processes=1`. I’m going to create an issue and see if anyone cares :).

nkeim:
This is ready to merge! Thanks for this enhancement, @rbnvrw .

One inconsistency I noticed is that the default for `locate_brightfield_ring()` is `processes=‘auto’`, while for `batch()` it’s `processes=1`. It seems like it should be the same for both. I’m going to create an issue and see if anyone cares.

Issue #601, opened 2020-03-11
Add note about tracks sorting to release notes


nkeim:
This is a follow-up to #597 . The release notes should make it very clear that the assigned track numbers may have changed when running trackpy with identical parameters on the same data, even though the tracks themselves should be identical. Of course, the release notes should also make it clear that the linking algorithm is now significantly faster!

danielballan:
Do you think it's worth bumping the minor version over this, releasing as 0.5.0? My general guideline for this is, "Would the results change in a way that isn't unambiguously fixing an error?"

apiszcz:
FYSA: I grabbed the latest trackpy from the repo and ran my standard linking task.
The new arrangement appears to be slower for my use case.

```
0.4.2  00:17:46 - 00:19:03  1m 17s
0.4.2+ 00:22:50 - 00:24:15 1m 25s
```

anntzer:
I can try having a look if you can provide your data.

nkeim:
I'm preparing release notes for this issue (about not sorting tracks in #597 ) and would like to resolve an ambiguity. Does this change mean that

> although particle numbers may be different than in the previous release (v0.4.2), as long as the input features are identical, particle numbers will be stable from run to run.

or, 

> particle numbers are now unstable from run to run, just as the ordering of `set` and `dict` objects is unstable.

If it's the latter, I'm a bit concerned about reproducibility. For instance, I might write down a trajectory number in my lab notebook as being noteworthy (or invalid). I could be confident that as long as I didn't change any particle tracking parameters and kept my Python environment the same, that identifier would be valid, even if I lost the trackpy output or gave my scripts to someone else to run.

Does that sound like a realistic scenario? I suppose that in all cases, the much more correct thing for me to do would be to write down "the trajectory beginning with the particle at (*x*, *y*) in frame *t*." Maybe we should have a convenience function to do that? 😉

nkeim:
Just to clarify: if we really do think that not sorting tracks presents too high a barrier to reproducibility, all we have to do is make it optional. I still think it's a good addition to trackpy!

anntzer:
No guarantees, but my guess is that you'll still get stable results by setting PYTHONHASHSEED to enforce dict ordering.

nkeim:
[...]

PR #606, opened 2020-04-28
Parallel batch() by default

nkeim:
This also spiffs up the release notes and fixes a bug in parallel `batch()`. It does **not** yet relabel this release as v0.5.0, which was sensibly proposed in #601. See the discussion there.

The fact that there was a bug lurking in parallel batch() + pims suggests that very few people have been using that feature in v0.4.2. Which means
1. this version should get a lot more testing before it's released, and
2. this change could make a lot of people very happy 🎉


nkeim:
This PR now indicates that the next release will be v0.5, which is appropriate both for the change to `batch()` and for the change to trajectory numbering discussed in #601 .

rbnvrw:
This looks great! I think the documentation is also getting much clearer now. Thanks :+1:

Issue #616, opened 2020-07-02
Drop official Python 2 support


nkeim:
Given that Python 2 is past its end-of-life, compatibility should not be a goal for trackpy v0.5. #615 will drop CI testing on Python 2, but we also need to at least do the following:

- [x] Remove Python 2 from `setup.py`; make sure the minimum versions for other dependencies equal the oldest versions in `.travis.yml`.
- [x] Prepare (but don't merge) an update to the conda recipe on https://github.com/conda-forge/trackpy-feedstock .
- [x] Update the installation instructions.

PR #619, opened 2020-07-11
Drop support for EOL Python 2

hugovk:
For https://github.com/soft-matter/trackpy/issues/616.

Remove redundant code and modernise the codebase to use newer syntax.


caspervdw:
Thanks @hugovk ! A bit late, but still, I'll merge

Issue #629, opened 2020-08-26
VisibleDeprecationWarning


ivanovmg:
Repro steps
------------

1. Clone repo, master branch
2. Run unit tests: ``python -m unittest trackpy/tests/*.py``
3. Observe multiple instances of deprecation warnings.

```
VisibleDeprecationWarning: Creating an ndarray from ragged nested sequences (which is a list-or-tuple
of lists-or-tuples-or ndarrays with different lengths or shapes) is deprecated.
If you meant to do this, you must specify 'dtype=object' when creating the ndarray
  r = np.sqrt(np.sum(np.array(coords)**2, axis=0))
```

System information
------------------

Windows 10

Packages installed:
```
$ python --version
Python 3.8.3

$ pip freeze
certifi==2020.6.20
cycler==0.10.0
kiwisolver==1.2.0
llvmlite==0.34.0
matplotlib==3.3.1
numba==0.51.0
numpy==1.19.1
pandas==1.1.0
Pillow==7.2.0
pyparsing==2.4.7
python-dateutil==2.8.1
pytz==2020.1
PyYAML==5.3.1
scipy==1.5.2
six==1.15.0
-e git+https://github.com/ivanovmg/trackpy.git@cdab60a9d9572bb1e214b69f680dfaf355c370ec#egg=trackpy
```

tacaswell:
This is a change in the behavior of in numpy (starting with np.1.19).

PR #630, opened 2020-08-26
Fix VisibleDeprecationWarning

ivanovmg:
Fixes https://github.com/soft-matter/trackpy/issues/629

### DIFF
--- a/trackpy/artificial.py
+++ b/trackpy/artificial.py
@@ -107,7 +107,7 @@ def draw_feature(image, position, size, max_value=None, feat_func='gauss',
         rect.append(slice(lower_bound, upper_bound))
         vectors.append(np.arange(lower_bound - c, upper_bound - c) / s)
     coords = np.meshgrid(*vectors, indexing='ij', sparse=True)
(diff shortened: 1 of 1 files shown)


tacaswell:
This seems to break on newer versions of scipy?

ivanovmg:
> 
> 
> This seems to break on newer versions of scipy?

Actually I get the same errors in tests locally on the master branch. So, looks like the problem with the build failing is not caused by the changes introduced.

tacaswell:
https://travis-ci.org/github/soft-matter/trackpy/builds/707634015 passed (about a month ago).

tacaswell:
I just added a weekly cron-job to re-run master, lets see what CI has to say for its self...

anntzer:
I can confirm this already test is broken as of master.  Actually looking at its contents it is unclear to me how it can have passed before, as it seems to check that the computed drift from a set of random walkers is *exactly* zero (it should be close to zero, certainly, but not *exactly* zero).

caspervdw:
The VisibleDeprecation warning is now solved as part of #641. I am trying to address other test failures as well.

PR #641, opened 2021-02-13
Fix Numpy 1.20 and scipy 1.5 deprecationwarnings

caspervdw:
Ref. https://numpy.org/devdocs/release/1.20.0-notes.html#using-the-aliases-of-builtin-types-like-np-int-is-deprecated

`np.int` is identical to the builtin `int`. I didn't want to change too much so that is what I replaced. Same story for `np.float` and `np.bool` (and a few other dtypes, but they did not occur in trackpy)

### DIFF
--- a/trackpy/tests/test_motion.py
+++ b/trackpy/tests/test_motion.py
@@ -21,9 +21,9 @@ def conformity(df):
     """ Organize toy data to look like real data. Be strict about dtypes:
     particle is a float and frame is an integer."""
     df['frame'] = df['frame'].astype(int)
-    df['particle'] = df['particle'].astype(np.float)
-    df['x'] = df['x'].astype(np.float)
-    df['y'] = df['y'].astype(np.float)
+    df['particle'] = df['particle'].astype(float)
+    df['x'] = df['x'].astype(float)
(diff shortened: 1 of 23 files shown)


trackpy/tests/test_motion.py
```diff
@@ -78,10 +78,6 @@ def test_no_drift(self):
         actual_rolling = tp.compute_drift(self.dead_still, smoothing=2)
         assert_frame_equal(actual_rolling, expected[['y', 'x']])
 
-        # Small random drift
```
caspervdw, line 81: @anntzer   I disabled this test. For some reason, `assert_frame_equal` was OK with comparing `1E-7` with `0.0` for ` pandas<1.1.0` Apparently `assert_frame_equal` was made more strict in pandas 1.1.0. I haven't been able to find this mentioned in the [release notes](https://pandas.pydata.org/pandas-docs/stable/whatsnew/v1.1.0.html#backwards-incompatible-api-changes).

caspervdw:
I am going to self-merge this, because there is just maintenance stuff in here.

nkeim:
Thanks @caspervdw !

Issue #647, opened 2021-02-23
Cannot find docs on reconnect_traj_patch.


loomcode:
Hello and thank you for the project!  It's been very helpful.  I'd like to run tracking in parallel and the **trackpy.link_partia**l documentation says: _The description for A dataset can be divided into several patches (time intervals) and linked separately with this function, for e.g. parallel processing. The results can then be stitched back together with reconnect_traj_patch()._  The issue is that the link to reconnect_traj_patch is broken and I can't find any documentation for how to use it.  I also can't find a reference to the function in my copy (version 0.4.2).  Any guidance would be great.  Thank you.

nkeim:
Thanks! You should be able to get more information (including the documentation text for `reconnect_traj_patch()` from the [source code](https://github.com/soft-matter/trackpy/blob/master/trackpy/linking/partial.py). I'll leave this open so that we fix the online documentation.

loomcode:
Awesome!  I'll try it out tomorrow.  Thanks!

jackyko1991:
This issue seems still not fixed after an year.

nkeim:
Thanks for noting this @jackyko1991 . The related PR #648 was made when our automated build system was in flux; now it just needs a rebase before we can merge it. I added it to the 0.6 milestone.

PR #648, opened 2021-02-24
API/DOC: Expose reconnect_traj_patch()

nkeim:
Addresses #647 . This should not be merged unless the generated docs can be checked immediately before or after. The fix should show up in the API Reference page and in the "See also" section of the `link_partial()` page.


nkeim:
Merged after confirming that I can build the docs build correctly

Issue #642, opened 2021-02-21
TestBrightfieldRing fails on all but Py3.9 NumPy1.19 envs


caspervdw:
For example: https://github.com/caspervdw/trackpy/runs/1946300540

On Py3.8 environment, only one error remains.

@rbnvrw  Do you have time to take a look?

Tracebacks all look like:

```
______________ TestLocateBrightfieldRing.test_cluster_more_noisy _______________
multiprocessing.pool.RemoteTraceback: 
"""
Traceback (most recent call last):
  File "/opt/hostedtoolcache/Python/3.8.7/x64/lib/python3.8/multiprocessing/pool.py", line 125, in worker
    result = (True, func(*args, **kwds))
  File "/home/runner/work/trackpy/trackpy/trackpy/locate_functions/brightfield_ring.py", line 158, in _get_refined_coords
    result = refine_brightfield_ring(image, radius, positions,
  File "/home/runner/work/trackpy/trackpy/trackpy/refine/brightfield_ring.py", line 55, in refine_brightfield_ring
    result = _refine_brightfield_ring(image, r, coords_df, rad_range=rad_range,
  File "/home/runner/work/trackpy/trackpy/trackpy/refine/brightfield_ring.py", line 122, in _refine_brightfield_ring
    r_dev = _min_edge(intensity, **kwargs)
  File "/home/runner/work/trackpy/trackpy/trackpy/refine/brightfield_ring.py", line 172, in _min_edge
    argmin = np.where(row < min_val)[0]
FloatingPointError: invalid value encountered in less
"""

The above exception was the direct cause of the following exception:

self = <test_brightfield_ring.TestLocateBrightfieldRing testMethod=test_cluster_more_noisy>

    def test_cluster_more_noisy(self):
>       actual, expected = artificial_cluster(self.image_size,
                                              self.n_feat_cluster,
                                              self.radius, noise_level=51,
                                              separation=self.cluster_sep)

trackpy/tests/locate/test_brightfield_ring.py:190: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
trackpy/tests/locate/test_brightfield_ring.py:58: in artificial_cluster
    return draw_artificial_image(shape, pos, radius, noise_level, dip,
trackpy/tests/locate/test_brightfield_ring.py:32: in draw_artificial_image
[...]

PR #654, opened 2021-03-31
Fix for issue #642: refine/brightfield_ring.py _min_edge(), refactor and add `min_percentile` kwarg

rbnvrw:
This commit does a few things:

1. Adds a `min_percentile` kwarg to refine/brightfield_ring.py `_min_edge()`. Before, I just used `np.nanmin()` there, this was changed to `np.nanpercentile(arr, 5)` because that worked better for experimental data. In the tests however, the edge is very sharp and therefore it would sometimes fail. With this new kwarg, users can define the threshold based on their use case.
2. Refactor refine/brightfield_ring.py `_min_edge()` to get rid of the for loop and only use array operations.
3. Use `np.errstate(invalid='ignore')` to allow comparison to np.nan in refine/brightfield_ring.py `_min_edge()`, these are filtered out later. Also suppress warnings on all `NaN` slices there, this is OK because those are rows for which a minimum could not be found, these are handled later in the same function.

This fixes the tests that failed in issue #642, that were apparently caused by a different default value for the numpy error state (related to point 3 above).

### DIFF
--- a/trackpy/refine/brightfield_ring.py
+++ b/trackpy/refine/brightfield_ring.py
@@ -4,6 +4,7 @@ bright interior part. Based on https://github.com/caspervdw/circletracking
 """
 import numpy as np
 import pandas as pd
+import warnings
 
 from scipy.ndimage import map_coordinates
 from scipy import stats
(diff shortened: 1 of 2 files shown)


caspervdw (approved):

rbnvrw:
Hmm OK weird.. I have no experience with GitHub actions, so I'm not sure where it could be coming from.. it is only for one (Pip / Python 3.7) so maybe there is something wrong in the environment config there?

caspervdw:
My first guess would be an incompatibility between specific versions of some compiled packages and/or the python interpreter. But we have many of them, it will be hard to find it. Best chance is trying to reproduce the exact environment (so on a linux machine).

Issue #655, opened 2021-03-31
Fix segfault after pytest on Py3.7, pip environment


caspervdw:
See: https://github.com/soft-matter/trackpy/runs/2236528914 and discussion in #654 

The command `pytest trackpy` ends with:

```
=========== 929 passed, 17 skipped, 7 warnings in 353.98s (0:05:53) ============
Fatal Python error: Segmentation fault

Current thread 0x00007f14cdbda740 (most recent call first):
/home/runner/work/_temp/1ef82e28-b658-4675-8755-17c6857f00fb.sh: line 1:  2060 Segmentation fault      (core dumped) pytest trackpy
Error: Process completed with exit code 139.
```

So the segfault happens probably when the interpreter shuts down.

caspervdw:
I was able to reproduce locally with the following dockerfile:
```
FROM python:3.7

RUN apt-get update && apt-get install -y libhdf5-dev --no-install-recommends

RUN pip install --disable-pip-version-check --upgrade pip && pip install --upgrade wheel 
RUN pip install --no-build-isolation numpy==1.16.*
RUN pip install --no-build-isolation pytest scipy==1.3.* matplotlib==2.2.* pillow==5.3.* pandas==0.23.* scikit-image==0.14.* tables==3.5.* scikit-learn==0.20.* pyyaml==3.13 numba==0.39.* llvmlite==0.24.*

COPY . /code

WORKDIR /code
```

The problems are caused by numba. I started to gradually bump numpy and they dissappeared with numba==0.45.
I think this has to do with https://github.com/numba/numba/issues/4323 ; so an incompatibilty between numba <0.45 and python 3.7.4 (we test with 3.7.10, but I guess the incompatibility persists accross these micro versions).

Solution is: test with numba 0.45 in our py37 env.

rbnvrw:
Sounds good. Cool that you found it so fast!

PR #656, opened 2021-03-31
Change minimum versions and remove compatibility fixes

### LINKED ISSUE #645: Update trackpy's version constraints
Our current setup.py lists the following version constraints:

```
    install_requires = ['numpy>=1.9', 'scipy>=0.14',
                        'pandas>=0.15', 'pyyaml', 'matplotlib'],
    python_requires=">=3.5",
```

On the python timescale, these versions are ancient: end 2014.
- Python 3.5 has reached its end of life as of Sept 2020 (https://www.python.org/dev/peps/pep-0478/)
- Currently, the earliest supported numpy is 1.17 (https://numpy.org/neps/nep-0029-deprecation_policy.html)

 In https://github.com/soft-matter/trackpy/blob/master/.github/workflows/test-pip.yml I have been much more conservative than these official recommendations (Python 3.5, Numpy 1.13). I think that is a good thing, as trackpy is used by researchers which tend to be less on the forefront with upgrading their scripts. I am sure there also plenty Py2.7 scripts around.

We do have to draw a line somewhere. What do you think @tacaswell @danielballan @nkeim ? Just bump the dependencies to the first ones I happened to test with?

And maybe we should do something with the `extras_require` to list optional dependencies (together with their version constraints) explicitly (https://setuptools.readthedocs.io/en/latest/userguide/dependency_management.html#optional-dependencies), like numba and (py)tables.

caspervdw:
Closes #645 and #655


rbnvrw (commented):
It looks good, but the Travis test is failing. It seems to me that we need to bump the Python version from 3.5 to (at least) 3.6 in https://github.com/soft-matter/trackpy/blob/master/.travis.yml as well to fix this.

caspervdw:
I am afraid that the doc build might fall apart if I touch it... Maybe better to leave Py3.5 in after all for this release, and only fix #655  in v0.5? What do you think?

rbnvrw:
I've tried building the docs locally on Python 3.8.6 and that worked without problem, so I'm not sure why you think they'd break?
While building the docs, I found that we don't have a toctree that includes the full API, so I've added it in this PR: #658 Probably not crucial to have, but maybe a good idea for discoverability of functions that are not included in the main API document.

caspervdw:
I am mostly concerned about the notebooks. But let's see what happens.

caspervdw:
Everything seems to be in order with the new doc (and notebook) env. @nkeim could you also take a look at http://soft-matter.github.io/trackpy/dev/ ?

There still are a couple of warnings in the doc generation. I am ok leaving that for now, but if someone could make time to fix them... preferably I like to build docs with “warn-as-error” and use a separate GH actions runner to check them.

nkeim:
Looks great!! Thanks, @caspervdw and @rbnvrw ! It feels so good to drop all that compatibility code…

PR #658, opened 2021-04-02
DOC: add full API reference using Sphinx recursive option

rbnvrw:
I was trying to build the docs locally and found that we do not have an overview of the full trackpy API.
This adds a toctree at the bottom of the API page where you can click through all files / modules and functions.
I think it is a useful addition because also the less often used functions are browse-able through the docs in this way.

### DIFF
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -1,7 +1,5 @@
 .. _api_ref:
 
-.. currentmodule:: trackpy
-
 API reference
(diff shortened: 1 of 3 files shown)


caspervdw:
Thanks @rbnvrw . I’ll merge, but let’s have a close look at the result and see what functions we actually want to document. Not all is part of the external API.

PR #661, opened 2021-04-13
Add zenodo metadata file

caspervdw:
This file should be picked up by Zenodo to automatically set the metadata of the publication. I had to manually set them for the last release (and v0.4.2 as well). I think it is justified to add @rbnvrw to the list of authors as he did a major contribution in #527.

Other changes to the metadata can be done easily, please let me know if you think there is anyone else that should be in the authors list, or other metadata fields that should be updated.

### DIFF
--- a/.zenodo.json
+++ b/.zenodo.json
@@ -14,7 +14,7 @@
         }, 
         {
             "orcid": "0000-0003-0746-0547", 
-            "affiliation": "California Polytechnic State University", 
+            "affiliation": "Pennsylvania State University", 
             "name": "Keim, Nathan C."
         }, 
         {
--- a/.zenodo.json
+++ b/.zenodo.json
@@ -1,0 +1,32 @@
+{
+    "license": "BSD-3-Clause", 
+    "upload_type": "software", 
+    "creators": [
+        {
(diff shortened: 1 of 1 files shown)


.zenodo.json
```diff
@@ -0,0 +1,32 @@
...
+            "name": "Allan, Daniel B."
+        }, 
+        {
+            "orcid": "0000-0003-4692-608X", 
+            "affiliation": "Brookhaven National Lab", 
+            "name": "Caswell, Thomas"
+        }, 
+        {
+            "orcid": "0000-0003-0746-0547", 
+            "affiliation": "California Polytechnic State University",
```
nkeim, line 17: Since only one affiliation is allowed, please change to `"Pennsylvania State University"`.

nkeim (changes requested):
Thanks! 👍 for adding Ruben.

rbnvrw (approved):
Great, thank you @caspervdw !
And congratulations all on the new release :+1:

.zenodo.json
caspervdw, line 17:
```suggestion
            "affiliation": "Pennsylvania State University", 
```

PR #683, opened 2021-10-29
Fix versioneer compat with py311

tacaswell:
SafeConfigParser has been deprecated since Python 3.2 and will be removed in py311.

https://github.com/python/cpython/pull/28292
https://bugs.python.org/issue45173


nkeim:
Failing tests were only because numpy etc. wheels are not yet available for Python 3.10.

Issue #691, opened 2022-02-02
Allowing `DistanceMetric` strings with scikit-learn `BTree`


freemansw1:
Hi,

I'm curious if the maintainers would be amenable to a PR that allows for users to input a string that is passed to scikit-learn for `dist_func` in `trackpy.link` et al. Allowing a user-passed python `dist_func` is great, but some of the `DistanceMetric`s available in scikit-learn (https://scikit-learn.org/stable/modules/generated/sklearn.metrics.DistanceMetric.html?highlight=distancemetric#sklearn.metrics.DistanceMetric) are useful and can be substantially faster than passing along a python function. 

I'm happy to put together a PR allowing `dist_func` to be a string or a python function, and if it is a string, to pull a scikit-learn `DistanceMetric` object, if the maintainers are interested.

nkeim:
This sounds awesome! It's always a big win when the call signature doesn't change. That said, it looks like some of the sklearn metrics require arguments, most notably Minkowski. Would it make sense for `dist_func` to accept any `DistanceMetric` instance instead? Or, is it worth the extra code (and docstring text) to allow both strings and instances?

freemansw1:
Okay, great! I put in a PR to allow the user to pass any `DistanceMetric` class. I agree that that is generally better than allowing the user to pass in a string. I think it would also enable the user to extend the `DistanceMetric` class with their class and/or metrics, which would be great generally!

PR #692, opened 2022-02-07
Added ability to use custom DistanceMetric

freemansw1:
Per #691, I added the ability to use a custom scikit-learn `DistanceMetric` function as a `dist_func` parameter in `trackpy.link`. I also added a new test for this method. I *think* that this would allow users to extend the `DistanceMetric` class, too, with compiled distance functions (and therefore faster than passing in a python function).

### DIFF
--- a/trackpy/linking/linking.py
+++ b/trackpy/linking/linking.py
@@ -53,9 +53,11 @@ def link_iter(coords_iter, search_range, **kwargs):
         algorithm used to resolve subnetworks of nearby particles
         'auto' uses hybrid (numba+recursive) if available
         'drop' causes particles in subnetworks to go unlinked
-    dist_func : function, optional
-        a custom distance function that takes two 1D arrays of coordinates and
-        returns a float. Must be used with the 'BTree' neighbor_strategy.
+    dist_func : function or ```sklearn.metrics.DistanceMetric``` class, optional
+        A custom python distance function or Scikit Learn DistanceMetric class. 
+        If a python distance function is passed, it must take two 1D arrays of 
+        coordinates and return a float. Must be used with the 'BTree' 
+        neighbor_strategy.
     to_eucl : function, optional
         function that transforms a N x ndim array of positions into coordinates
         in Euclidean space. Useful for instance to link by Euclidean distance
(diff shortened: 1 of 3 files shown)


trackpy/linking/linking.py
```diff
@@ -53,9 +53,11 @@ def link_iter(coords_iter, search_range, **kwargs):
         algorithm used to resolve subnetworks of nearby particles
         'auto' uses hybrid (numba+recursive) if available
         'drop' causes particles in subnetworks to go unlinked
-    dist_func : function, optional
-        a custom distance function that takes two 1D arrays of coordinates and
-        returns a float. Must be used with the 'BTree' neighbor_strategy.
+    dist_func : function or ```sklearn.metrics.DistanceMetric``` class, optional
```
nkeim, line 56: I think this is supposed to be an instance of `DistanceMetric`, not a class.

trackpy/linking/subnet.py
```diff
@@ -10,6 +10,7 @@
 
 try:
     from sklearn.neighbors import BallTree
+    from sklearn.metrics import DistanceMetric
```
nkeim, line 13: This line appears to raise an `ImportError` in many of our automated tests, which in turn breaks `BallTree` even though `sklearn` is installed and detected. Here is an example: https://github.com/soft-matter/trackpy/runs/5127629137?check_suite_focus=true#step:6:2336 . Is `DistanceMetric` absent from earlier sklearn versions?

nkeim: Indeed, it appears to be the difference between e.g. scikit-learn 1.0.2 and scikit-learn 0.24.2 . Since DistanceMetric was only released in late 2021, this code should check for its presence separately from BallTree.

nkeim (changes requested):
Hooray! Just a few changes are needed to gracefully handle sklearn versions before late 2021.

trackpy/linking/linking.py
freemansw1: Ah, good catch.

trackpy/linking/subnet.py
freemansw1: Thanks for interpreting the CI output, I was seeing a failure but was having a hard time deducing the cause without further investigation. Looking through the API versions, it looks like they recently moved `DistanceMetric` over to `sklearn.metrics`. Before, it was in `sklearn.neighbors`. I'll update the PR to resolve this- first try importing `DistanceMetric` from `sklearn.metrics` and if an `ImportError` is raised, import it from `sklearn.neighbors`. It looks like `sklearn.neighbors.DistanceMetric` should work as far back as at least scikit-learn 0.16 (March 2015); should I still have it separately check for presence, or is it safe to assume that users will have a version newer than this? Happy to do either, but of course assuming that it is there is easier.

trackpy/linking/linking.py
freemansw1: My new commits should hopefully resolve my poor choice of documentation here.

nkeim (approved):

nkeim:
Those are (almost) certainly errors due to a new Pandas or matplotlib release, not in your code. To be safe, let's plan to merge once that's resolved.

Here's an instructive failed test that leads me to believe that: https://github.com/soft-matter/trackpy/runs/5276655344?check_suite_focus=true#step:5:41

freemansw1:
Sorry, I just want to check that you aren't waiting for me on anything @nkeim

nkeim:
Looks like whatever dependency (or GHA) issues were at play in February are now gone; all checks pass. Thanks for your contribution @freemansw1 and for your patience!

PR #706, opened 2022-07-29
Fix scipy and distutils deprecation warnings

crisp-snakey:
[Scipy 1.8 introduced depraction warnings](https://docs.scipy.org/doc/scipy-1.8.1/release.1.8.0.html#clear-split-between-public-and-private-api) for importing the submodules of `ndimage`. The solution is to import the functions directly from the `ndimage` module itself and not through one of the sub-modules that it is (accidentally) exposed through. Furthermore [distutils has been slated for deprecation for quite a while now and will be removed in python 3.12](https://docs.python.org/3/library/distutils.html). The `utils.py` file uses functionality from it to handle various versions of pandas. The distutils `LooseVersion` functionality is re-exposed through the looseversion package and is supposed to be a drop-in replacement for version the distutils counterpart.

### DIFF
--- a/.github/workflows/test-conda.yml
+++ b/.github/workflows/test-conda.yml
@@ -24,7 +24,7 @@ jobs:
       - name: Setup Environment
         shell: bash
         run: |
-          conda create --name test python=${{ matrix.python }} pytest numpy scipy matplotlib pandas pytables numba scikit-learn pyyaml
+          conda create --name test python=${{ matrix.python }} pytest numpy scipy matplotlib pandas pytables numba scikit-learn pyyaml looseversion
           source activate test
           python -V
           conda info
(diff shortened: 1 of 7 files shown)


nkeim:
Thanks! Looks great. We also need to mention `looseversion` in `doc/installation.rst`. Normally I can do this but it requires you to be placing this PR from a different branch (not `master`). Would you mind doing it instead?

At line 131, to the list of dependencies please add

```
* `looseversion <https://pypi.org/project/looseversion/>`__
```

crisp-snakey:
I've listed the dependency in the documentation like you asked.

nkeim:
Thanks for this soon-to-be essential contribution!! Merging now.

alessiamarcolini:
Hi @nkeim is there a timeline for releasing this? :)

nkeim:
Hi @alessiamarcolini ! Thank you for your patience. Yes, I've finally had a little time to work on some issues that were blocking a release. Please watch for a release candidate in the next few days.

Issue #734, opened 2023-05-24
Pandas ValueError when calling subtract_drift


kevin-duclos:
Running motion.subtract_drift sometimes produces the error below, depending on the data.

```
/Users/kevinadmin/miniconda3/envs/trackpy/lib/python3.9/site-packages/trackpy/motion.py:315: FutureWarning: reindexing with a non-unique Index is deprecated and will raise in a future version.
  traj[col] = traj[col].sub(drift[col], fill_value=0, level='frame')
Traceback (most recent call last):
  File "/Users/kevinadmin/Work/Code/pythons/diatom-sinking-processing/main.py", line 231, in <module>
    main(filepath_seq)
  File "/Users/kevinadmin/Work/Code/pythons/diatom-sinking-processing/main.py", line 147, in main
    traj_minus_drift = tp.subtract_drift(traj=traj, drift=drift_sub)
  File "/Users/kevinadmin/miniconda3/envs/trackpy/lib/python3.9/site-packages/trackpy/motion.py", line 315, in subtract_drift
    traj[col] = traj[col].sub(drift[col], fill_value=0, level='frame')
  File "/Users/kevinadmin/miniconda3/envs/trackpy/lib/python3.9/site-packages/pandas/core/frame.py", line 3980, in __setitem__
    self._set_item(key, value)
  File "/Users/kevinadmin/miniconda3/envs/trackpy/lib/python3.9/site-packages/pandas/core/frame.py", line 4174, in _set_item
    value = self._sanitize_column(value)
  File "/Users/kevinadmin/miniconda3/envs/trackpy/lib/python3.9/site-packages/pandas/core/frame.py", line 4912, in _sanitize_column
    return _reindex_for_setitem(Series(value), self.index)
  File "/Users/kevinadmin/miniconda3/envs/trackpy/lib/python3.9/site-packages/pandas/core/frame.py", line 12025, in _reindex_for_setitem
    raise err
  File "/Users/kevinadmin/miniconda3/envs/trackpy/lib/python3.9/site-packages/pandas/core/frame.py", line 12020, in _reindex_for_setitem
    reindexed_value = value.reindex(index)._values
  File "/Users/kevinadmin/miniconda3/envs/trackpy/lib/python3.9/site-packages/pandas/core/series.py", line 5094, in reindex
    return super().reindex(**kwargs)
[...]

PR #735, opened 2023-05-24
Fix for issue #734

kevin-duclos:
Change traj.set_index call to avoid Pandas error: 'ValueError: cannot reindex on an axis with duplicate labels'.


caspervdw (changes requested):
Thanks for your PR @kevin-duclos. I haven't worked with pandas for a while so I am not sure why this error is happening. But looking at the code, I think it might be a fix for your specific issue, but in general, the code should work for inputs without the `"particle"` column too. This is not reflected by the unittests however (they all use inputs with the "particle" column).

Maybe @nkeim can comment on the specific intended usage of this function?

nkeim:
Thanks @kevin-duclos . The error in #734 makes sense, except it should occur any time that there are many particles, including in the trackpy tests. So I'm confused.

I made changes to `subtract_drift()` and its docstring that I think address @caspervdw 's concern. However, at least temporarily, I added some lines to the tests that are supposed to trigger this error. We'll see if any of the CI environments catch it.

nkeim:
OK, I think this is ready to merge. In addition to the fix by @kevin-duclos , we now include two tests with no `particle` column:
- DataFrame with only one particle. This is unlikely to cause any reindexing problem.
- DataFrame with many (unlabeled) particles. This would be the case if one tried to use `subtract_drift()` with feature data prior to tracking. Note that  `compute_drift()` requires a `particle` column, so the user would need to compute the drift some other way.

I think this is ready to merge. @caspervdw does this address your question? @kevin-duclos do you think this still fixes the bug?

nkeim:
Just bumping this issue, especially for @kevin-duclos : Is this OK to merge?

caspervdw (approved):

kevin-duclos:
All good with me, @nkeim. Thanks!

xyc2718:
After the 0.6.2 update, the dataframe processed by **subtract_drift()** poses an issue for **imsd()** due to the presence of "particle" in both row and column indices. To address this, it needs to use **dataframe.reset_index(drop=True)** before running the function to ensure proper execution.

PR #739, opened 2023-06-21
BUG in utils.py (default_pos_columns)

VeraMTitze:
solution for error "unsupported operand type(s) for +: 'map' and 'list'" when locating features in 4D image sequence: default_pos_columns should return a list.


caspervdw (approved):
Thanks!

PR #743, opened 2023-11-22
MNT: scipy, Pandas, sklearn compatibility

nkeim:
Fixes failing tests for latest pandas and scipy.

### DIFF
--- a/trackpy/refine/brightfield_ring.py
+++ b/trackpy/refine/brightfield_ring.py
@@ -9,7 +9,8 @@ import warnings
 from scipy.ndimage import map_coordinates
 from scipy import stats
 
-from ..utils import (validate_tuple, guess_pos_columns, default_pos_columns)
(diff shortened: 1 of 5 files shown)


nkeim:
This is failing in a very special way on Windows. On that platform, somehow, `compute_drift()` returns a DataFrame with an `int32` index. Obviously this is completely inconsequential and the tests could ignore it… but why is it happening at all?!?

```
    def test_no_drift(self):
        N = 10
        expected = DataFrame({'x': np.zeros(N), 'y': np.zeros(N)}).iloc[1:]
        expected = expected.astype('float')
        expected.index.name = 'frame'
        expected.columns = ['x', 'y']
        # ^ no drift measured for Frame 0
    
        actual = tp.compute_drift(self.dead_still)
>       assert_frame_equal(actual, expected[['y', 'x']])
```

nkeim:
I still plan to merge and create a separate issue for the Windows test failures. Almost no actual users on Windows should be affected by this issue.

PR #747, opened 2024-01-03
[Done] Test with newer Python, deprecation fixes, drop Py3.7 support

caspervdw:


caspervdw:
Ref. https://github.com/scipy/scipy/issues/19801

caspervdw:
@tacaswell  or @danielballan   or @nkeim   if you've got a couple of minutes, please sign off this PR making trackpy ready for 2024

trackpy/utils.py
```diff
@@ -418,16 +417,14 @@ def get_pool(processes):
     return pool, map_func
 
 
-def stats_mode_scalar(*args, **kwargs):
+def stats_mode_scalar(a):
```
tacaswell, line 420: This one is a bit worrying to me as it is changing the signature.

But I am also open to "stop worrying, it won't actually break anyone".

tacaswell (approved):

trackpy/utils.py
caspervdw: Thanks for the review !

I think we’re clear because @nkeim introduced this util function in https://github.com/soft-matter/trackpy/commit/2021f28eff355d5fc0c01ad2e66c8baeea80adb0

Issue #757, opened 2024-03-07
Error in trackpy.emsd() function: TypeError: mean() got an unexpected keyword argument 'level'


Jelle111:
I encountered an error while using the trackpy.emsd() function in the trackpy library. When calling trackpy.emsd() with my trajectory data;

`em = tp.emsd(tm, 100/117., 20)`

 I received the following error:

`TypeError: mean() got an unexpected keyword argument 'level'`

Steps to Reproduce:

Execute the walkthrough tutorial with own Tif file
Call the trackpy.emsd() function with the trajectory data.
Expected Behavior:
I expected the trackpy.emsd() function to compute the ensemble mean squared displacement (EMSD) without encountering any errors.

Actual Behavior:
Instead, I received a TypeError indicating that the mean() function received an unexpected keyword argument 'level'.

Additional Information:

Version of trackpy library: 0.6.2
Operating System: MacOS Sonoma 14.2.1 with Apple M2 chip
Python version: 3.11.5

BptGrm:
Can you post the full trace ?
Which version of pandas are you using ?

xyc2718:
Due to changes in pandas version 2.0 and later, df.sum(level=1) is replaced by df.groupby(level=1).sum(), causing the emsd() function to be unavailable. I hope the author can update it as soon as possible.

Jelle111:
> Can you post the full trace ? Which version of pandas are you using ?

Full trace: 
```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[65], line 2
      1 #error: mean() got an unexpected keyword argument 'level'
----> 2 em = tp.emsd(tm, 100/117., 20)

File /Applications/anaconda3/lib/python3.11/site-packages/trackpy/motion.py:235, in emsd(traj, mpp, fps, max_lagtime, detail, pos_columns)
    233     ids.append(pid)
    234 msds = pandas_concat(msds, keys=ids, names=['particle', 'frame'])
--> 235 results = msds.mul(msds['N'], axis=0).mean(level=1)  # weighted average
    236 results = results.div(msds['N'].mean(level=1), axis=0)  # weights normalized
    237 # Above, lagt is lumped in with the rest for simplicity and speed.
    238 # Here, rebuild it from the frame index.

File /Applications/anaconda3/lib/python3.11/site-packages/pandas/core/frame.py:11335, in DataFrame.mean(self, axis, skipna, numeric_only, **kwargs)
[...]

PR #758, opened 2024-03-09
fix issue #757

hz-xiaxz:
issue #757 is due to the current update of pandas function `df.sum(level=1)` which should be replaced by `df.groupby(level=1).sum()`. I noticed that this bug has alright been fixed in the v0.62 release, however only two of three these problems were found. This PR solves the third.

I hope the PR can be merged soon.

PS: I'm new to github and may be unconcious of some rules in PR. If any unproper adjustment is made, please let me know.

### DIFF
--- a/doc/installation.rst
+++ b/doc/installation.rst
@@ -66,13 +66,14 @@ run this in the command prompt:
 Latest Version Under Development
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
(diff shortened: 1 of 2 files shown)


nkeim:
Thanks for this fix, @hz-xiaxz ! And thank you for your patience until I fixed our automated tests.

vivarose:
Good catch, @hz-xiaxz!

PR #763, opened 2024-05-24
Fix conda workflow

### LINKED ISSUE #755: TST: Need to replace a GitHub action
CI tests are failing because of a change in GitHub actions:

> Node.js 16 actions are deprecated. Please update the following actions to use Node.js 20: actions/checkout@v2, s-weigand/setup-conda@v1. For more information see: https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/.

Changing to checkout@v4 is easy enough, but the conda action has not been updated. We need to switch to another one, which may entail changing how we describe the environments.

nkeim:
Resolves #755 . I will self-merge if this is working, to unblock other PRs.

### DIFF
--- a/.github/workflows/test-conda.yml
+++ b/.github/workflows/test-conda.yml
@@ -13,25 +13,31 @@ jobs:
         python: [3.8, 3.9, "3.10", "3.11"]
 
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v4
 
-      - name: Setup Conda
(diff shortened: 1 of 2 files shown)


nkeim:
Ready to merge. In the absence of feedback, I'll let it mellow for at least a couple days. This is probably not the most efficient or correct way to set up the test environment, but it's close enough to the previous one and to what users will do.

Issue #772, opened 2024-08-16
Incorrect zeros included in emsd average when particles are dropped


vivarose:
Have you ever noticed that trackpy imsd functions always look like they drop to zero at the longest lag time?

They shouldn't, but they do.

This zero is included in the emsd average. This appears to be incorrect.

I don't think this is particularly rare. The walkthrough shows many imsds that drop to zero. Each of them would pull down the emsd incorrectly.

I have posted a short jupyter notebook that shows the issue here: [test_trackpy.ipynb](https://github.com/horowitz-lab/using_trackpy/blob/master/test_trackpy.ipynb)

vivarose:
As I wrote in my pull-request, now that I've fixed the issue, I find that it didn't make that big of a difference on my data. But I still can't stand inaccuracies, so I plan to use my fixed version going forward!

nkeim:
Closed by #773 . Thanks, @vivarose !!

PR #773, opened 2024-08-18
Fix issue #772, Incorrect zeros included in emsd average when particles are dropped 

vivarose:
I've tested this code with imsd() and emsd() on test data and it works, and on actual movie data and it is more accurate but it doesn't make as big a difference as I feared it would. So Issue #772 isn't so bad unless there are a lot of gaps in the data.

Pros of accepting my pull-request: Slightly more accurate emsd data! If there are a lot of gaps, then it's MUCH MUCH more accurate.

Cons: It takes slightly more time to run. With timit, I see 30.6 ms ± 1.55 ms per loop before my edits versus 32.1 ms ± 1.43 ms per loop after my edits.

Issue #772 occurs when there are gaps in the data, so the problem shows up in _msd_gaps().

Here are the edits:
I updated _msd_gaps() with "skipna=False" on the sum() function. This way values based on a lack of data will show up as NaN (instead of showing up as zero).

I also added some code to reset the estimated number of datapoints N to be 0 if there is no data.
    result['N'] = np.where(result['msd'].isna(), 0, result['N'])
 
I was hoping this would be enough but it created another problem that I had to take care of: When the emsd is calculated, there's a problem because we are averaging in a NaN (albeit weighted as N=0), and it turns out NaN * 0 = NaN. I wanted it to come out to zero. So I had to reset the NaN to zero again. Awkward, but it works. Then it's 0*0 = 0, which is the weight I wanted, and it calculates the emsd correctly.

    # remove np.nan because it would make the rest of the calculation break
    msds['msd'] = np.where(msds['msd'].isna(), 0, msds['msd'])

So this update does provide a more accurate emsd() calculation, and it removes the weird DROPS that show up in the imsd data (see Walkthrough output 31).

I plan to use this version going forward because I can't stand inaccuracies!

### DIFF
--- a/trackpy/motion.py
+++ b/trackpy/motion.py
@@ -108,11 +108,15 @@ def _msd_gaps(traj, mpp, fps, max_lagtime=100, detail=False, pos_columns=None):
 
     result = pd.DataFrame(_msd_iter(pos.values, lagtimes),
                           columns=result_columns, index=lagtimes)
-    result['msd'] = result[result_columns[-len(pos_columns):]].sum(1)
+    result['msd'] = result[result_columns[-len(pos_columns):]].sum(1, skipna=False)
     if detail:
         # effective number of measurements
         # approximately corrected with number of gaps
         result['N'] = _msd_N(len(pos), lagtimes) * len(traj) / len(pos)
+        # If MSD is nan that's because there were zero datapoints. Reset N to 0.
+        result['N'] = np.where(result['msd'].isna(), 0, result['N'])
+        # An alternative option at this point would be to scale up N for the rest of the column
+        
     result['lagt'] = result.index.values/float(fps)
     result.index.name = 'lagt'
     return result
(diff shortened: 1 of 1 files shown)


vivarose:
It looks like the NaN makes the tests unhappy.

I prefer the NaN because an MSD of 0 means that there was a particle that didn't move, but we are really dealing with a situation of a particle that wasn't observed. But if it makes the tests unhappy, then, what can I do?

vivarose:
I see that my attempt to appease the checks (https://github.com/soft-matter/trackpy/pull/774) didn't work. That makes me think that my pull-request isn't the problem. Maybe there is something else going on.

  I'm re-opening this pull-request. If you are trying to decide whether to incorporate one of these pull-requests, this is the one that leaves NaN in the imsd whereas https://github.com/soft-matter/trackpy/pull/774 is the one that uses 0 instead. Both versions fix Issue #772.

vivarose:
@caspervdw Thanks for considering my pull-request. I'm curious what you think about my comment about scaling up N for the rest of the column:
`    if detail:
        # effective number of measurements
        # approximately corrected with number of gaps
        result['N'] = _msd_N(len(pos), lagtimes) * len(traj) / len(pos)
        # If MSD is nan that's because there were zero datapoints. Reset N to 0.
        result['N'] = np.where(result['msd'].isna(), 0, result['N'])
        # An alternative option at this point would be to scale up N for the rest of the column`

I think for calculating the effective number of measurements N, len(traj) / len(pos) is intended to scale N. But then I set some of those N to zero and I end up with the wrong overall N: the rest of the N should be boosted up somehow to account for some of the N (at specific lagt) being (correctly!) set to zero.

PR #774, opened 2024-08-18
Fix Issue #772 

vivarose:
My previous pull-request, https://github.com/soft-matter/trackpy/pull/773 included NaN output, which was intentional (and I believe it was correct!) but caused some tests to fail.

This is a new version intended to calculate the emsd correctly in cases of particle gaps (missing data), but to avoid the NaN output. The trick is to set the number of effective measurements N to 0 in cases where there are zero effective measurements.


nkeim:
Unless I'm mistaken, this is superseded by the updates to #773 .

PR #777, opened 2024-11-04
Fix plot_traj(colorby='frame') for intermittent particles.

T0T0R:
This fix was suggested here: https://github.com/soft-matter/trackpy/issues/177#issuecomment-1059550177 but unfortunately, not included in a pull request.

Here is an example with the `viridis` cmap:
```
plt.figure()
tp.plot_traj(t2, colorby='frame', cmap='viridis');
```

And after the fix taking into account appearing and disappearing particles:

### DIFF
--- a/trackpy/plots.py
+++ b/trackpy/plots.py
@@ -335,18 +335,23 @@ def plot_traj(traj, colorby='particle', mpp=None, label=False,
         for i, trajectory in unstacked.iterrows():
             _plot(ax, mpp*trajectory, pos_columns, **_plot_style)
     if colorby == 'frame':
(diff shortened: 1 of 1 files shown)


caspervdw (approved):

caspervdw:
Thanks @T0T0R for your contribution. And sorry for the late response; all of the maintainers are doing this "on the side" and have almost no time for reviewing PRs.

I think we can just merge this as this is a clear improvement on the state.

nkeim:
Thanks for this @T0T0R ! This pairs wonderfully with #773 .

PR #786, opened 2025-06-19
MNT: Avoid deprecated DataFrame.applymap

nkeim:
When running the walkthrough with recent Pandas, this avoids the warning `FutureWarning: DataFrame.applymap has been deprecated. Use DataFrame.map instead.`


anntzer:
@nkeim Should be worth revisiting now, as pandas 2.3 has removed applymap.  I'll leave it to you to decide whether to create a shim or to just require pandas 2.1.

PR #789, opened 2025-06-25
Clarify what parameters are kwonly in documented signature of link_foo()

anntzer:
The documented signatures did not clearly show keyword-only parameters; fix that.

### DIFF
--- a/trackpy/linking/linking.py
+++ b/trackpy/linking/linking.py
@@ -19,9 +19,10 @@ logger = logging.getLogger(__name__)
 
 def link_iter(coords_iter, search_range, **kwargs):
     """
-    link_iter(coords_iter, search_range, memory=0, predictor=None,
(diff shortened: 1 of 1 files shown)


nkeim (approved):

nkeim:
Thanks for this quality-of-life fix, @anntzer !

anntzer:
> since surely some users have been treating memory as a positional argument, and their code will break.

To be clear, there's no code breakage here.  I simply changed the *documented* signature to match the actual signature.  batch() and locate() don't have a signature documented in the docstring so they are not affected.

Changing the actual implementation to use kwonly args could be nice too, but that's a bigger PR which I'm not particularly keen on writing right now.

nkeim:
Wow, I read it as an actual change to the signature… Sorry about that. Yes, this is great. I'll take a look at the slicerator PRs.

Issue #450, opened 2017-09-16
Add Hungarian linking algorithm.


danielballan:
Way back in #108 (June 2014) I mentioned my interest in integrating the Hungarian algorithm with trackpy. As illustrated in [Jaqaman 2008](http://www.nature.com/nmeth/journal/v5/n8/abs/nmeth.1237.html), it handles trajectories that split and merge. I recently received an email from a trackpy user who desires this feature, and it reminded me of my interest in it. I think I still have a very old branch somewhere with about 50% of the work done. Anyway, I couldn't find a dedicated issue for this -- just my old aside in #108 -- so this is one.

PR #795, opened 2025-11-04
Match subnets in O(n^3) via linear_sum_assignment.

anntzer:
To solve subnets, instead of brute-forcing the n! combinations, use scipy.optimize.linear_sum_assignment (commonly known as the Hungarian/Munkres algorithm, although scipy actually uses a different algorithm) which provides a solution in O(n^3).  This is not an original idea; see e.g. #450 and
https://imagej.net/imagej-wiki-static/TrackMate_Algorithms.html#Solving_LAP

Locally, this method solves the "slow" example in adaptive-search.ipynb (`tracks_regular = trackpy.link_df(cg, 0.75)`) in ~50ms instead of 1min. linear_sum_assignment ("lsa") was extensively benchmarked in scipy PR#12541 (which was most about adding a *sparse* variant, but one can just look at the performance of lsa), which shows sub-second runtimes for thousands of inputs.  This is also why this PR fully skips the use of MAX_SUB_NET (at most, one may consider setting an alternate value in the thousands for this strategy...).

(In fact, it may perhaps(?) be possible to get even better performance via by completely dropping the subnet paradigm and just passing the whole distance matrix (sparsified by removing edges greater than the search range) to scipy.sparse.csgraph.min_weight_full_bipartite_matching, but 1) one needs to figure out how to handle unmatched points (maybe by adding "dummy" points with a high but finite link cost), and 2) this would require more massive code surgery anyways.  Edit: Maybe this doesn't actually work if we want to maintain the current approach of having too-long edges within a subnet be set to search_range\*\*2.)

The docs will need to be updated, but this PR is just to discuss the implementation.

### DIFF
--- a/trackpy/linking/subnetlinker.py
+++ b/trackpy/linking/subnetlinker.py
@@ -5,11 +5,43 @@ can be linked in more than one way.
 """
 
 from collections import deque
+import operator
 
 import numpy as np
+from scipy import optimize
 
 from .utils import SubnetOversizeException
 from ..try_numba import try_numba_jit
+
+
+def subnet_linker_lsa(source_set, dest_set, search_range, max_size=None):
+    uuid = operator.attrgetter("uuid")
+    src = sorted(source_set, key=uuid)
+    dst = sorted(dest_set, key=uuid)
+    dst_uuid2idx = {d.uuid: i for i, d in enumerate(dst)}
+    # "Too-far" pairs are actually assigned a distance of search_range; see test_penalty.
+    inf = search_range ** 2
+    d2s = np.full((len(src), len(dst)), inf)
+    for i, s in enumerate(src):
+        for d, dist in s.forward_cands:
+            d2s[i, dst_uuid2idx[d.uuid]] = dist ** 2
+    src_idxs, dst_idxs = optimize.linear_sum_assignment(d2s)
+    keep = d2s[src_idxs, dst_idxs] < inf  # Other pairs were actually too far.
+    src_idxs = src_idxs[keep].tolist()
+    dst_idxs = dst_idxs[keep].tolist()
+    lost_src = sorted({*range(len(src))} - {*src_idxs})
+    lost_dst = sorted({*range(len(dst))} - {*dst_idxs})
+    sn_spl = [
+        *[src[i] for i in src_idxs],
+        *[src[i] for i in lost_src],
+        *[None] * len(lost_dst),
+    ]
+    sn_dpl = [
+        *[dst[i] for i in dst_idxs],
+        *[None] * len(lost_src),
+        *[dst[i] for i in lost_dst],
+    ]
+    return sn_spl, sn_dpl
 
 
 def recursive_linker_obj(s_sn, dest_size, search_range, max_size=30, diag=False):
--- a/trackpy/tests/test_linking.py
+++ b/trackpy/tests/test_linking.py
@@ -304,12 +304,13 @@ class CommonTrackingTests(StrictTestCase):
     def test_oversize_fail(self):
         with self.assertRaises(SubnetOversizeException):
             df = contracting_grid()
-            self.link(df, search_range=2)
+            self.link(df, search_range=2, link_strategy='recursive')
 
     def test_adaptive_fail(self):
         """Check recursion limit"""
         with self.assertRaises(SubnetOversizeException):
-            self.link(contracting_grid(), search_range=2, adaptive_stop=1.84)
+            self.link(contracting_grid(), search_range=2, adaptive_stop=1.84,
+                      link_strategy='recursive')
 
     def link(self, f, search_range, *args, **kwargs):
         kwargs = dict(self.linker_opts, **kwargs)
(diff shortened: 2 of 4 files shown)


b-grimaud:
I pulled this PR locally to try it out, and while I can see a slight difference in speed when running on individual files, it's not much different than numba when running on larger samples of my own data. 

Just to make sure I'm not missing something, is it enough to pass `lsa` as linking strategy or is there something else than needs to be specified ? Are there cases where this method is more advantageous than others, e.g. a higher or lower number of features per frame ?

anntzer:
> Just to make sure I'm not missing something, is it enough to pass lsa as linking strategy or is there something else than needs to be specified ?

Yes.  (Or in fact leaving the default value also works, *if you don't have numba installed*, because I changed the default only in that case.)

> Are there cases where this method is more advantageous than others, e.g. a higher or lower number of features per frame ?

Yes, this is specifically better in the case where you have a lot of features that are within search_range of each other, but which all move together in some coordinated fashion (e.g., the "slow" example at https://soft-matter.github.io/trackpy/v0.7/tutorial/adaptive-search.html).  In my specific use case, this occurs because I have multiple features that are all detected on a single larger solid object, these features simply track the solid's motion, and the spacing between the features is similar to the distance by which the object (and thus the features) moves from one frame to the next.

trackpy/linking/subnetlinker.py
```diff
@@ -7,11 +7,41 @@
...
 import numpy as np
+from scipy import optimize
 
 from .utils import SubnetOversizeException
 from ..try_numba import try_numba_jit
 
 
+def subnet_linker_lsa(source_set, dest_set, search_range, max_size=None):
+    src = [*source_set]
+    dst = [*dest_set]
```
nkeim: In #802 we now sort the candidates by `uuid` so that degenerate cases are stable. There are also new tests for this. Sorting will probably need to be added here.

trackpy/tests/test_linking.py
```diff
@@ -304,12 +304,13 @@ def to_eucl(arr):
     def test_oversize_fail(self):
         with self.assertRaises(SubnetOversizeException):
             df = contracting_grid()
-            self.link(df, search_range=2)
+            self.link(df, search_range=2, link_strategy='recursive')
```
nkeim, line 307: This is mind-blowing. It feels weird to not have a `SubnetOversizeException`. I think it would be user-friendly to throw it if there are many particles and `search_range` was chosen very poorly. O(n^3) is so gentle that it's really hard to pick a sensible value that works for everybody. A little benchmarking of our own would be required.

trackpy/tests/test_linking.py
nkeim, line 1: Do you need to subclass the subnet tester to use `lsa`?

nkeim (commented):
Wow, @anntzer ! This is so exciting. It would be great to reduce our dependency on numba, too. Is there a minimum scipy version required?

We already compute a sparse distance matrix, in effect, thanks to `search_range` being passed to cKDTree etc. This is what makes the upstream part of the linking algorithm (which assigns trivial links and identifies the non-trivial subnets) O(N log N) instead of N^2. But I think that taking advantage of this fact would require, as you say, a major overhaul that would lock us into the LSA algorithm exclusively. Maybe in a later version of trackpy.

trackpy/linking/subnetlinker.py
anntzer: Sure, done.

trackpy/tests/test_linking.py
anntzer: As noted above, my use case actually has a large number of particles, all on the same solid (but slightly deformable) object that are moving by more than the inter-particle distance between each frame, and I am effectively using (abusing?) trackpy to track the motion of the whole object.  So it's not clear to me that you can have a universally useful definition of "poorly chosen search_range".

trackpy/tests/test_linking.py
anntzer: I'm not really sure what you mean here (although to be honest I didn't really think very hard about it either).

anntzer:
linear_sum_assignment is available since scipy 0.17.0, whereas trackpy already requires scipy 1.4.

trackpy/tests/test_linking.py
nkeim: My group routinely tracks 10^5 particles, so I look forward to trying this sometime :).

trackpy/tests/test_linking.py
nkeim: I just pushed this to the PR. The tests pass!

trackpy/tests/test_linking.py
nkeim: In any case, I agree that a `SubnetOversizeException` is not really suitable for this algorithm, but we may revisit it once more users try it for more things.

trackpy/linking/subnetlinker.py
nkeim: Did you already implement this? It looks like it didn't get pushed yet.

anntzer: Oops, forgot to push, should be done now.

PR #796, opened 2025-11-07
emsd results index is lagt

### LINKED ISSUE #607: trackpy.utils.fit_powerlaw() fits emsd vs index, not necessarily emsd vs lag time if details = True
Hello,

I have noticed that if the emsd is calculated with details = True, then the index of the emsd DataFrame remains as row number, and not the lag time. This means that if I subsequently perform a linear fit with trackpy.utils.fit_powerlaw(), the x-data will be the index and not the lag time. 

This choice for the index in emsd happens in motion.py, line 246:

 "if not detail:
        return results.set_index('lagt')['msd']"

What purpose does this serve? Shouldn't the index of emsd always be the same, and the same as that of the imsd DataFrame, which is always the lag time?  

Or am I missing something? 

Thank you very much!

vivarose:
Fix #607

### DIFF
--- a/trackpy/motion.py
+++ b/trackpy/motion.py
@@ -265,7 +265,7 @@ def emsd(traj, mpp, fps, max_lagtime=100, detail=False, pos_columns=None):
         return results.set_index('lagt')['msd']
     # correctly compute the effective number of independent measurements
     results['N'] = msds['N'].groupby(level=1).sum()
-    return results
+    return results.set_index('lagt')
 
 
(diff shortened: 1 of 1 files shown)


nkeim (approved):
Thanks!

nkeim:
@vivarose Re: your worry about breakage, as long as we're putting it in the next "major" release (v0.8) and mentioning it in the release notes, I think it's great to slightly break some users' workflows so that future users can get the correct behavior :).

PR #797, opened 2025-11-07
fix(plot): correctly render mu

### LINKED ISSUE #678: strange axis labels when plotting with physical units
According to the [documentation](http://soft-matter.github.io/trackpy/v0.3.0/generated/trackpy.plot_traj.html#trackpy.plot_traj), I can use the mpp (microns per pixel) argument in plot_traj to show physical units in the axis labels, right?
But when I pass my calibration (in my  case, 2.07) as mpp I get a strange label as '**xb5m**' 
Below are the plots produced by plot_traj using pixels (mpp=none) and microns (mpp=2.07):

tracksAA

tracksAA_micron

vivarose:
Close #678
correctly render mu symbol in plot labels

### DIFF
--- a/trackpy/plots.py
+++ b/trackpy/plots.py
@@ -317,10 +317,7 @@ def plot_traj(traj, colorby='particle', mpp=None, label=False,
         _set_labels(ax, '{} [px]', pos_columns)
         mpp = 1.  # for computations of image extent below
     else:
-        if mpl.rcParams['text.usetex']:
-            _set_labels(ax, r'{} [\textmu m]', pos_columns)
-        else:
-            _set_labels(ax, r'{} [\xb5m]', pos_columns)
+        _set_labels(ax, r'{} [$\mu$m]', pos_columns)
     # Background image
     if superimpose is not None:
         ax.imshow(superimpose, cmap=plt.cm.gray,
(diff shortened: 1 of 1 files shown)


trackpy/plots.py
```diff
@@ -317,10 +317,7 @@ def plot_traj(traj, colorby='particle', mpp=None, label=False,
         _set_labels(ax, '{} [px]', pos_columns)
         mpp = 1.  # for computations of image extent below
     else:
-        if mpl.rcParams['text.usetex']:
-            _set_labels(ax, r'{} [\textmu m]', pos_columns)
-        else:
-            _set_labels(ax, r'{} [\xb5m]', pos_columns)
+        _set_labels(ax, r'{} [$\mu$m]', pos_columns)
```
T0T0R: This seems to write µm with a slanted 'mu' symbol _µ_ m, which is not the correct prefix. Maybe you can simply replace `r'{} [\xb5m]' ` by `'{} [\xb5m]'`.

vivarose: Absolutely!

T0T0R:
??
Maybe I was not right.
Well, if you have any idea why removing that tiny `r` caused some checks to fail, I'm all ears.

vivarose:
Last time my pull request failed, it was unrelated to my edits. The main branch also would have failed. It's a lot of work to keep up with numpy, pandas, and all.

nkeim:
Thanks @T0T0R and @vivarose ! This looks like a great improvement, and type rendering is hard to test thoroughly except by using it. So, time to merge!

PR #799, opened 2026-01-26
Add SPIFF correction

### LINKED ISSUE #413: new approach for correcting errors in particle localization
Read the abstract and thought you folks might be interested:
http://www.pnas.org/content/early/2016/12/22/1619104114.short

### LINKED ISSUE #695: Implementing SPIFF to improve subpixel accuracy
I came across two papers (see below) which implement an algorithm to correct for the pixel-locking bias that occurs during subpixel localisation. Maybe it's worth implementing this in trackpy?

Papers:
https://www.pnas.org/content/114/2/221
https://www.nature.com/articles/s41598-017-14166-6?WT.feed_name=subjects_optical-manipulation-and-tweezers#Sec9

putnokiabel:
Closes #695. Closes #413.

## Feature
Based on https://www.nature.com/articles/s41598-017-14166-6?WT.feed_name=subjects_optical-manipulation-and-tweezers#Sec9 .
Use SPIFF to remove pixel-locking bias and improve sub-pixel accuracy of located features.
In the test cases shown (see `test_spiff.py`), you can see an order of magnitude better accuracy when the SPIFF correction is applied.

## Sample results
Verify by adding the following line in `test_spiff.py` and running the tests:
image

**2D**
*error before*: 0.1300
*error after SPIFF*: 0.01998

**3D**
*error before SPIFF*: 0.2486
*error after SPIFF*: 0.0227

## Note on documentation and further integration
I did not invest time into updating the documentation and further integrating it into the basic features yet, as I wanted to see what the maintainers think of the feature first.

Since it seems like SPIFF improves accuracy pretty much universally, I'd suggest the following (but feel free to ignore or suggest something else):
- Include an `apply_spiff` argument in `tp.locate()` and `tp.batch()` (could be enabled or disabled by default) and have those functions apply the SPIFF correction so the user doesn't have to know or think about it necessarily.
- Add documentation on using the new `apply_spiff` argument as well as the underlying `apply_spiff_correction` function.
- Update documentation (and walkthrough example) on subpixel bias to mention that using SPIFF is preferred to increasing the location radius (see the paper mentioned in the `apply_spiff_correction` method for more details on this).

### DIFF
--- a/trackpy/api.py
+++ b/trackpy/api.py
@@ -24,6 +24,6 @@ from .refine import refine_com, refine_leastsq
 from . import predict
 from . import utils
 from . import artificial
-from . import spiff
+from .spiff import apply_spiff_correction
 from .utils import handle_logging, ignore_logging, quiet
 from .try_numba import try_numba_jit, enable_numba, disable_numba
--- a/trackpy/api.py
+++ b/trackpy/api.py
@@ -24,5 +24,6 @@ from .refine import refine_com, refine_leastsq
 from . import predict
 from . import utils
 from . import artificial
+from . import spiff
 from .utils import handle_logging, ignore_logging, quiet
 from .try_numba import try_numba_jit, enable_numba, disable_numba
--- a/trackpy/api.py
+++ b/trackpy/api.py
@@ -24,6 +24,6 @@ from .refine import refine_com, refine_leastsq
 from . import predict
 from . import utils
 from . import artificial
-from .spiff import apply_spiff_correction
+from .spiff import apply_spiff
 from .utils import handle_logging, ignore_logging, quiet
 from .try_numba import try_numba_jit, enable_numba, disable_numba
--- a/trackpy/feature.py
+++ b/trackpy/feature.py
@@ -205,7 +205,7 @@ def locate(raw_image, diameter, minmass=None, maxsize=None, separation=None,
            noise_size=1, smoothing_size=None, threshold=None, invert=False,
            percentile=64, topn=None, preprocess=True, max_iterations=10,
            filter_before=None, filter_after=None,
-           characterize=True, engine='auto'):
+           characterize=True, engine='auto', spiff=False):
     """Locate Gaussian-like blobs of some approximate size in an image.
 
     Preprocess the image by performing a band pass and a threshold.
@@ -272,6 +272,15 @@ def locate(raw_image, diameter, minmass=None, maxsize=None, separation=None,
     characterize : boolean
         Compute "extras": eccentricity, signal, ep. True by default.
     engine : {'auto', 'python', 'numba'}
+    spiff : boolean or 'auto'
+        Apply the SPIFF sub-pixel bias correction
+        (``trackpy.spiff.apply_spiff_correction``) to the located features
+        before returning. False by default. If True, the correction is
+        applied and a warning is emitted if there are too few features for
+        a reliable correction. If ``'auto'``, the correction is applied
+        silently when there are enough features, and skipped otherwise.
+        Note that SPIFF generally works best when applied to features
+        pooled from many frames at once; see ``batch``.
 
     Returns
     -------
@@ -452,6 +461,13 @@ def locate(raw_image, diameter, minmass=None, maxsize=None, separation=None,
             ep = pd.DataFrame(ep, columns=['ep_' + cc for cc in pos_columns])
             refined_coords = pandas_concat([refined_coords, ep], axis=1)
 
+    # Optionally apply the SPIFF sub-pixel bias correction.
+    if spiff:
+        from .spiff import apply_spiff_correction
+        refined_coords = apply_spiff_correction(
+            refined_coords, pos_columns=pos_columns,
+            warn_if_insufficient=(spiff != 'auto'))
+
     # If this is a pims Frame object, it has a frame number.
     # Tag it on; this is helpful for parallelization.
     if hasattr(raw_image, 'frame_no') and raw_image.frame_no is not None:
@@ -460,7 +476,7 @@ def locate(raw_image, diameter, minmass=None, maxsize=None, separation=None,
 
 
 def batch(frames, diameter, output=None, meta=None, processes='auto',
-          after_locate=None, **kwargs):
+          after_locate=None, spiff=False, **kwargs):
     """Locate Gaussian-like blobs of some approximate size in a set of images.
 
     Preprocess the image by performing a band pass and a threshold.
@@ -497,6 +513,16 @@ def batch(frames, diameter, output=None, meta=None, processes='auto',
         - ``features``: a DataFrame containing the detected features.
 
         Furthermore it must return a DataFrame like ``features``.
+    spiff : boolean or 'auto'
+        Apply the SPIFF sub-pixel bias correction
+        (``trackpy.spiff.apply_spiff_correction``) to the combined set of
+        features after all frames have been processed. False by default.
+        If True, the correction is applied and a warning is emitted if
+        there are too few features for a reliable correction. If
+        ``'auto'``, the correction is applied silently when there are
+        enough features, and skipped otherwise. Pooling features across
+        many frames is the recommended way to use SPIFF. Note that this
+        argument is not compatible with ``output``.
     **kwargs :
         Keyword arguments that are passed to the wrapped `trackpy.locate`.
         Refer to its docstring for further details.
@@ -521,8 +547,18 @@ def batch(frames, diameter, output=None, meta=None, processes='auto',
     if "raw_image" in kwargs:
         raise KeyError("the argument `raw_image` musn't be in `kwargs`, it is "
                        "provided internally by `frames`")
+    if spiff and output is not None:
+        raise ValueError(
+            "The `spiff` correction is applied to the combined features "
+            "after all frames are processed, which is incompatible with "
+            "streaming `output`. Pass output=None to use spiff, or apply "
+            "trackpy.spiff.apply_spiff_correction yourself.")
     # Add required keyword argument
     kwargs["diameter"] = diameter
+    # SPIFF is most accurate when applied to features pooled from all
+    # frames at once, so we override any per-frame spiff request and
+    # apply the correction below.
+    kwargs["spiff"] = False
 
     if meta:
         # Gather meta information and save as YAML in current directory.
@@ -582,7 +618,12 @@ def batch(frames, diameter, output=None, meta=None, processes='auto',
 
     if output is None:
         if len(all_features) > 0:
-            return pandas_concat(all_features).reset_index(drop=True)
+            result = pandas_concat(all_features).reset_index(drop=True)
+            if spiff:
+                from .spiff import apply_spiff_correction
+                result = apply_spiff_correction(
+                    result, warn_if_insufficient=(spiff != 'auto'))
+            return result
         else:  # return empty DataFrame
             warnings.warn("No maxima found in any frame.")
             return pd.DataFrame(columns=list(features.columns) + ['frame'])
--- a/trackpy/feature.py
+++ b/trackpy/feature.py
@@ -515,14 +515,20 @@ def batch(frames, diameter, output=None, meta=None, processes='auto',
         Furthermore it must return a DataFrame like ``features``.
     spiff : boolean or 'auto'
         Apply the SPIFF sub-pixel bias correction
-        (``trackpy.spiff.apply_spiff_correction``) to the combined set of
-        features after all frames have been processed. False by default.
+        (``trackpy.spiff.apply_spiff_correction``). False by default.
         If True, the correction is applied and a warning is emitted if
         there are too few features for a reliable correction. If
         ``'auto'``, the correction is applied silently when there are
-        enough features, and skipped otherwise. Pooling features across
-        many frames is the recommended way to use SPIFF. Note that this
-        argument is not compatible with ``output``.
+        enough features, and skipped otherwise.
+
+        When ``output`` is None, the correction is applied to the
+        combined set of features after all frames have been processed.
+        Pooling features across many frames is the recommended way to
+        use SPIFF. When ``output`` is specified, features cannot be
+        pooled because they are streamed to ``output`` one frame at a
+        time; in that case the correction is applied on a frame-by-frame
+        basis instead, which is less reliable but still better than no
+        correction.
     **kwargs :
         Keyword arguments that are passed to the wrapped `trackpy.locate`.
         Refer to its docstring for further details.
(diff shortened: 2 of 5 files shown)


putnokiabel:
@nkeim
Hi, I'm new here!
Is there a process for contributing to trackpy (other than just submitting a PR and waiting)?

nkeim:
Hi! Thanks for this PR!! I’ve seen SPIFF before and I’m glad it could be part of trackpy.

The more “active” maintainers are also faculty, so our attention to trackpy (for non-urgent matters) is somewhat seasonal. I can’t give you a definite timeline, but obviously our activity tends to pick up in late Spring/early summer. We’ll do our best to take a look before then!

Nathan

On Feb 3, 2026, at 2:17 PM, Abel Putnoki ***@***.***> wrote:

﻿
[https://avatars.githubusercontent.com/u/29568469?s=20&v=4]putnokiabel left a comment (soft-matter/trackpy#799)<https://github.com/soft-matter/trackpy/pull/799#issuecomment-3843175891>

@nkeim<https://github.com/nkeim>
Hi, I'm new here!
Is there a process for contributing to trackpy (other than just submitting a PR and waiting)?

trackpy/tests/test_spiff.py
```diff
@@ -0,0 +1,45 @@
+import unittest
+
+from build.lib.trackpy.utils import default_pos_columns
```
nkeim:
```suggestion
from trackpy.utils import default_pos_columns
```

trackpy/spiff.py
```diff
@@ -0,0 +1,64 @@
...
+    :return: DataFrame
+
+    Notes
+    ----------
+    The algorithm used is inspired by "Analysis and correction of errors in
+    nanoscale particle tracking using the Single-pixel interior filling function
+    (SPIFF) algorithm" by Yuval et al.
+    The accuracy of this algorithm improves with the number of features. When
+    tracking features across multiple frames (e.g. in a video), consider locating
+    the features across all features first (using tp.batch) before applying this function
```
nkeim:
```suggestion
    the features across all frames first (using tp.batch) before applying this function
```

trackpy/spiff.py
```diff
@@ -0,0 +1,64 @@
...
+    (as opposed to applying this function for each individual frame).
+    If f contains less than 100 features, f is returned as-is, due to lack of data.
+    """
+    if len(f) < 100:
```
nkeim: How did you settle on 100?

nkeim (changes requested):
Looks great! See comments. 

I'm open to adding a `spiff` option to `locate` and `batch`. We'd have to be deliberate about the behavior when there aren't enough features. 
- Perhaps the values could be True, False, and None (or 'auto')? Only `True` would raise a warning if there were too few features.
- I'd prefer to leave it turned off by default in the next trackpy release. We can add it to the walkthrough to let people experiment with it. But you're right that making it the default in a later release could make sense — the performance penalty is small, and it's (currently) hard to imagine cases where the correction would be unwelcome.

nkeim:
I thought of one reason not to make SPIFF automatic: it can mask a poor choice of locate parameters. SPIFF works best when the corrections are small—and it does not correct other data like mass and eccentricity. 

At least until we think of something better, it's probably best to let users decide when the raw coordinates are good enough, and then apply the correction before linking—and after applying cuts on eccentricity, etc. to remove spurious particles.

trackpy/spiff.py
putnokiabel: It's not based on data or experiments, just intuition. I extracted it to a variable and commented on this to make the assumption explicit (and invite further optimization). I also reduced the requirement from 100 to 50.

My reasoning behind the number is:
- 50-100+ features seems clearly enough to make a decent correction (at least to an extent that the result is better than without a correction).
- 5 features seems clearly not enough, to an extent that applying the correction might make the results worse.
- 10-20 features would probably, on average, make results slightly better, but might make results worse in a few cases
- Prefer going for a conservative requirement (higher minimum number of features) so that we know we practically never make the results worse by applying a correction

To optimize this even further, instead of optimizing the minimum number of features, I'd sooner go for something more elaborate, like:
Not just considering the number of features, but also the distribution of the sub-pixel values across features. That way, we could lower the minimum number of features (e.g. from 50 to 10), if some basic assumptions are met, for example:
- the subpixel values are somewhat symmetric across the center of the pixel.
- There is some minimum spread (or minimum standard deviation) to the subpixel values, e.g. if we only have 10 features, and some of the subpixel values are very close to each other or even equal, we are perhaps better off not doing the correction.

I'd prefer to keep further optimization of the minimum feature requirement outside the scope of this PR (and go for a conservative limit for now) to see how well the correction actually works for the users of trackpy (across different usecases), then optimize the feature requirement further in a future PR. I could create separate Github issues for further optimizing the SPIFF correction.

putnokiabel:
Thank you @nkeim for the review!
I applied your code suggestions, and added a `spiff` option to `locate` and `batch` (with `True`, `False` and `'auto'` options, defaulting to `False` for now).
I reduced the minimum feature requirement from 100 -> 50, this is subject to further optimization.

trackpy/api.py
```diff
@@ -24,5 +24,6 @@
 from . import predict
 from . import utils
 from . import artificial
+from . import spiff
```
nkeim: `apply_spiff_correction` should be imported. Given the problems of applying SPIFF on the first `locate()` pass, many users would want to do it as a separate step, just before linking.

trackpy/feature.py
```diff
@@ -452,6 +461,13 @@ def locate(raw_image, diameter, minmass=None, maxsize=None, separation=None,
             ep = pd.DataFrame(ep, columns=['ep_' + cc for cc in pos_columns])
             refined_coords = pandas_concat([refined_coords, ep], axis=1)
 
+    # Optionally apply the SPIFF sub-pixel bias correction.
+    if spiff:
+        from .spiff import apply_spiff_correction
```
nkeim: Since this module is short and guaranteed to exist, the import should be at the top of the file.

trackpy/feature.py
```diff
@@ -497,6 +513,16 @@ def batch(frames, diameter, output=None, meta=None, processes='auto',
...
+    spiff : boolean or 'auto'
+        Apply the SPIFF sub-pixel bias correction
+        (``trackpy.spiff.apply_spiff_correction``) to the combined set of
+        features after all frames have been processed. False by default.
+        If True, the correction is applied and a warning is emitted if
+        there are too few features for a reliable correction. If
+        ``'auto'``, the correction is applied silently when there are
+        enough features, and skipped otherwise. Pooling features across
+        many frames is the recommended way to use SPIFF. Note that this
+        argument is not compatible with ``output``.
```
nkeim: Good catch. But, if `output` is enabled, why not apply spiff to individual frames? As an example, my group typically uses streaming not because there are too many frames, but because each frame has O(10^5) features.

nkeim (changes requested):

nkeim:
@putnokiabel I just merged #803 so that the CI tests pass. If you can, please rebase your branch onto upstream/master so we can run the checks.

Now that we can see it implemented, I want to check our reasoning one last time. `locate` and `batch` already have too many options :). By adding `spiff` as an option, we are treating it as a kind of post-processing that could be done as a separate step, but that many users will want to do routinely. I *think* the only other examples are `minmass`, `maxsize`, and `topn`, so that is a very selective group!

The argument against is what I stated above: while you're trying to figure out `diameter` etc., SPIFF removes an important signal that you should be optimizing. Would it be appropriate to caution users in the docstring? Should we just make a point of discussing this in the walkthrough?

The argument *for* is that you typically optimize your parameters on a single frame, and then you run feature-finding on all frames. So you'd only use `spiff=True` on the subsequent invocations of `locate()`. And, `spiff` is a big convenience for `batch()` with streaming, because otherwise you have to write a custom `after_locate()` function.

So, at this point I think we're making the right choice, but adding kwargs is serious business, so I wanted to lay out the reasoning and make a little room for debate!

trackpy/feature.py
putnokiabel: Good point! I updated this to apply spiff to individual frames when `output` is enabled.

Future work:
We could split `apply_spiff_correction` into two methods, one of which creates the SPIFF function based on the data, and the other actually applies the SPIFF function to data.
This way we could apply SPIFF frame-by-frame but update the SPIFF function with additional data every frame so that the correction gets more and more accurate as there as more frames.

putnokiabel:
@nkeim 
I understand the concern about adding this as a (default) kwarg, leading to users not optimizing `diameter` as the issue is masked to some extent.

I do think that the way of optimizing `diameter` based on subpixel bias that's recommended in the docs should likely be updated / nuanced if we merge this (or at least when we enable SPIFF correction by default).
The logic behind optimizing diameter based on subpixel bias (as I understand) is that you take a larger-than-necessary diameter in order to fix sub-pixel bias by including neighbouring "background" pixels.
The issue here is that the "background" pixels have noise and may even have other features/particles (in a dense system where features/particles are quite close to each other and may even come into contact).
The paper I used (https://www.nature.com/articles/s41598-017-14166-6?WT.feed_name=subjects_optical-manipulation-and-tweezers#Sec2) discusses this in the Results section, the outcome is especially visible on Figure 1 and Figure 3.
In Figure 3 you see that diameter optimization has a huge effect on tracking error without SPIFF (R=1 has a significantly higher error rate than R=2), but has a much smaller impact with SPIFF (R=1 has a slightly higher error rate than R=2 but the difference is barely noticeable, and in either case much smaller than both versions without SPIFF).
In Figure 1 and 2 you see the limitation in increasing diameter in order to "fix" sub-pixel bias. A higher diameter leads to less sub-pixel bias, but becomes less accurate because:
- nearby particles may be interpreted as a single particle (figure 1), and
- even when nearby particles are interpreted as separate, the higher diameter means that the algorithm will show the particles as being closer to each other than they actually are (figure 2).

As I understand it, in the ideal case, we don't optimize `diameter` to reduce subpixel bias, but determine the optimal diameter by other means (leading to a smaller diameter and more subpixel bias) after which the subpixel bias is corrected by SPIFF. 
In other words, especially in dense systems (where there are particles close to each other), if there is no significant subpixel bias before applying SPIFF, the diameter is actually too high and should likely be reduced.

I mentioned the diameter should likely be determined by other means. I'm not sure on what other means would be best at the moment. 
One way would be "visually" (look at a few particles, and determine their diameter in pixels): this may not be super accurate (especially if the particles are not completely uniform in shape & size) as there is an element of particle choice and subjective measurement, but it allows for an informed tradeoff (e.g. in denser systems you might choose a slightly lower diameter to prevent neighboring particles from affecting tracking, whereas in a sparser system you might choose a slightly higher diameter for more accurate measurements, since the chances of particles being very close to each other is small).
Alternatively, one could run an optimization function (e.g. from scipy.optimize) that picks the diameter by minimizing the error function based on a prior assumption on the feature shape / the feature's pixel intensity distribution (e.g. it would need to optimize differently if the features are a circle that's completely filled with a sharp outline, if feature intensity has a Gaussian distribution, or a different distribution).
I can totally see this as a useful tool that trackpy could offer, where users would go through this process:
1. Call `trackpy.optimize_diameter(frames, diameter_estimate, min_diameter=diameter_estimate/2, max_diameter=diameter_estimate*2, shape='gaussian')` (or similar)
2. Do a visual check of the diameter (trackpy could provide a function that randomly displays `n` particles with a ring of a certain diameter around them)
3. Run `track` or `batch` with the chosen diameter.

That said, diameter optimization is a somewhat unrelated issue, that, for now, could be addressed with updating documentation / warning users to first optimize parameters before enabling SPIFF. (especially since the importance of diameter optimization seems to slightly get less when spiff is enabled)

I do believe the benefit of adding this as a (disabled-by-default) argument outweighs the drawback of some users not optimizing the diameter as well, but happy to change it if you think otherwise! 
And agree on it being a post-processing step that users could separately themselves, but it being a useful added argument since it seems like something most users would want to do most of the time.

I could either:
- Update the docstrings to mention the need to do parameter optimization before enabling spiff
- Update the docs (I'd need some pointers on this as I'm not sure where/how docs are updated, I haven't seen them in this repo?). Though this is somewhat more sensitive and it might be better for someone with more experience and domain knowledge to update them.
- Remove the parameter from `locate` and `batch` (in which case SPIFF would need to be mentioned in the docs in order for users to find out about it).

What do you think?

nkeim:
Thanks! It sounds like this should be an update to the walkthrough notebook in the [trackpy-examples](https://github.com/soft-matter/trackpy-examples) repo. There just needs to be different guidance about choosing `diameter`. I'm still in favor of doing the first `locate` pass without `spiff`, just to keep things simple for novices and avoid the impression that SPIFF is a magical black box. We should save for later the (brief) explanation of SPIFF and the importance of having enough features. And we can cite the journal articles for further reading. 

I think this PR is ready to merge once the walkthrough changes are drafted!

One last thing: Could the journal article citations please be in the docstring of `apply_spiff_correction`? That way they are surfaced in the API reference docs. Those are helpful papers!

putnokiabel:
@nkeim 
Sounds good! I've updated the docstring to include a proper citation.

I also created https://github.com/soft-matter/trackpy-examples/pull/66 with an updated walkthrough.

Do you think a method to optimize the diameter automatically (like I mentioned in my previous comment) would be a useful addition to trackpy? In that case I'll create an issue (and file a separate PR for this).

trackpy/feature.py
```diff
@@ -497,6 +513,22 @@ def batch(frames, diameter, output=None, meta=None, processes='auto',
...
         Furthermore it must return a DataFrame like ``features``.
+    spiff : boolean or 'auto'
+        Apply the SPIFF sub-pixel bias correction
+        (``trackpy.spiff.apply_spiff_correction``). False by default.
+        If True, the correction is applied and a warning is emitted if
+        there are too few features for a reliable correction. If
+        ``'auto'``, the correction is applied silently when there are
+        enough features, and skipped otherwise.
+
+        When ``output`` is None, the correction is applied to the
```
nkeim: This should be more concise. Suggestion:

```
When ``output`` is None, the correction is computed once for all features in all frames (recommended). 
When ``output`` is specified, the correction is computed for each frame, which may be less accurate.
```

nkeim (changes requested):
I have this one last change to suggest. We might also consider renaming `apply_spiff_correction` to be shorter, in keeping with most other API function names in trackpy. Do you have an idea? `spiff_correct`? If not, it can be a separate PR :).

putnokiabel:
@nkeim good point! I updated the function name to just `apply_spiff` (both here and in the walkthrough) which matches the function parameters and is semantically more correct (since the last "f" stands for function, so applying a function is sufficient context).

Alternatively it could be called `correct_pixel_locking` which is more intuitive / self-explanatory for people who are not familiar with SPIFF, but on the other hand is again a longer name.

trackpy/feature.py
```diff
@@ -497,6 +513,16 @@ def batch(frames, diameter, output=None, meta=None, processes='auto',
...
         Furthermore it must return a DataFrame like ``features``.
+    spiff : boolean or 'auto'
+        Apply the SPIFF sub-pixel bias correction
+        (``trackpy.spiff.apply_spiff``). False by default.
+        If True, the correction is applied and a warning is emitted if
+        there are too few features for a reliable correction. If
+        ``'auto'``, the correction is applied silently when there are
+        enough features, and skipped otherwise.
+
+        When ``output`` is None, the correction is computed once for all features in all frames (recommended).
```
nkeim: This line and the next should be wrapped to conform with the line length in other docstrings.

nkeim (changes requested):
I think that's a better name! Thanks. But please don't let your AI agent post comments without editing.

Just one formatting change and then I think this is ready to merge before the next release. I'm not sure if @vivarose or anyone else wanted to weigh in, especially on the science side. But I think this is going to be a meaningful addition that lots of people will use.

vivarose:
Thanks for tagging me! I don't know enough about this to weigh in on the science but you are right that I am super excited about more accurate tracking. My subpixel bias graphs always look really uneven, with a peak in the center and the edges of the pixel. I'm hoping this will improve that.

putnokiabel:
@nkeim I added the line breaks. 

I looked for code style / formatting rules on the repo but didn't find any (other than setup.cfg which explicitly ignores the line length lint). Perhaps it'd be good to add some linter rules so stuff like this is done (and checked) automatically and doesn't need to be reviewed manually :)

I did not use AI in writing / posting any comments here (I did use it as essentially a fancy auto-complete when writing code), I'm curious what made you suspect otherwise.

nkeim (approved):

nkeim:
Thanks for this change. Trackpy’s code mostly adheres to PEP 8 formatting, and specifically the line length really stood out in my text editor :).

It’s fine that you’re getting some help from AI with your code (I do too). The comment that it’s helpful for function names to end in “f” is… well, I don’t know where else that could have come from. Sorry if you are in fact new to Python. 

I’m going to merge this once we've got a solution for the failing test, which looks unrelated. Really, thanks!

PR #800, opened 2026-03-07
Add colorbar and an option to plot the velocities to plot_traj()

T0T0R:
A color bar can now be added when plotting the trajectories, which is helpful with `colorby='frame'`. The title of this colorbar can be selected between "Frame number" or "Time (s)".
Example with `tp.plot_traj(t, mpp=1.0, fps=24, colorby='frame', cmap='cividis', colorbar_units='frames', superimpose=background)`:

Example with `tp.plot_traj(t, mpp=1.0, fps=24, colorby='frame', cmap='cividis', colorbar_units='real', superimpose=background)`:

Another feature is added, `colorby='velocity'` allowing to color the trajectories by the speed of the objects. The title of the color bar can be selected between "Speed (px/s)" or "Speed (µm/s)".
Example with `tp.plot_traj(t, mpp=1.0, fps=24, colorby='velocity', cmap='inferno', colorbar_units='frames', superimpose=background);`:

Example with `tp.plot_traj(t, mpp=1.0, fps=24, colorby='velocity', cmap='inferno', colorbar_units='real', superimpose=background);`:

### DIFF
--- a/trackpy/plots.py
+++ b/trackpy/plots.py
@@ -43,6 +43,13 @@ def make_axes(func):
             show_plot = (plt.get_backend() != "agg")
         else:
             show_plot = False
+        
+        #if kwargs.get('fig') is None:
+        #    kwargs['fig'] = plt.gcf()
+        #    # show plot unless the matplotlib backend is headless
+        #    show_plot = (plt.get_backend() != "agg")
+        #else:
+        #    show_plot = False
 
         # Delete legend keyword so remaining ones can be passed to plot().
         legend = kwargs.pop('legend', False)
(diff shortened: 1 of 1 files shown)


nkeim (commented):
Thanks for this pull request! In the interest of keeping trackpy's API simple and memorable, I wonder if the `colorbar_units` argument is really necessary. It should be possible to choose the units automatically based on whether `mpp` or `fps` (or both) are given, which is currently the behavior for the position axes. This also avoids the ambiguity of the half-`real` units µm/frame and px/s.

Also, the current `master` branch fixes the automated checks which would be helpful for testing. If you're comfortable with rebasing this PR, please do so; no worries if not.

PR #801, opened 2026-05-04
FIX: Stable degeneracy

### LINKED ISSUE #776: `memory` not stable between runs
It seems that `memory`, under certain conditions, is not stable/deterministic. I've put an example below. At some times, the value of `particle` at index 10 is 3, and at others it is 4. Both are reasonable choices, but it should always select one or the other, not both. 

I've tried to dig into the source code with an eye to a PR, but frankly I'm struggling to understand how the memory module works. I'm happy to put some effort into a PR, but need a bit more insight there before putting one together. 

I'm on an M2 Mac, Python 3.12.7, `trackpy` version `0.6.4`. 

```
import pandas as pd
import trackpy
import numpy as np

y_coords = [2, 3, 0, 10, 3, 1, 11, 4, 1, 2, 12, 5, 0]
x_coords = [3, 4, 4, 14, 4, 5, 15, 5, 2, 6, 16, 6, 1]
frame =    [2, 2, 2,  2, 3, 3,  3, 4, 4, 4,  4, 5, 5]

trackpy_df = pd.DataFrame({"y": y_coords, "x": x_coords, "frame": frame})
pred = trackpy.predict.NearestVelocityPredict(                
        initial_guess_positions=[[1,2],[4,5]],
        initial_guess_vels=[[1,1],[-1,-1]],
        pos_columns=['y', 'x'],
        span=1,
)
out = pred.link_df(trackpy_df, search_range = 1.8, memory=1, t_column="frame", neighbor_strategy="BTree",
                adaptive_step=None,
                adaptive_stop=None,
)

```

nkeim:
This resolves #776 in most if not all cases, by making the order of a `set` of `Point` objects non-random. 

While the order of a set is never guaranteed, it is based on the objects' `hash` values. Starting with Python 3.3 [these hashes were randomized](https://docs.python.org/3/reference/datamodel.html#object.__hash__) as a security measure. This issue was causing repeated linking runs to resolve degenerate cases randomly, even within the same session. This PR makes Python use the Point's `uuid` attribute (basically, its serial number) instead.

This PR adds tests to make sure that degenerate linking cases (with and without memory) always resolve the same way within the same session. It does *not* ensure that results are consistent across linkers or across sessions. (It turns out this requires a change to the linking algorithm itself, so it will be a separate PR.)

### DIFF
--- a/trackpy/linking/utils.py
+++ b/trackpy/linking/utils.py
@@ -166,6 +166,10 @@ class Point:
     def track(self):
         """Returns the track that this :class:`Point` is in.  May be `None` """
         return self._track
+    
+    def __hash__(self):
+        """Returns a deterministic value so that set() order is more stable."""
+        return self.uuid
 
 
 class TrackUnstored:
--- a/trackpy/tests/test_find_link.py
+++ b/trackpy/tests/test_find_link.py
@@ -16,6 +16,7 @@ from trackpy.tests.test_linking import SubnetNeededTests, _skip_if_no_sklearn
 class FindLinkTests(SubnetNeededTests):
     def setUp(self):
         super().setUp()
+        self.coordinates_exact = False  # b/c we roundtrip to images
         self.linker_opts['separation'] = 10
         self.linker_opts['diameter'] = 15
         self.linker_opts['preprocess'] = False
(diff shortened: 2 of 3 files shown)


nkeim:
I should note that besides fixing the issues with GitHub actions, it's important to **evaluate the effect of this PR on performance**. It's easy to see how using a custom hash could dramatically slow down linking. I would love for someone to check this with their own data!

PR #802, opened 2026-05-08
FIX: Stable linking with degeneracy, via sorting

### LINKED ISSUE #776: `memory` not stable between runs

nkeim:
This fixes #776 by making sure candidates are considered in order during subnet linking. It supersedes #801 because that fix's performance regression was much worse (and, it did not actually guarantee ordering).

Each subnet linker implementation now sorts source candidates by the number of forward candidates first, and then by their uuid. The performance penalty appears to be within 1%, too small to measure easily.

Previously, the numba linker did not sort by the number of forward candidates, which is a helpful optimization for this depth-first algorithm. Because the core linking algorithm was so fast, this optimization was not always worthwhile in practice. However, introducing sorting by uuid in the numba linker caused a serious performance regression for large subnets, much worse than for the other linker implementations. I found that when the numba linker also sorts by forward candidates, the regression seems to go away. So I include that optimization in this PR.

### DIFF
--- a/trackpy/linking/subnetlinker.py
+++ b/trackpy/linking/subnetlinker.py
@@ -27,7 +27,7 @@ class SubnetLinker:
         self.search_range = search_range
         self.max_size = max_size
         self.s_lst = [s for s in s_sn]
-        self.s_lst.sort(key=lambda x: len(x.forward_cands))
+        self.s_lst.sort(key=lambda p: (len(p.forward_cands), p.uuid))
         self.MAX = len(self.s_lst)
 
         self.max_links = min(self.MAX, dest_size)
@@ -85,7 +85,7 @@ class SubnetLinker:
 
 def nonrecursive_link(source_list, dest_size, search_range, max_size=30, diag=False):
     source_list = list(source_list)
-    source_list.sort(key=lambda x: len(x.forward_cands))
+    source_list.sort(key=lambda p: (len(p.forward_cands), p.uuid))
     MAX = len(source_list)
 
     if MAX > max_size:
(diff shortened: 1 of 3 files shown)


nkeim:
I've now checked the performance of linking in the walkthrough tutorial, and there's no significant change.

Merging, in the hope that we get performance data for more use cases. The new tests will also help us finalize #795 .

PR #805, opened 2026-06-27
MNT: Numpy 2.5.0

nkeim:


nkeim (commented):
Tests with the latest numba were failing because of this dtype change. The actual eigenvalue is still real.

https://numpy.org/devdocs/release/2.5.0-notes.html#linalg-eig-and-linalg-eigvals-now-always-return-complex-arrays

PR #806, opened 2026-07-01
Update docs

### LINKED ISSUE #666: trackpy.locate returns clarification
I am using trackpy.locate (or rather, .batch) and then .link to form trajectories, which are then processed even further.
This includes calculating the volume of the 2D gaussian for each feature found, which made me wondered what exactly is the meaning of the `mass` return of .locate. It is described as `mass means total integrated brightness of the blob`, but that data is sometimes orders of magnitude smaller than the integral I calculate separately.
Furthermore, the `signal` return isn't actually described anywhere. It might seem self-explanatory, but trackpy.locate's documentation skips over it. I know this issue has previously been mentionned here #625 and here #483, but I never found a clear explanation of what it returns.

### LINKED ISSUE #679: seems that `image` is not an input to tp.locate
https://github.com/soft-matter/trackpy/blob/3b280ea9ccbf25dd2e3f07512398071a882ef0c1/trackpy/feature.py#L222

the docstring says there are two inputs: `raw_image` and `image` - maybe I'm wrong, but there's only one input `raw_image`

putnokiabel:
Closes https://github.com/soft-matter/trackpy/issues/679 Closes #666.

- Remove unused `image` from the `locate` docstring (as suggested in #679).
- Document the output columns of `locate` and `batch` (as suggested in #666).

### DIFF
--- a/trackpy/feature.py
+++ b/trackpy/feature.py
@@ -217,11 +217,7 @@ def locate(raw_image, diameter, minmass=None, maxsize=None, separation=None,
     ----------
     raw_image : array (any dimensions)
         Image used for final characterization. Ideally, pixel values of
-        this image are not rescaled, but it can also be identical to
-        ``image``.
-    image : array (same size as raw_image)
-        Processed image used for centroid-finding and most particle
-        measurements.
+        this image are not rescaled.
     diameter : odd integer or tuple of odd integers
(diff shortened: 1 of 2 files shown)


nkeim:
Thanks @putnokiabel for working on this old issue! I revised the text a bit and made some further corrections that were not noted in the previous issues.

